PR #27, opened 2025-01-14
[ATfL] A script which generates modulefiles directory tree that could be pointed at with MODULEPATH

pawosm-arm:
This is a proof of concept script which generates a directory tree with modulefiles. It may be used later when the RPM and DEB post-install shell scripts are being implemented.

Currently it can be used as such:

In the ATfL tarball (as created by build.sh) a new directory called `arm` is being created containing the `mkmodulesdir.sh` script. This script is being amended by `build.sh` in order to adjust the ATfL version info. After unpacking the tarball, an user can execute this script, it will create three directories in the current directory:

```
moduledeps
modulefiles
moduleglobals
```

These contain all the files needed to load ATfL as a module.

Assuming this script has been executed in the `atfl/arm` directory in the user's `$HOME`, ATfL can be loaded as such:

```
$ MODULEPATH=$HOME/atfl/arm/modulefiles module load atfl/0.0
```

### DIFF
--- a/arm-software/linux/build.sh
+++ b/arm-software/linux/build.sh
@@ -11,6 +11,7 @@ fi
 
 BASE_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 README_MD_PATH=${README_MD_PATH:-"${BASE_DIR}/README.md"}
+MKMODULEDIRS_PATH=${MKMODULEDIRS_PATH:-"${BASE_DIR}/mkmoduledirs.sh.var"}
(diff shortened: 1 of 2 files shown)


pratlucas (commented):
The script changes LGTM. I'm not too familiar with modulefiles, though, so I would prefer it if someone else has a second look and approves this PR.

DavidTruby (approved):
LGTM too but I'm also not that familiar with module files. Probably we will find it tricky to find anyone that _is_ that familiar though?

pawosm-arm:
Rebased before merge

PR #30, opened 2025-01-16
Fix build directory for ATfE's build and test scripts

pratlucas:
This fixes the build directory used by the Arm Toolchain for Embedded's
build and test scripts, ensuring the directory is consistent across
them.


HugoSilvaSantos (commented):
LGTM

PR #33, opened 2025-01-17
Ensure variant names are unique and match filenames

dcandler:
Each variant has a name property which should match the filename, however there are a number of variants where differences have crept in. In particular, the "unaligned" suffix appears missing in many cases.


pratlucas (approved):
LGTM.

Do you think we could have an automated check for that, to prevent future problems? That can be covered later, though.

dcandler:
Yes, it's easy enough to check, although having the name as a property feels somewhat redundant if we're already using the filename as the identifier, so it might also be possible to remove it entirely.

PR #34, opened 2025-01-17
Read Git repository URL from versions.json

dcandler:
Currently, the location of the Git repository for dependencies is hard-codedinto the FetchContent. This patch instead reads it from the existing versions.json file, if a different location is desired - for example, using a mirror of the repo.


pratlucas (approved):
LGTM.

PR #36, opened 2025-01-17
Enable exceptions/rtti builds of libcxx with newlib

voltur01:
Also turn localization on to support iostream.


dcandler (approved):
Instead of having a single large newlib patch, it might be an idea to start trying to break it up into individual changes. That way we can use the patch commit messages to track what is being changed and _why_ it needs to change to make newlib work.

But that's probably better tackled separately.

voltur01:
Thanks, make sense: I am about to share some of the simple changes from the patch in the newlib mail list, so hopefully they will be accepted and for newlib 4.6.0 it will be much smaller too.

PR #38, opened 2025-01-20
Add newlib samples

voltur01:


pratlucas:
I noticed some of the files contain the copyright and license information, but others don't. Should all of them include those details as a comment?

voltur01:
This is what we have in all samples - these ones are copies adapted to work with newlib - I do not remember the reasoning, but all *.c and *.cpp files do not have the comment at the top, while build scripts do.

Maybe we should have a separate PR to add the top comment to all of them, including in llvmlibc-samples folder.

pratlucas (approved):
LGTM.

PR #39, opened 2025-01-21
[ATfL] Make build.sh script more suitable to work with this codebase

pawosm-arm:
The `build.sh` script has been developed independently without any awareness of the reality in which is being used in this repository.

The source directory is now the top level one, and the extra libraries may not always be available.


arm-software/linux/build.sh
```diff
@@ -194,6 +194,14 @@ Environment Variables:
 EOF
 }
 
+libraries_resent() {
```
schander-arm, line 197: should this be libraries_present()

pawosm-arm: fixed

schander-arm (approved):

pawosm-arm:
Rebased and merged.

PR #43, opened 2025-01-22
Run GitHub Actions on ARM processors

cclauss:
[Standard GitHub-hosted runners for public repositories](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories) --> `ubuntu-24.04-arm`


pratlucas (approved):
LGTM. Thank you for the patch.

PR #45, opened 2025-01-22
Clarify QEMU dependency in newlib READMEs

voltur01:


smithp35 (approved):
Please could the text be added to the commit message as it will be easier to see from git log alone what the problem was.

PR #46, opened 2025-01-22
Add Arm copyright notices to samples

voltur01:
Add missing Arm copyright notices to C and C++ files under the samples folders.


smithp35 (approved):
Changes LGTM. Please could we mention licenses in the commit message too. May be a bit too long for the subject.

I've checked that the files changed match my search of .c/.cpp without Copyright
```
find . -type f -name \*.c\* | xargs grep -L Copyrig
ht
```

PR #48, opened 2025-01-22
Add arm-toolchain CONTRIBUTING.md including downstream patch policy

smithp35:
This is based on the "LLVM Embedded Toolchain for Arm" contributing.md with adaptations for the arm-toolchain.

The largest change is the introduction of the downstream patch policy. Arm wants the arm-toolchain to be as close to the upstream open-source projects as possible, we impose some additional requirements on downstream patches so that we can summarize all the patches and know why they were made.

Replace llvm-project's CONTRIBUTING.md rather than using another location like .github as we expect many developers to read the file from the terminal.


CONTRIBUTING.md
```diff
@@ -1,17 +1,183 @@
...
+* Testing of the above.
+* Documentation.
 
-Join the [LLVM Discourse forums](https://discourse.llvm.org/), [Discord
-chat](https://discord.gg/xS7Z362), or #llvm IRC channel on
-[OFTC](https://oftc.net/).
+The Arm Toolchain repository is synchronized with the upstream
+*llvm-project* including the main and release branches. Arm Toolchain
+for Embedded aims to stay close to the head of the *picolibc* main
+branch. The newlib overlay package aims to track the most recent
```
vhscampos, line 20: I suggest we stick to a consistent formatting of project names. In some places they are emphasized with stars, in others they aren't. For instance in this line, newlib does not have emphasis.

CONTRIBUTING.md
```diff
@@ -1,17 +1,183 @@
...
+
+## Ways to contribute
+
+### Report an issue
+
+Please make sure that the issue is specific to the *arm-toolchain*
+before reporting. If the issue exists in *llvm-project* or *picolibc*
+then please report the issue in the upstream project.
+
+Please create a Github issue in the *arm-toolchain* project
```
vhscampos, line 41: Github -> GitHub

CONTRIBUTING.md
```diff
@@ -1,17 +1,183 @@
...
+[Issues](https://github.com/arm/arm-toolchain/issues)
+list and label is as a `bug`.
```
vhscampos, line 43: We ought to check that external contributors have permissions to add labels. That is not the case for example in the ABI and the ACLE repositories.

CONTRIBUTING.md
```diff
@@ -1,17 +1,183 @@
...
+
+For a small change, please create a Pull Request as described in
+_How to submit a change_ section below.
+
+### Suggest a feature or bigger change
+
+For a bigger change, please create an issue in the
+*arm-toolchain* project
+[Issues](https://github.com/arm/arm-toolchain/issues)
+list and label is as an `rfc` (Request for Comments) to initiate the discussion
```
vhscampos, line 55: Same remark about permissions to add labels

CONTRIBUTING.md
```diff
@@ -1,17 +1,183 @@
...
+
+All contributions that can be made in an upstream project, must be
+made in the upstream project.
+
+### Suitable categories for downstream patches
+
+#### Toolchain integration
+
+Patches that are specific to integrating components of the Arm
+toolchain or the the projects that it depends on, may be implemented
```
vhscampos, line 113: "the the"

CONTRIBUTING.md
```diff
@@ -1,17 +1,183 @@
...
+  as when the upstream equivalent lands, must include Removes
+  downstream issue:#<issue number>.
+
+* New files in *llvm-project* must include the LLVM license, with the
+  standard header comment and SPDX identifier.
+
+* New files added to an external project like *picolibc* or *newlib*
+  must follow the licensing and copyright of the external project.
+
+* Any patch that causes the arm toolchain to deviate from upstream
```
vhscampos, line 179: Capitalise arm toolchain

CONTRIBUTING.md
```diff
@@ -1,17 +1,183 @@
...
+  behavior, such as a new command-line option, must be documented in
+  the arm-software/doc/features.md file. The documentation for a
```
vhscampos, line 181: Put the filepath between ticks

smithp35 (commented):
Thanks for the review. I'll upload a fixup patch.

vhscampos (approved):

CONTRIBUTING.md
```diff
@@ -1,17 +1,183 @@
...
-Thank you for your interest in contributing to LLVM! There are many ways to
-contribute, and we appreciate all contributions.
+The Arm Toolchain repository is a fork of the *llvm-project*, with
+build scripts to produce Arm Toolchain for Embedded and Arm Toolchain
+for Linux.
 
-To get started with contributing, please take a look at the
-[Contributing to LLVM](https://llvm.org/docs/Contributing.html) guide. It
-describes how to get involved, raise issues and submit patches.
+Arm Toolchain for Embedded integrates *llvm-project* with the
```
voltur01, line 7: Do we need to mention llvm-libc separately as well? It is part of llvm-project so implicitly is there, however we seem to go into much more details in regard to the other two libraries.

Taking into account that we want to use llvm-libc more, we may want to mention is explicitly.

CONTRIBUTING.md
```diff
@@ -1,17 +1,183 @@
...
+
+### Suggest a feature or bigger change
+
+For a bigger change, please create an issue in the
+*arm-toolchain* project
+[Issues](https://github.com/arm/arm-toolchain/issues)
+list and put [RFC] in the subject line (Request for Comments) to initiate the discussion
+first, before submitting a PR for the change.
+
+There is no formal template for an `rfc`, however it would be good to
```
voltur01, line 58: `rfc` was probably a label previously, so may be clearer as RFC now with the new recommendation.

CONTRIBUTING.md
```diff
@@ -1,17 +1,183 @@
...
+$ ./run-precommit-checks.sh
+```
+
+## Downstream Patch Policy
+
+### What is considered a downstream patch?
+
+A patch is considered a downstream patch if:
+
+* It touches any part of the *llvm-project* repository outside the
```
voltur01, line 99: Should this be *arm-toolchain* repository? Or maybe more complicated statement like llvm-project files inside arm-toolchin repo?

CONTRIBUTING.md
```diff
@@ -1,17 +1,183 @@
...
+
+* The source change should be annotated with a comment including the
+  #<issue number>, if there are multiple lines changed a single
+  comment for the block can be used.
+
+* The commit message that removes an existing downstream patch, such
+  as when the upstream equivalent lands, must include Removes
+  downstream issue:#<issue number>.
+
+* New files in *llvm-project* must include the LLVM license, with the
```
voltur01, line 173: Same as above: llvm-prooject files inside arm-toolchain repo?

CONTRIBUTING.md
```diff
@@ -1,17 +1,183 @@
...
+  standard header comment and SPDX identifier.
+
+* New files added to an external project like *picolibc* or *newlib*
+  must follow the licensing and copyright of the external project.
+
+* Any patch that causes the Arm Toolchain to deviate from upstream
+  behavior, such as a new command-line option, must be documented in
+  the `arm-software/doc/features.md file`. The documentation for a
```
voltur01, line 181: Should we add the placeholder file now as well?
Maybe it should have sections for common and embedded/linux specific features?

CONTRIBUTING.md
smithp35: The distinction I was trying to make was in-tree llvm-project, which includes llvm-libc, and out-of-tree picolibc and newlib. LLVM libc

I've updated the wording to hopefully make it clearer.

CONTRIBUTING.md
smithp35: ACK

CONTRIBUTING.md
smithp35: I've used the *arm-toolchain* repository and tweaked the language a bit.

CONTRIBUTING.md
smithp35: ACK

CONTRIBUTING.md
smithp35: Yes, I was in two minds over whether to do that here or in a follow up patch.

smithp35: There's an interesting design decision here. The current directory structure is:
```
arm-software
|-- ci
|-- embedded
|-----docs
|-- linux
```
With the docs directory for embedded being part of the installed package.

I think we would want the documentation of downstream features to be available in the installed package for LLVM embedded at least. Not sure about what is supposed to happen in linux.

One way of doing this is to put the downstream feature documentation in `embedded/docs/features.md` and `linux/docs/features.md` with duplication for common changes. In theory all changes are common to both, but I suspect that the users for each toolchain will care about different downstream features. For example ATfA won't care about M-profile code-generation changes.

An alternative is to add a new doc directory to `arm-software` for common docs for both embedded and linux. The tool installation could then copy the `features.md` file into the installation.

My instinct is that ATfE will have more of the downstream features, and that these won't be that useful to ATfA, so I think each toolchain can have its own features.md file.

smithp35: I'll start with a patch for embedded/docs/features.md

voltur01: Yes, this is a tricky one. On one hand, most of the LLVM changes will technically impact both, so would be easier to document in one place, but then you are right it is not obvious how to nicely include into the distribution.

I think I prefer to have a shared folder to document the changes in one place, then an extra CMake rule to copy contents into the distribution.

We may have 3 feature files: common to copy to both, embedded and linux for very clearly specific things to be copied to respective distributions.

smithp35:
I've uploaded a new patch with the common docs folder and some skeleton feature pages.

I've also removed the embedded/contributing.md as this replaces it.

voltur01 (approved):
Thank you, LGTM!

PR #49, opened 2025-01-23
[ATfL] According to rpmfind, libzstd-static is available for RHEL9, so we should use it

pawosm-arm:
This is to add support for compressed binaries. If this library is not
present, such support is silently disabled. We already have it enabled
in Ubuntu builds.


pratlucas (approved):
LGTM.

tblah:
What does using this achieve? Please could you add that answer to the commit message.

pawosm-arm:
Rebased, plus extended commit message.

PR #54, opened 2025-01-23
Reduce nesting of subproject build folders

dcandler:
Currently, each library library variant is built as a subproject, which in turn generate more subprojects within its build folder.

This patch instead uses a shared location for all subproject builds, to slightly reduce path lengths and make the structure more organized.

It also adds an option to avoid using the full variant name to label build folders, since this can make paths too long on systems with path length limits.


pratlucas (approved):
LGTM.

PR #56, opened 2025-01-24
llvm-libc build fixes

statham-arm:
Two patches here, both aimed at making the embedded build have fewer failures in the llvmlibc variant.

I'm not confident that this fixes _all_ the problems in the llvm-libc build. In a clean attempt with this change and the previous one, I had a mysterious failure, which went away when I just ran `ninja` again. That suggests some kind of intermittent issue related to build parallelism, maybe a missing dependency of some kind.

But even so, I think each of these changes removes _a_ problem, even if there's still another one yet to be found.


dcandler (approved):
LGTM

statham-arm:
Pushed the two commits manually, since "squash and merge" would have folded them together unreadably.

PR #58, opened 2025-01-24
Use variant id in stamp and binary dirs for llvmlibc-support

dcandler:
Like the other directories, the stamp and build directories for llvmlibc-support project need to use a unique identifier for each variant build, otherwise multiple versions of the same project may attempt to use them. The download and temp directories are already set properly.


statham-arm (approved):
I guess the fact that the new directory name is several levels deep won't cause a problem, because cmake will do the equivalent of `mkdir -p` and not expect all the intermediate directories to exist already?

Assuming so, LGTM.

dcandler:
It's not a problem I've observed yet: even with multiple processes potentially generating directories CMake seems to handle it.

PR #60, opened 2025-01-26
Add newlib-nano as multilib

stevew817:
See https://github.com/ARM-software/LLVM-embedded-toolchain-for-Arm/issues/628 for context.

This enables newlib-nano as a multilib. I didn't find any release scripts, so I don't know how to add it to the CI such that it also generates an overlay package for newlib-nano on release.

Note: this also fixes an issue with the current newlib builds where they don't get built correctly for anything above ARMv6 as the `-march`, `-mfloat-abi` and `-mfpu` arguments weren't being passed on to the newlib compilation configuration. A little unsure why this hasn't popped up before...


voltur01:
Hi Stephen,

Thank you for working on this and posting the patch!

I was playing with the changes, I can confirm that with this patch I can build the `newlib-nano` overlay package which works with the C sample https://github.com/arm/arm-toolchain/tree/arm-software/arm-software/embedded/newlib-samples/src/baremetal-semihosting-newlib (after adjusting the config file name in the Makefile), however the C++ sample https://github.com/arm/arm-toolchain/tree/arm-software/arm-software/embedded/newlib-samples/src/cpp-baremetal-semihosting-newlib reports missing symbols: `ungetwc`, `fgetwc`, `swprintf`. Is this expected for the nano variant?

Surprisingly, I am not able to build the normal newlib package anymore: `newlib/libm/common/math_config.h` gives multiple errors: couldn't allocate output register for constraint 'w' in `__asm__ __volatile__ ("", : "+w" (x))`

stevew817:
Thanks for the swift feedback!

> Is this expected for the nano variant?

I don't think so - according to the readme for newlib, enabling `newlib-nano-formatted-io` shouldn't impact any of the wchar-related functionality. But, when looking at what the newlib makefile is doing, it is actually dropping all wide-char-related functions from the library build when compiling with nano-formatted IO set. See [here](https://sourceware.org/git/?p=newlib-cygwin.git;a=blob;f=newlib/Makefile.in;h=211e10bfd19c53785d3aa529f83e16abac1c2c9f;hb=HEAD), line 298 - 335.
I guess that means the C++ libs accompanying nanolib need to be built without wchar too in order for that to not show? Does libstdc++ somehow account for that without a specific flag, and this is popping up here since we're using libc++ here? Or am I just not finding it in the ARMGCC build script over [here](https://gitlab.arm.com/tooling/gnu-devtools-for-arm/-/blob/main/build-baremetal-toolchain.sh#L1175)? I see that picolibc is also compiling the C++ libs without wchar support, so that's probably the 'fix'.

> Surprisingly, I am not able to build the normal newlib package anymore: newlib/libm/common/math_config.h gives multiple errors: couldn't allocate output register for constraint 'w' in __asm__ __volatile__ ("", : "+w" (x))
I'll admit I didn't retry building the regular newlib locally (as each build run takes about four hours for me... I do need a new machine). Since I didn't change any of the newlib config flags, it's probably due to now passing the actual architecture and architectural option flags to the compiler? Before I spend hours rebuilding newlib from scratch, so you have any more information on the failure (such as line number etc)?

voltur01:
Please see the `newlib` build log for one of the variants - looks like there is the same pattern of errors in multiple variants.

[2025-01-27_newlib_buildlog.txt](https://github.com/user-attachments/files/18573031/2025-01-27_newlib_buildlog.txt)

voltur01:
LLVM libc++ seems to have this option to turn wide characters support on/off: https://github.com/llvm/llvm-project/blob/83433d936195c612a51b54397f82ab0d97369d86/libcxx/CMakeLists.txt#L98

stevew817:
> Please see the `newlib` build log for one of the variants - looks like there is the same pattern of errors in multiple variants.

Thanks! Is it only on AArch64 or does it also appear on ARM?

voltur01:
I rerun the build, apparently it tries to build the first configuration which is `aarch64a_soft_nofp_exn_rtti`, fails and stops.

When I try to build armv6m and armv8.m-main separately, the build succeeds, so likely the above is an AArch64 specific issue.

stevew817:
Sorry for the delayed response, I was out on vacation.

After digging back into this, I came to the following:

- AArch64 targets with forcibly-disabled FPU ("nofp") won't be able to compile newlib with LLVM. This is due to the fact that newlib ships a whole lot of assembly optimisations for AArch64, and a lot of them use the floating-point registers. GCC happily converts that under the hood it seems, but LLVM balks at it. Since AArch64 without FPU seems to be a bit of an academic case (are there any such real-life targets?), I suggest we don't build newlib/nanolib for nofp AArch64.
- Nano specs as a whole aren't supported on AArch64 at all in ARMGCC, so I suggest we don't try to support it here either.
- Newlib built with nano formatted I/O apparently doesn't provide any of the wchar stdlib functions, even though it should according to that flag's doc. To work around this (and in the spirit of "nano"), I am suggesting to build libc++ without wide character support when it is used in conjunction with newlib-nano. However, that uncovered an issue in libc++ where it needs `mbstate_t` to be declared. According to the C standard, that type should be declared in `wchar.h`, but if libc++ is compiled without wide character support it won't even try to include `wchar.h` (even though even newlib-nano provides it). I'm providing a patch for libc++ here to fix that. If this is the way forward, that patch should probably be upstreamed to LLVM.

@voltur01 can you check the changes I just pushed?

voltur01:
Hi Steven,

Thank you for the updated patch!

I can confirm that both `newlib` and `newllib-nano` packages build successfully and I can run `newlib-samples` with both.

Code change look good as well. As you rightly pointed out, the `libcxx` changes would be best done in upstream LLVM. Would you mind creating the pull request in LLVM, please?

If you take the `libcxx` change form this PR I can approve and merge it: we only build `newlib` package in our CI so it will keep working, we can add `newlib-nano` when the upstream `libcxx` change lands.

voltur01:
Thank you! I will approve and merge now.

voltur01 (approved):

PR #62, opened 2025-01-29
Fixes in FVP invocation

vhscampos:
This patches fixes two bugs:
 - Properly propagate FVP's return code to the caller.
 - Prepend `INST=` to the `--application` argument value. This is required when the application's path contains equal signs (`=`). This sign is interpreted as a special case inside FVP: its presence in the path leads to ambiguity. One way to work around this is to do the prepend. After that, any `=` sign that comes up in the path is treated as part of the path.

This is the help message from Corstone's `--application`:
> -a, --application FILE         application to load, format: -a [INST=]FILE (use -a INST=FILE for a specific instance, use -a INST*=FILE to match multiple instances using wildcards e.g. for SMP cores)


dcandler:
The first change is a straightforward fix and looks fine, but I might be missing something for the second: the argument to `--application` is the path to the image, so why would a path contain an equal sign?

vhscampos:
There are a few examples under the libcxx tests. For instance, all tests under the directory named
`libcxx/test/std/strings/basic.string/string.nonmembers/string_op!=`, or `std/iterators/predef.iterators/move.iterators/move.iter.ops/move.iter.op.+=/`

vhscampos:
Done that. Thanks.

dcandler (approved):

PR #66, opened 2025-02-03
Disable libcxx tests on armebv6m_soft_nofp variants

vhscampos:
Some libcxx tests are failing on this variant, despite whether any of the failures are exclusive to this variant is unknown.

For now the libcxx tests will remain disabled while we focus on having C library tests passing for new variants we add.


dcandler (approved):
LGTM, but it might be worth noting in the commit message why the tests are being disabled for when we revisit them later.

PR #69, opened 2025-02-04
Fix llvmlibc sample, by adding -lm to the link command.

statham-arm:
The sample program calls two libm functions (lround and atanf) to turn a floating-point number into decimal digits. But libm.a is built as a separate library from libc.a, so linking the sample program fails because the compile command doesn't use the -lm option. Easily fixed.


pratlucas (approved):
LGTM.

PR #70, opened 2025-02-05
[ATfE] Use python executable path for picolibc tests

pratlucas:
Currently, the wrapper command for running picolibc tests in ATfE
includes a direct call to the `picolibc-test-wrapper.py` script, relying
on the script's shebang line to invoke the python interpreter. This is a
problem for invoking tests on Windows platofrms, as the naming of the
Python 3 executable is not consistent with the one used on Unix - i.e.
python vs python3.
To fix this problem, this change adds the path for the python executable
to that call. The path used is the one discovered by CMake as part of
the `find_package(Python3 ...)` command.


statham-arm (approved):
LGTM. This isn't even _only_ useful on Windows: it also helps other systems that are crossing the `python-is-python3` divide, and maybe also systems that have installed Python somewhere weird that isn't on PATH but the build system can find it some other way.

PR #71, opened 2025-02-05
[llvmlibc] Enable FP printf, using the new float320 code.

statham-arm:
The new option `LIBC_CONF_PRINTF_FLOAT_TO_STR_USE_FLOAT320` enables a strategy for floating-point printf which is both smaller and faster than the existing `USE_DYADIC_FLOAT`, and _much_ smaller than any version using a precomputed table of Ryū coefficients. I think it's reasonable to enable it in our llvm-libc builds.

Also updated the llvmlibc sample application, which was previously using a trick with `lround` to retrieve the first 6 digits of pi, precisely because FP printf wasn't available. Now that it is, it makes the sample program more comprehensible if we just use it!


statham-arm:
This depends on https://github.com/llvm/llvm-project/pull/125864 in order to build successfully, due to our `-Werror` settings. But they can be code-reviewed in parallel.

smithp35 (approved):
LGTM

PR #73, opened 2025-02-06
Add support for downloading AArch64 versions of FVPs

dcandler:
The get_fvps.sh script currently only downloads and installs the Linux64 versions of the FVPs, which does not necessarily match the host. AArch64 versions of the downloads are all available, so this patch adds a simple check to download the appropriate ones.

This also updates the version of the Corstone-310 FVP to 11.27, and modifies the script to infer the package names from the URL to reduce duplication when modifying links.


pratlucas (approved):
LGTM.

PR #76, opened 2025-02-07
Update scripts and config for Corstone-310 v11.27

dcandler:
The new version of Corstone-310 FVP model has a few differences compared to v11.24 which require changes:

* The `ID_ISAR5.PACBTI` parameter has been deprecated and removed.
* The install now includes its own copy of pythonlib and fmtplib, the paths to which need to be set as environment variables before starting the model. `PYTHONHOME` should be set to `<install>/python`, and `LD_LIBRARY_PATH` to `<install>/fmtplib:<install>/python/lib`.
* The boilerplate message printed to stdout has a different date, and the Info message on stop is no longer present. The regex has been updated to still match.

### DIFF
--- a/arm-software/embedded/arm-runtimes/test-support/run_fvp.py
+++ b/arm-software/embedded/arm-runtimes/test-support/run_fvp.py
@@ -5,6 +5,7 @@
 import re
 import subprocess
 import sys
+from os import environ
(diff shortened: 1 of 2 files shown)


pratlucas (approved):
LGTM.

PR #77, opened 2025-02-07
[ATfL] Stop using timezone database in our build of libc++

pawosm-arm:

The format of the /etc/localtime file can be different across the different Linux installations, and tzdb routines in libc++ currently cannot cope with that. In order to improve portability and to avoid any struggle with failing tzdb test cases, we should disable it.


pratlucas (approved):
LGTM.

PR #78, opened 2025-02-07
[ATfL] Fix LOG_DIR vs. LOGS_DIR inconsistency

pawosm-arm:

This patch fixes two problems:

1. Following line caused inconsistency between the variable described in help (LOG_DIR) and the one actually read (LOGS_DIR):

LOG_DIR=${LOGS_DIR:-"${BASE_DIR}/logs"}

2. All the CI scripts including the ones which execute this script are setting and using the LOGS_DIR variable.


pratlucas (approved):
LGTM.

PR #84, opened 2025-02-10
(release 20.x non-upstream patch)  [LV] Teach the vectorizer to cost and vectorize llvm.sincos intrinsics

MacDue:
This teaches the loop vectorizer that `llvm.sincos` is trivially vectorizable. Additionally, this patch updates the cost model to cost intrinsics that return multiple values correctly. Previously, the cost model only thought intrinsics that return `VectorType` need scalarizing, which meant it cost intrinsics that return multiple vectors (that need scalarizing) way too cheap (giving it the cost of a single function call).

The `llvm.sincos` intrinsic also has a custom cost when a vector function library is available, as certain VFs can be expanded (later in code-gen) to a vector function, reducing the cost to a single call (+ the possible loads from the vector function returns values via output pointers).

---

Downstream issue: #87


MacDue:
Depends on #82. This PR is a copy of the LLVM PR https://github.com/llvm/llvm-project/pull/123210, which was not possible to upstream for the release.

MacDue:
> Does this need to be kept in-sync with the upstream version - [llvm/llvm-project#123210](https://github.com/llvm/llvm-project/pull/123210)?

It's not hugely important to as all the changes to the upstream version have all been NFCs (so far), but I will update it anyway once #84 lands.

pawosm-arm (approved):

MacDue:
I'd like to hold of landing this till tomorrow morning, just in case there's any reports upstream from the changes in https://github.com/llvm/llvm-project/pull/128035, but otherwise this is ready to go.

MacDue:
@kiranchandramohan Can you merge this PR?

PR #93, opened 2025-02-17
Rebase picolibc patches

dcandler:
Commit 0dad9b1 (Add support for target-specific section names in picolibc.ld) in the picolibc repository prevents
0001-Enable-libcxx-builds.patch from being applied cleanly. Only the context around our patches has changed, so a simple rebase is enough to make them apply cleanly again.


pratlucas (approved):
LGTM.

PR #99, opened 2025-02-17
Add logic to multilib.yaml to handle cases of armebv7a_soft_nofp and armebv7a_hard_vfpv3_d16 variants

vhscampos:
These variants were missing some logic in multilib.yaml: some of the flag matching rules existed only for little endian. Thus they had to be copied for big endian as well.

Besides the changes to multilib.yaml, tests have been added to properly test the library selection of the variants.


pratlucas (approved):
LGTM. Thanks for extending the tests!

PR #100, opened 2025-02-18
[ATfE] Enable FVP testing in build script if available

dcandler:
If FVPs have been installed using the `get_fvps.sh` script, then this patch allows them to be used in the `build.sh` script by setting the `FVP_INSTALL_DIR` environment variable to the install location.


pratlucas (approved):
LGTM

PR #105, opened 2025-02-20
[ATfE] Adjust expected FVP file locations based on platform

dcandler:
The `run_fvp.py` script currently only supports running the x86_64 versions of the models, due to the paths containing a directory that only exists in the x86_64 package. This patch adjusts the paths depending on the platform, which allows the script to run both x86_64 and AArch64 versions.


pratlucas (approved):
LGTM.

PR #107, opened 2025-02-21
[ATfE] Update downstream performance patches

dcandler:
The `Additional unrolling in LTO` patch has been rebased against an upstream change, while `Prefer MEMCPY LDM/STM inlining for v7-m` has had the original commit message restored in order to provide context for the change.

The filenames now also match the subject of the commit message, as they have been regenerated using git format-patch.


VladiKrapp-Arm:
LGTM

VladiKrapp-Arm (approved):
LGTM

PR #108, opened 2025-02-21
[ATfE] Disable use of zstd when building the toolchain

pratlucas:
This patch disables the use of zstd when building the toolchain, as
using it creates an extra dependency on the `libzstd` shared library
which is not guaranteed to be present on all supported platforms.


voltur01 (approved):
Thank you!

PR #113, opened 2025-02-25
[ATfE] Disable audio device in FVP tests

dcandler:
The FVP models include an implementation of a sound device, however when running a large number of tests concurrently this can cause errors with the sound driver.

Since we do not need audio output for our tests, this patch disables the audio device.


simpal01:
LGTM

simpal01 (approved):

PR #115, opened 2025-02-25
Remove -mno-unaligned-access flag from AArch32 big endian variants.

simpal01:
The  big endian library can be selected with or without the -mno-unaligned-access flag. Currently, selecting the variants below requires explicitly passing -mno-unaligned-access. This change removes the flag from the selection criteria, ensuring that below variants can be selected regardless of whether -mno-unaligned-access is used.

- armebv7a_soft_nofp
- armebv7a_soft_nofp_exn_rtti
- armebv7a_hard_vfpv3_d16
- armebv7a_hard_vfpv3_d16_exn_rtti
- armebv7a_soft_vfpv3_d16_exn_rtti
- armebv7a_soft_vfpv3_d16


pratlucas (approved):
LGTM.

PR #116, opened 2025-02-25
[ATfE] Add additional test targets to test script

dcandler:
The `check-llvm-toolchain` target misses some available tests, however the `check-all-llvm-toolchain` target would include libcxx tests which we do not want to run from `test.sh` due to the long duration.

This patch instead adds the following targets:
* `check-all` to run the upstream clang/llvm tests, to ensure we get the same results after any downstream changes.
* `check-cxxabi` and `check-unwind` since these are much quicker to run compared to `check-cxx`.
* `check-package-llvm-toolchain` to check the package and run samples

The lit option `--ignore-fail` also allows the tests to continue after a failure, so that a failure in one variant does not prevent testing of another.


pratlucas (approved):
LGTM. Thanks for updating it!

PR #117, opened 2025-02-25
Disable debug symbols in picolibc builds

voltur01:


pratlucas:
Do we want to disable the debug symbols for all build types, or only for release builds?

voltur01:
Build types: do you mean that we have `release` and `minsize` e.g. in https://github.com/arm/arm-toolchain/blob/arm-software/arm-software/embedded/arm-multilib/json/variants/armv8m.main_hard_fp_exn_rtti.json and https://github.com/arm/arm-toolchain/blob/arm-software/arm-software/embedded/arm-multilib/json/variants/armebv6m_soft_nofp.json ?

Then yes, both.

I believe that the debug symbols are not usable anyway because they refer to source files from the build server. There may be a separate question if we want to make it possible to rebuild libraries locally with usable symbols - not sure that we have an immediate use case for that.

smithp35:
Meson looks to have some strange interaction with buildtype and debug
https://mesonbuild.com/Builtin-options.html#details-for-buildtype

I think buildtype=release would use O3 and disable debug
buildtype=minsize enables Debug so we'd need to manually disable with -debug.

I think it is reasonable to have the library debug settings separate from the tools. I think if a developer wants picolibc debug then they can alter the buildscript.

voltur01:
> I think buildtype=release would use O3 and disable debug

this was my reading of the docs as well, but still variants built with release had debug symbols included.

voltur01:
Would it help if we derive picolibc debug symbols option value based on CMAKE_BUILD_TYPE? Should be straightforward to do.

smithp35:
> Would it help if we derive picolibc debug symbols option value based on CMAKE_BUILD_TYPE? Should be straightforward to do.

Personally I don't think its worth it. It may be a bit more work than that if we want to try and map the CMAKE_BUILD_TYPEs to meson. 

Thinking of what would happen for llvm-project libraries. As I understand it they would inherit the CMAKE_BUILD_TYPE value that is used for the tools. However for picolibc we hard code the optimisation level on a per library level.

I think we could simply map meson's debug=true to CMAKE_BUILD_TYPE=Debug or CMAKE_BUILD_TYPE=RelWithDebInfo without tinkering with the optimization option. However that won't make for a good debugging experience, Ideally we'd want to turn off optimization with CMAKE_BUILD_TYPE=Debug. That would mean overriding the per-library choice.

PR #119, opened 2025-02-26
[ATfL] Get rid of the dot from the version suffix

pawosm-arm:

Apparently, an empty string is sufficient as well.


pratlucas (approved):
LGTM.

PR #122, opened 2025-02-27
[ATfL] Do not overwrite the COMMON_CMAKE_FLAGS variable, extend it instead

pawosm-arm:
This will allow us to tweak the CMake configuration.


arm-software/linux/build.sh
```diff
@@ -46,6 +46,7 @@ STAGES=(
 )
 ZLIB_STATIC_PATH=${ZLIB_STATIC_PATH:-"/usr/lib/aarch64-linux-gnu/libz.a"}
 COMMON_CMAKE_FLAGS=(
+    ${COMMON_CMAKE_FLAGS}
```
schander-arm, line 49: To me at least, it feels a little strange having 2 sets of variables with the same name.
I'm guessing ${COMMON_CMAKE_FLAGS} is coming from some external source?

pawosm-arm: Not yet, but that's the plan, to have ability to pass extra -D flags which would be seen by all the stages started by build.sh

schander-arm (approved):

PR #123, opened 2025-02-27
Disable libcxx tests on armebv6m_soft_nofp variants (#66)

simpal01:
Some libcxx tests are failing on this variant, despite whether any of the failures are exclusive to this variant is unknown.

For now the libcxx tests will remain disabled while we focus on having C library tests passing for new variants we add.

(cherry picked from https://github.com/arm/arm-toolchain/pull/66)


pratlucas (approved):
LGTM.

Issue #124, opened 2025-02-27
[Downstream Change] Arm Toolchain ID


dcandler:
Since there are potentially different versions of the toolchain that can be built (e.g. Arm Toolchain for Embedded and Arm Toolchain for Linux), we need a way to uniquely identify the toolchain from the version output, and differentiate from the upstream version. This can be done by appending an additional message: 
`Arm Toolchain ID: tag.`

Where `tag` comprises a one letter code for the product:
- E for Embedded
- L for Linux

Followed by an identifier for the build such as build number.

This issue tracks the downstream changes to implement this, since it will cause the code to diverge from upstream.

pawosm-arm:
Why can't we reuse already existing `-DCLANG_VENDOR` and `-DFLANG_VENDOR` CMake flags for this instead of adding new code?

pawosm-arm:
Anything left to do here?

dcandler:
Not currently, but now that the downstream change has been implemented this issue should track the ongoing maintenance burden, and any attempt to upstream the work so that the change can be removed.

PR #126, opened 2025-02-27
[Downstream change] Implement Arm Toolchain ID

dcandler:
Since there are potentially different versions of the toolchain that can be built (e.g. Arm Toolchain for Embedded and Arm Toolchain for Linux), we need a way to uniquely identify the toolchain from the version output, and differentiate from the upstream version. This can be done by appending an additional message:
`Arm Toolchain ID: tag.`

The contents of the tag will need to be generated by the product's build script. This patch add the field to version output, which can be set by the CMake option ARM_TOOLCHAIN_ID when building the toolchain.

Clang and clang-format have their own version functions, whereas the other LLVM tools use a shared utility function. This patch appends the new message into both.

---

Downstream issue: #124


pawosm-arm:
@kiranchandramohan are we happy about introducing this?

voltur01:
The reason why we agreed to add this at the end of the version output is to keep compatibility with upstream clang - any scripts that parse clang output to detect its version should see the standard output.

The extra Arm Toolchain ID is needed for humans to cross check the specific build in case there is a defect report.

dcandler:
My understanding was also that we wanted to leave the upstream version output alone, and instead have our own line that we can keep consistant across all executables.

Combining both `CLANG_VENDOR` and `PACKAGE_VENDOR` would allow us to insert the ID without a downstream change, but only inside the existing message.

pawosm-arm:
> My understanding was also that we wanted to leave the upstream version output alone, and instead have our own line that we can keep consistant across all executables.
> 
I think we need some demonstration of the trouble we can run into if we continue with CLANG_VENDOR, FLANG_VENDOR and PACKAGE_VENDOR. I can't believe they introduced those without good research of the consequences for vendors (like us) to use...

dcandler:
Just to get an idea of the change in formatting, setting an ID via `PACKAGE_VENDOR` (`CLANG_VENDOR` is a clang specific override for `PACKAGE_VENDOR` so both aren't necessary) for LLVM:
```
llvm-objcopy, compatible with GNU objcopy
LLVM (http://llvm.org/):
  LLVM version 21.0.0git
  Optimized build.
```
becomes
```
llvm-objcopy, compatible with GNU objcopy
Arm Toolchain ID <tag> LLVM version 21.0.0git
  Optimized build.
```

While for clang:
```
clang version 21.0.0git
Target: aarch64-unknown-linux-gnu
Thread model: posix
InstalledDir: -
```
becomes
```
Arm Toolchain ID <tag> clang version 21.0.0git
Target: aarch64-unknown-linux-gnu
Thread model: posix
InstalledDir: -
```

So while there's a more noticable difference in the LLVM output due to the missing line break, it does come down to what we're expecting to be parsing this (and what they are expecting) and whether having the ID and the LLVM version on the same line is an issue.

kiranchandramohan:
> @kiranchandramohan are we happy about introducing this? As I asked in #124, why can't we reuse already existing -DCLANG_VENDOR and -DFLANG_VENDOR CMake flags for this instead of adding new code?

Please wait till we discuss this with Will.

Will there be a separate patch for changing the ATfE and ATfL build scripts?

voltur01:
Setting the ID is here https://github.com/arm/arm-toolchain/pull/127

pawosm-arm:
> Will there be a separate patch for changing the ATfE and ATfL build scripts?

I suppose this would require branching off from currently used common development branch...

voltur01:
> this would require branching off from currently used common development branch

Could you please elaborate? Build scripts, that will inject the ID to be printed, are separate already, so ATfE and ATfL build scripts need to be updated independently to set ID correctly.

kiranchandramohan:
> > @kiranchandramohan are we happy about introducing this? As I asked in #124, why can't we reuse already existing -DCLANG_VENDOR and -DFLANG_VENDOR CMake flags for this instead of adding new code?
> 
> Please wait till we discuss this with Will.
> 
> Will there be a separate patch for changing the ATfE and ATfL build scripts?

Will would like to have a common toolchain ID format for our compiler tools. So feel free to go ahead with the patch. Please check whether the format chosen will affect detection of the compiler as clang/flang and also whether it could hamper porting from ACfL to ATfL.

voltur01:
Just to confirm, we will add a new line to the output and will not change `PACKAGE_VENDOR` or `CLANG_VENDOR` as suggested above.

pratlucas (approved):
From the discussion in this PR it appears this change will work well for both ATfE and ATfL. It also shouldn't have any impact on the compiler detection, as no change is made to the first line of the version output.

Unless there are any objections, LGTM.

PR #127, opened 2025-02-27
[ATfE] Set Arm Toolchain ID

dcandler:
In order to uniquely identify ATfE builds, this patch populates the new Arm Toolchain ID with an auto-generated tag.

This tag contains both the product code (E for Embedded), as well as a unique identifier for the build. In a Jenkins environment, this will be based on the build URL and build number.

### DIFF
--- a/arm-software/embedded/CMakeLists.txt
+++ b/arm-software/embedded/CMakeLists.txt
@@ -334,6 +334,12 @@ install(
     COMPONENT llvm-toolchain-config-files
 )
 
+# Generate a toolchain ID.
+# Product code 'E' for Embedded.
+set(LLVM_TOOLCHAIN_PROJECT_CODE "E")
+set(LLVM_TOOLCHAIN_VERSION_SUFFIX "pre")
+include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/generate_toolchain_id.cmake)
+
 set(llvmproject_src_dir ${CMAKE_CURRENT_SOURCE_DIR}/../..)
 add_subdirectory(
     ${llvmproject_src_dir}/llvm llvm
--- a/arm-software/embedded/cmake/generate_toolchain_id.cmake
+++ b/arm-software/embedded/cmake/generate_toolchain_id.cmake
@@ -1,0 +1,37 @@
+# Generate a toolchain ID.
+# This should identify the product, and where the build came from.
+
+# The ID begins with a product code (e.g. E for Embedded), followed
+# by build identifier. The build identifier will depend on the
+# build environment: for Jenkins builds, the pipeline and build
+# number will be sufficient.
+
+if(NOT DEFINED LLVM_TOOLCHAIN_PROJECT_CODE)
+    message(FATAL_ERROR "LLVM_TOOLCHAIN_PROJECT_CODE must be set before including generate_toolchain_id.cmake")
+endif()
+
+# JENKINS_URL, BUILD_URL and BUILD_NUMBER will be set in a Jenkins environment.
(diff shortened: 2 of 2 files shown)


arm-software/embedded/cmake/generate_toolchain_id.cmake
```diff
@@ -0,0 +1,37 @@
...
+# The ID begins with a product code (e.g. E for Embedded), followed
+# by build identifier. The build identifier will depend on the
+# build environment: for Jenkins builds, the pipeline and build
+# number will be sufficient.
+
+if(NOT DEFINED LLVM_TOOLCHAIN_PROJECT_CODE)
+    message(FATAL_ERROR "LLVM_TOOLCHAIN_PROJECT_CODE must be set before including generate_toolchain_id.cmake")
+endif()
+
+# JENKINS_URL, BUILD_URL and BUILD_NUMBER will be set in a Jenkins environment.
```
voltur01, line 13: Generalize to "CI" or job/agent/etc? avoid Jenkins specifically.

arm-software/embedded/CMakeLists.txt
```diff
@@ -334,6 +334,12 @@ install(
     COMPONENT llvm-toolchain-config-files
 )
 
+# Generate a toolchain ID.
+# Product code 'E' for Embedded.
+set(LLVM_TOOLCHAIN_PROJECT_CODE "E")
```
voltur01, line 339: Possible drawback of defining this inside CMake could be that downstream derived toolchains will need to change the CMake, not only CI pipelines building them.

I do not expect this to cause merge conflicts, though.

voltur01 (commented):
Thanks, I think this is workable, please see some questions if it can be generalized.

arm-software/embedded/cmake/generate_toolchain_id.cmake
```diff
@@ -0,0 +1,37 @@
...
+if(DEFINED ENV{JENKINS_URL} AND DEFINED ENV{BUILD_URL} AND DEFINED ENV{BUILD_NUMBER})
+    set(builder_name $ENV{BUILD_URL})
+    set(build_number $ENV{BUILD_NUMBER})
+    # BUILD_URL also contains the build number, which is not needed.
+    string(REGEX REPLACE "/${build_number}/$" "/" builder_name ${builder_name})
```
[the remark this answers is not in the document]
dcandler: This is intentionally CI specific, since that allows the toolchain ID to be consistently generated for a given environment. If we start using other environments, then we can expand this script with how the ID should be generated for that environment.

While we could set `BUILD_NAME` and `BUILD_NUMBER` as CMake options to create a generalized solution, that leaves it up to the CI code to decide what the build name should be. And I feel it better to have the toolchain ID implementation in the toolchain codebase.

arm-software/embedded/CMakeLists.txt
dcandler: I think that will depend on how the other toolchains derive from ATfE. At the moment, these scripts haven't been designed to make that easy: aside from the product code, there a number of other references that would need changing such as the project name and description. We could make it all parameterizable and set it just one convenient place, which would limit the potential for merge conflicts, but that would require rearranging the CMake file beyond the scope of this particular change.

voltur01: Good point, thanks.

arm-software/embedded/cmake/generate_toolchain_id.cmake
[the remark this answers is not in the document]
voltur01: Agreed, let us solve one issue at a time.

voltur01 (approved):
Thank you for the comments, I will approve now - we can extend when we will work on the public CI.

PR #130, opened 2025-02-28
[ATfL] Do not use libzstd.a when building shared runtime libs

pawosm-arm:
At the shared_lib_build stage we do not need ZSTD as the main goal of this stage is to provide shared runtime libraries, while ZSTD is being used at compile time. Sadly, there is no way (yet) to eliminate remaining attempts to build other things at this stage, and some of this stuff is linked against libzstd. Unfortunately, on some Linux distributions, those static variants of libraries that we are using to build ATfL were not built with -fPIC, resulting in linker failures. The only way to prevent this is to override one of the COMMON_CMAKE_FLAGS and disable the use of ZSTD at this stage.

### DIFF
--- a/arm-software/linux/build.sh
+++ b/arm-software/linux/build.sh
@@ -350,7 +350,7 @@ shared_lib_build() {
     run_command cmake ${CMAKE_ARGS} -G Ninja "${SOURCES_DIR}/llvm" \
         -DBUILD_SHARED_LIBS=True \
         -DLIBOMP_ENABLE_SHARED=False \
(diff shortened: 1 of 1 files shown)


DaftanoPro:
what are (or could be) other consequences of disabling ZSTD at this stage?

pawosm-arm:
> what are (or could be) other consequences of disabling ZSTD at this stage?

The binaries that we do not copy into the final tarball at this stage will not have support for ZSTD compression. The ones that land into the product tarball come from the previous state and are not affected and have full support for ZSTD compresison.

schander-arm:
If we build with ZSTD then will there be a requirement that those libraries are available on the host. We found this with the embedded toolchain builds. Or have I misunderstood.

pawosm-arm:
> If we build with ZSTD then will there be a requirement that those libraries are available on the host. We found this with the embedded toolchain builds. Or have I misunderstood.

In common flags we demand to link against static libzstd.a, so we don't demand availability on the host.

schander-arm (approved):

PR #146, opened 2025-02-28
[ATfL] Make it more adaptive to the build platform

pawosm-arm:
The compiler we build will still be targeting AArch64 only, but with some luck this may work on some other platforms too.


schander-arm (commented):
BTW on macOS platform the 'uname -m' will return arm64

pawosm-arm:
build.sh wouldn't work on non-Linux systems anyway.

pawosm-arm:
ping...?

schander-arm (approved):

PR #150, opened 2025-03-01
[ATfL] Update the list of dependencies

pawosm-arm:

Two things were changed:

- We do not require `figlet` anymore, its installation is optional

- There are packages which are missing in default installations of some of the RPM-based systems

### DIFF
--- a/arm-software/linux/build.sh-HOWTO.md
+++ b/arm-software/linux/build.sh-HOWTO.md
@@ -16,7 +16,6 @@ The following packages need to be installed (using `apt`):
 - `binutils-dev`
 - `build-essential`
 - `cmake`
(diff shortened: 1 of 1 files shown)


schander-arm (commented):
I see that we haven't stipulated the perl should be installed. I know that on some of the redhat oses just installing perl will install some, if not all, of the dependancies listed

pawosm-arm:
Yes, these things are installed automatically on RedHats. It was the AL2023 installation on which they were missing. The perl itself isn't listed as installing any of those will result in installation of perl itself, if it wasn't installed yet (which would be unusual anyway).

schander-arm (approved):

PR #153, opened 2025-03-03
[ATfE] Cherry-pick libcxx change in wchar inclusion guarding into 20.x

pratlucas:
This change is required for the newlib-nano library overlay
implementation to work correctly, as it provides the definition
`mbstate_t` despite not providing wide-character support.


voltur01 (dismissed):

statham-arm (commented):
Since this is a cherry-pick of an upstream commit on main, it's probably a good idea to mention the original commit id somewhere.

The original commit hash on main is 762001118c068317ec67274221497be2e6499c6a. But that doesn't match the value in the header line of the new patch file, presumably because it was cherry-picked before running `git format-patch`. Perhaps if you redid it with `cherry-pick -x` it would include a footer citing the original id?

pratlucas:
Good catch @statham-arm , I've updated the patch file using `cherry-pick -x`.

statham-arm (approved):

PR #162, opened 2025-03-04
[ATfE] Handle meson test return code in picolibc tests

dcandler:
The meson test command will return the number of test failures. If any test fails, this will be non-zero, which the CMake script will interpret as the command failing. The test command is immediately followed by a second command to process the results, so this will not get executed on a failure. Since the processing sets the names in the results file to make them variant specific, this makes it very unclear from the XML results which tests failed, and causes other tests not to run.

To get around this, this patch wraps the meson test command in a new CMake script which can handle the return code. Since the other lit-based tests already use the `--ignore-fail` option to control whether or not to continue on a test failure, this script will also honour the option in the `LIT_OPTS` environment variable.

If the tests fail to build, there is another special return code, so to ensure the tests failing to build (which will generate no XML results) can be distinguished from a test failing to execute, the --no-rebuild option to not bulid the tests at the same time. Instead, the tests will be built in the build step.

### DIFF
--- a/arm-software/embedded/arm-runtimes/CMakeLists.txt
+++ b/arm-software/embedded/arm-runtimes/CMakeLists.txt
@@ -429,26 +429,28 @@ if(C_LIBRARY STREQUAL picolibc)
         # So reconfigure to enable tests at a later point.
         ExternalProject_Add_Step(
             picolibc
-            enable-tests
+            compile-tests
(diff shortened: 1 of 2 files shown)


pratlucas (approved):
LGTM. Thanks for fixing this!

PR #164, opened 2025-03-05
Make Automerge script verbose

pratlucas:
This adds a `--verbose` option to the automerge script, and enables it
in the Automerge Workflow. This option will increase observability of
the steps being performed by the script, allowing for any unexpected
behaviours to be investigated.
For now the verbose option includes the logging of the following:
- All Git commands being executed by the script
- The current HEAD of the target branch before each merge attempt
- The diff containing the changes being commited for each merge

### DIFF
--- a/arm-software/ci/automerge.py
+++ b/arm-software/ci/automerge.py
@@ -47,6 +47,7 @@ class Git:
 
     def run_cmd(self, args: list[str], check: bool = True) -> str:
         git_cmd = ["git", "-C", str(self.repo_path)] + args
+        logger.debug("Running git command: %s", git_cmd)
         git_process = subprocess.run(git_cmd, check=check, capture_output=True, text=True)
         return git_process.stdout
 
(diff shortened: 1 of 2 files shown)


arm-software/ci/automerge.py
```diff
@@ -79,15 +80,21 @@ def prefix_current_commit_message(git_repo: Git) -> None:
...
+        logger.debug("Current HEAD of %s is %s", to_branch, current_head)
     git_repo.run_cmd(["merge", commit_hash, "--no-commit", "--no-ff"], check=False)
     restore_changes_to_ignored_files(git_repo, ignored_paths)
     if has_unresolved_conflicts(git_repo):
         logger.info("Merge failed")
         git_repo.run_cmd(["merge", "--abort"])
         raise MergeConflictError(commit_hash)
+    if verbose:
+        commit_content = git_repo.run_cmd(["diff", "--cached"])
+        logger.debug("Changes being commited:\n%s", commit_content)
```
statham-arm, line 97: I think this might be a step too far. Some commits will have _very_ large diffs, e.g. ones which add a huge number of tests, or churn a lot of existing ones. I think printing all the commit ids is enough for the things we're currently anticipating trying to debug.

arm-software/ci/automerge.py
```diff
@@ -79,15 +80,21 @@ def prefix_current_commit_message(git_repo: Git) -> None:
...
     git_repo.run_cmd(["commit", "--reuse-message", commit_hash])
     prefix_current_commit_message(git_repo)
```
statham-arm, line 99: On the other hand, I think it would be worth printing the hash of the merge commit once it's been made and finalized.

arm-software/ci/automerge.py
```diff
@@ -79,15 +80,21 @@ def prefix_current_commit_message(git_repo: Git) -> None:
...
     if dry_run:
```
statham-arm, line 100: I also think it would be useful to see the output of `git push`, because it will show the "before" and "after" value of the remote branch head in each case.

pratlucas: Would it make sense to print the stdout/stderr of each git command when `--verbose` is used? Or is the `git push` output enough?

arm-software/ci/automerge.py
pratlucas: That's a good point. I removed it for now.

arm-software/ci/automerge.py
pratlucas: Done.

arm-software/ci/automerge.py
pratlucas: I've included the printing of all git command outputs for now. We can remove them later if we decide it's too verbose.

statham-arm: I suppose it depends if we run any commands that can produce very large output, like `git diff`.

I see a couple of `git diff` in the script, but only with `--name-only`. That's probably all right. The list of names might get long _occasionally_, but not all the time!

statham-arm (approved):

PR #168, opened 2025-03-05
[Automerge] Detect unexpected errors when running git merge

pratlucas:
This updates the automerge script to detect unexpected errors when
running the `git merge` command for each of the commits being merged.

We cannot rely on the command's exist status for that, as it will return
a non-zero status in case a conflict is present, so we check that a
merge operation is actually in progress after `git merge` is executed.
This is done by checking for the presence of the `.git/MERGE_HEAD` file
in the repository dir.

### DIFF
--- a/arm-software/ci/automerge.py
+++ b/arm-software/ci/automerge.py
@@ -53,6 +53,12 @@ class Git:
         return git_process.stdout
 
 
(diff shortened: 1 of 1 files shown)


HugoSilvaSantos (approved):
LGTM

PR #171, opened 2025-03-06
[libc++][test] extend XFAIL clauses to cover Amazon Linux too

pawosm-arm:
It's a cherry-pick from the main branch.

The default triple of Amazon Linux on AArch64 is aarch64-amazon-linux, see issue highlighded by PR #109263, somewhat serious linker issues are encountered if any other triple is being used.

Unfortunately, this makes XFAIL lines like
`XFAIL: target=aarch64{{.*}}-linux-gnu` ineffective, making it impossible to complete all of the check-cxx on Amazon Linux without failing.


dcandler (changes requested):
Since this is a downstream change on the branch, it should follow the Downstream Patch Policy: https://github.com/arm/arm-toolchain/blob/arm-software/CONTRIBUTING.md#downstream-patch-policy

dcandler (approved):
Thanks - the policy does say to annotate the sources as well but I'm not sure how strictly that is going to be enforced on the branch. Having the tracking issue is the primary requirement.

Issue #177, opened 2025-03-07
[Downstream Change] Some test cases are failing due to the nature of time and space


pawosm-arm:
There are test cases which are very sensitive to the shared machines load. These are:

`llvm/utils/lit/tests/googletest-timeout.py`
`llvm/utils/lit/tests/max-time.py`
`llvm/utils/lit/tests/xunit-output-report-failures-only.py`
`llvm/utils/lit/tests/xunit-output.py`

As @DavidSpickett has summarized it to us:

> The issue is not speed, it's load on the machine. If you always run to the maximum capacity these tests will always be flaky, it's not a problem with the test it's a problem with the nature of time and space.
>
> The solution is not to ninja `use all the threads in the world` and not to run multiple test runs on the same instance. Leave some headroom for tests like this that spawn more threads.
>
> How much? How long is a piece of string? But some headroom anyway.
>
> Speed can help but with all the speed in the world, it would still be flakey a non-zero amount of times if you're running at and past the machine's capacity.
>
> So you have something you can tune downstream but if you really want to solve this generally, you could have some kind of "please run this at the end of the run when we probably have more resource" annotation. But this is upstream work and it does not solve the issue of shared hosts.
>
> A third option is to disable these tests downstream. The risk is very low given that they are testing the testing tool's ability to manage tests (very meta).
>
> So the thing you might have happen is some test will loop and it doesn't get killed, this would be very easy for an engineer to spot. Especially if your pipelines have a global timeout anyway (if not, great idea to add one!). Plus, if lit somehow forgot how to kill tests, upstream would have caught this because they will run this test.
>
> ...so whichever takes your fancy really. I'm not going to say one is better because I don't know the infrastructure.
>
> The other killer for shared machines is not just amount of resource, but fluctuating level of resource. This can really mess up ninja and lit.
>
> If you want to fix this right now, disabling the test downstream is the pick. Then you can come up with something more principled later, or forget all about it - I'm not judging.
>

DavidSpickett:
For `llvm/utils/lit/tests/googletest-timeout.py`, here is more about the test if you want to debate running it or not:
* It checks that lit can enforce a timeout on a googletest binary.
* It does this by starting such a binary, which then stalls, and lit should stop it before the binary finishes normally.
* If lit is stalled for a long time because it could not get a core to run on, the test binary might complete, or lit will have enforced a timeout but a greater one than expected, which is also a failure.

So if you were to not run this test in this project:
[...]

PR #178, opened 2025-03-07
[Downstream change][tests] Mark the tests failing with timeout as unsupported

pawosm-arm:
Initial upstream attempt: https://github.com/llvm/llvm-project/pull/126527

Downstream issue: https://github.com/arm/arm-toolchain/issues/177

There are some test cases which tend to fail with timeout when started on heavy loaded shared machines. As our CI machines are mostly shared, those failures tend to introduce unnecessary disturbance to our work. It is better to disable those tests.

For more context, please read issue #177.

### DIFF
--- a/llvm/utils/lit/tests/googletest-timeout.py
+++ b/llvm/utils/lit/tests/googletest-timeout.py
@@ -1,4 +1,5 @@
 # REQUIRES: lit-max-individual-test-time
+# UNSUPPORTED: system-linux
 
 ###############################################################################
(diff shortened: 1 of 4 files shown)


dcandler:
> how can I pass the --filter-out parameter to lit from here?

I don't think you can, which is why the `LIT_FILTER_OUT` environment variable is also provided for exactly this situation. Set the tests to be filtered out there, then then subsequent `ninja check-all` won't run them.

PR #179, opened 2025-03-07
Ensure automerge only runs on a clean git worktree

pratlucas:
This patch introduces an additional check at the start of the automerge
script to ensure that it only runs on a clean git worktree. This check
prevents issues where local changes might be included unintentionally as
part of merge commits.

### DIFF
--- a/arm-software/ci/automerge.py
+++ b/arm-software/ci/automerge.py
@@ -169,7 +169,7 @@ def pr_exist_for_label(project_name: str, label: str) -> bool:
 def is_worktree_clean(git_repo: Git) -> bool:
     # `git status --porcelain` returns an empty result if worktree is clean
     status_output = git_repo.run_cmd(["status", "--porcelain"]).strip()
-    return len(status_output) > 0
+    return len(status_output) == 0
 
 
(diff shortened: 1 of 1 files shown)


arm-software/ci/automerge.py
```diff
@@ -166,6 +166,12 @@ def pr_exist_for_label(project_name: str, label: str) -> bool:
     return len(json.loads(gh_process.stdout)) > 0
 
 
+def is_worktree_clean(git_repo: Git) -> bool:
+    # `git status --porcelain` returns an empty result if worktree is clean
+    status_output = git_repo.run_cmd(["status", "--porcelain"]).strip()
+    return len(status_output) > 0
```
statham-arm, line 172: An empty result means the worktree is clean? But here you seem to be returning `True` if the result is _not_ empty.

statham-arm (changes requested):

arm-software/ci/automerge.py
pratlucas: 🤦 
Thanks for catching this! It should be fixed now.

PR #180, opened 2025-03-07
Use pull to update target branch in Automerge Workflow

pratlucas:
This updates the Automerge Workflow to use a `git pull --unshallow`
command instead of `git fetch --unshallow` to obtain the history of the
target branch. Using `git pull` ensures the worktree contents will also
be correctly up-to-date if any new changes are added to the remote
between the 'Checkout' step and the 'Fetch Branches' step.

### DIFF
--- a/.github/workflows/automerge.yml
+++ b/.github/workflows/automerge.yml
@@ -37,7 +37,7 @@ jobs:
         run: |
           git remote set-branches --add origin ${{ matrix.branches.from_branch }}
           git fetch origin ${{ matrix.branches.from_branch }}:${{ matrix.branches.from_branch }}
-          git fetch origin --update-head-ok --unshallow ${{ matrix.branches.to_branch }}:${{ matrix.branches.to_branch }}
+          git pull --unshallow --ff-only origin ${{ matrix.branches.to_branch }}
       - name: Run automerge
         run: python3 arm-software/ci/automerge.py --project-name ${{ github.repository }} --from-branch ${{ matrix.branches.from_branch }} --to-branch ${{ matrix.branches.to_branch }} --verbose
         env:


.github/workflows/automerge.yml
```diff
@@ -37,7 +37,7 @@ jobs:
         run: |
           git remote set-branches --add origin ${{ matrix.branches.from_branch }}
           git fetch origin ${{ matrix.branches.from_branch }}:${{ matrix.branches.from_branch }}
-          git fetch origin --update-head-ok --unshallow ${{ matrix.branches.to_branch }}:${{ matrix.branches.to_branch }}
+          git pull --unshallow --ff-only origin ${{ matrix.branches.to_branch }}
```
statham-arm, line 40: I agree that this looks sensible in the case that previously failed, when the checked-out branch is the same as `to_branch`.

But what happens if it isn't? The whole design of this YAML, and of the automerge script itself, looks as if it's designed to be able to handle multiple pairs of branches to merge from/to. I guess this "Fetch Branches" step will run these same three commands for each pair, one at a time.

I'm not sure what to recommend instead. One possibility is to `git checkout ${{ matrix.branches.to_branch }}` before the pull. But then maybe you end up with a weird branch checked out at the end of all this, and the "Run automerge" step runs the `automerge.py` from the wrong place.

Perhaps `git fetch --update-head-ok` is fine if it's also followed by some vigorous commands to put everything back the way it should be, starting with `git reset --hard` and maybe also `git clean -dfx`?

statham-arm: Ah, I see – so for _every_ branch pair, the "Fetch Branch" step will find that `to_branch` is already the checked-out branch, because the "Checkout" step two phases earlier checked it out?

In that case, fair enough.

(And I guess that means that every `to_branch` runs its _own_ version of `automerge.py`? Even though this workflow file is itself committed on multiple branches, and the workflow is presumably running only the one from `arm-software`?)

statham-arm (approved):

PR #181, opened 2025-03-10
[ATfE] Rebase picolibc patches

dcandler:
Upstream picolibc changes mean the first patch no longer applies cleanly. Only the context around our downstream modification has changed, so the patches have been rebased.


simpal01 (approved):
LGTM

PR #182, opened 2025-03-10
[ATfL] Set ARM_TOOLCHAIN_ID for the build

pawosm-arm:


schander-arm:
The problem is that we are not passing the variables used to generate the toolchain id to the docker image being used to build ATfL toolchain.

pawosm-arm:
Yes, I've created you a task to propagate these correctly into the docker image. It is rather urgent as we have a requirement that all Arm toolchains have this ID.

pawosm-arm:
Nevertheless, this still can be merged (and cherry-picked into the release branch), as we need it, and current solution can generate some ID (correctly starting with `L`) even if those variables aren't being passed.

schander-arm (approved):

PR #191, opened 2025-03-11
[ATfE] Add newline to boilerplate pattern for Corstone model

vhscampos:
The regular expression for the boilerplate pattern for Corstone's output has one extra newline that must be accounted for.


simpal01 (approved):
LGTM

dcandler:
Which tests did you find had a different output?

vhscampos:
> Which tests did you find had a different output?

One whose suffix is `cout.sh.cpp`. It is the one test that motivated the creation of a boilerplate text filtering.

PR #197, opened 2025-03-12
[ATfE] Remove pre-release package suffix

dcandler:
Setting the suffix to nothing will correctly generate the toolchain ID, but leads to an incorrectly formatted package name. Instead, the suffix should not be set at all.


pratlucas (approved):
LGTM.

PR #183, opened 2025-03-10
Add SBOM JSON files for ATfE, llvmlibc overlay and newlib overlay

vrukesh:
Add SBOM JSON files for ATfE, llvmlibc overlay and newlib overlay; and Python scripts which generate these SBOM JSON files.

PR #198, opened 2025-03-13
Add SBOM JSON files for ATfE, llvmlibc overlay and newlib overlay (#183)

vrukesh:
Add SBOM JSON files for ATfE, llvmlibc overlay and newlib overlay; and Python scripts which generate these SBOM JSON files.
This is cherry-pick from arm-software branch commit: e5cde593ba2c951c0df5be388524f7b518efc4b2


pratlucas:
Is this a cherry-pick from the `arm-software` branch? If so, can you add this information, as well as the reference for the original commit, to the commit message?

PR #212, opened 2025-03-19
Update SBOM JSON file for ATfE newlib overlay; and Python script which generates this SBOM JSON file.

vrukesh:
Update SBOM JSON file for ATfE newlib overlay; and Python script which generates this SBOM JSON file.


simpal01 (approved):
LGTM

PR #214, opened 2025-03-19
[ATfE] Don't delete temporary semihosting-features file

dcandler:
The NamedTemporaryFile used to create the `:semihosting-features` file for FVP testing will automatically attempt to delete itself when closed. Since this file needs to persist in order to replace any existing copy of the file, the auto deletion must be disabled.

The `os.rename` operation has also been changed to `os.replace` since renaming to an existing file will throw an error on some operating systems such as Windows.

### DIFF
--- a/arm-software/embedded/arm-runtimes/test-support/run_fvp.py
+++ b/arm-software/embedded/arm-runtimes/test-support/run_fvp.py
@@ -131,10 +131,16 @@ def run_fvp(
     # to prevent one process reading an incomplete file being written from
     # another. This is done by creating a temporary file and replacing.
     shfeatures_path = path.join(working_directory, ":semihosting-features")
-    with tempfile.NamedTemporaryFile(dir=working_directory, delete=False) as fh:
+    try:
+        fh = tempfile.NamedTemporaryFile(dir=working_directory, delete=False)
         fh.write(b"SHFB\x01")  # NamedTemporaryFile is binary already.
-        fh.flush()  # Ensure file is complete before renaming.
+        fh.close()
         os.replace(fh.name, shfeatures_path)
+    finally:
+        try:
+            os.remove(fh.name)
+        except FileNotFoundError:
+            pass
 
     result = subprocess.run(
         command,
--- a/arm-software/embedded/arm-runtimes/test-support/run_fvp.py
+++ b/arm-software/embedded/arm-runtimes/test-support/run_fvp.py
@@ -129,12 +129,12 @@ def run_fvp(
     # Since multiple tests can potentially be run concurrently in the same
     # working directory, the file should be created in an atomic operation
     # to prevent one process reading an incomplete file being written from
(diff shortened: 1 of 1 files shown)


pratlucas (dismissed):
LGTM

statham-arm:
Sorry – I don't understand. What do you mean by the _temporary_ file needing to persist?

The _final_ file name `:semihosting-features` needs to continue existing. But why would the file also need to still exist under its original temporary name? Nothing is going to be referring to that name.

It seems to me that the effect of this change is to leave more temp-file debris lying around, in the event that the rename fails. Can you explain the scenario in which the auto-deletion causes a problem?

statham-arm:
Ah, now I think I understand. The point is that on Windows, there's a special flag `O_TEMPORARY` that you can set when opening a file, which will make the _kernel_ automatically delete it on close. `tempfile.NamedTemporaryFile` works by setting that flag, which is reasonable in most cases, because it's more reliable than the application doing the cleanup by hand (it might crash and leave it undone). But the `O_TEMPORARY` flag is preserved by the rename / replace, so yes, `:semihosting-features` disappears when the context manager cleans up!

statham-arm:
And even in the Win32 API I can't find any way to _clear_ the delete-on-close attribute for a file handle you've already opened. Ugh. So I suppose we have to do _something_ ugly. But we could at least make our own attempt to clean up the mess.

How about this:
```python
try:
    fh = tempfile.NamedTemporaryFile(dir=working_directory, delete=False)
    fh.write(b"SHFB\x01")
    fh.close()
    os.replace(fh.name, shfeatures_path)
finally:
    try:
        os.remove(fh.name)
    except FileNotFoundError:
        pass
```

That way we still delete the temporary name if we possibly can (tolerating failure of the deletion because we successfully renamed it).

This is all quite horrible! Makes you wish there was a standard library facility to do it all for you and get it right on multiple OSes.

dcandler:
Yes, the problem I encountered on Windows was that the final renamed file was getting deleted at the point the temporary file was closed. Additionally, on a different Linux system I found the close function was instead attempting to delete the  temporary filename, which no longer existed after the rename and so caused an error.

I agree it's better to clean up the temporary files if we can. Having to deal with removing the file ourselves isn't a pretty solution, but having the `:semihosting-features` file is itself a workaround to another issue.

statham-arm (approved):
> Additionally, on a different Linux system I found the close function was instead attempting to delete the temporary filename, which no longer existed after the rename and so caused an error.

Ah, I'm not _very_ surprised by that. When I tested the previous version of the code I did vaguely recall that `NamedTemporaryFile`'s destructor would complain if the file had already gone away when it tried to delete it. But it worked OK when I tried it locally. Perhaps the Python standard library changed that behaviour in a newer version. In that case we're definitely safer with the explicit `except` clause in this version.

LGTM!

PR #220, opened 2025-03-20
[ATFE] Ensure the correct bootcode is set based on the test executor environment.

simpal01:
Set the correct bootcode based on the test executor environment when compiler-rt and CXX tests are enabled.


pratlucas (approved):
LGTM

PR #230, opened 2025-03-25
[ATfE] Add multilib.yaml mapping form Armv8-A big endian to Armv7-A big endian

vhscampos:
This mapping is necessary to make the multilib variant selection work for Armv8-A big endian, since the library variants supposed to be used in this case are chosen on the presence of the multilib flag corresponding to Armv7-A.

Besides this change, this commit introduces more testing to cover more cases of this mapping.


smithp35 (approved):
LGTM. Thanks for improving the number of test cases.

Some thoughts that are mostly out of scope of this particular patch. I'm thinking that we may be able to auto-generate this test from a simpler input as there's a lot of repetition involved. This may help when we need to update the expected value to account for new library variants, or layering.

vhscampos:
Yes, the current way is not scalable. But we ought to investigate if we can auto generate these tests while keeping them compatible with the lit infrastructure.

PR #234, opened 2025-03-27
[ATfE] Add multilib.yaml mapping form Armv8-A big endian to Armv7-A big endian

vhscampos:
This is a cherry-pick of #230.

This mapping is necessary to make the multilib variant selection work for Armv8-A big endian, since the library variants supposed to be used in this case are chosen on the presence of the multilib flag corresponding to Armv7-A.

Besides this change, this commit introduces more testing to cover more cases of this mapping.


smithp35 (approved):
LGTM for clean cherry-pick.

PR #235, opened 2025-03-27
Add make.bat for the llvmlibc sample

smithp35:
This adapts the existing make.bat files from the picolibc samples to the llvmlibc sample so that it can run on Windows.


voltur01 (approved):
LGTM and seems to match the Makefile.

PR #236, opened 2025-03-27
Add make.bat files for the newlib samples

smithp35:
This adapts the existing Makefile from the newlib samples to a make.bat so that the samples can be run on Windows.


voltur01 (approved):
LGTM, thank you!

Issue #243, opened 2025-04-01
[Downstream Change] [LLD][AArch64] Increase alignment of AArch64AbsLongThunk to 8 for 20.x release


smithp35:
The ATfE samples linkerscript raspi3.ld has `__boot_flash = 0x80000` and `__flash = 0x10000000` which requires a long branch thunk from the `_start` in the `__boot_flash` to the remainder of the startup code in `__flash`. As the MMU is not enabled at this point the long branch thunk cannot assume unaligned access.

In LLD we need the AArch64AbsLongThunk to have an alignment of 8 so that the ldr is always loading from an 8-byte aligned address. 

This is a cherry pick of https://github.com/llvm/llvm-project/pull/133738 to the arm-software/20.x release

smithp35:
It was pointed out to me that we are doing patch files for the 20 release rather than commit the change. https://github.com/arm/arm-toolchain/pull/247 adds the patch file.

PR #247, opened 2025-04-01
[Downstream Change][LLD][AArch64] Increase alignment of AArch64AbsLon…

smithp35:
This permits an AArch64AbsLongThunk to be used in an environment where unaligned accesses are disabled.

The AArch64AbsLongThunk does a load of an 8-byte address. When unaligned accesses are disabled this address must be 8-byte aligned.

The vast majority of AArch64 systems will have unaligned accesses enabled in userspace. However, after a reset, before the MMU has been enabled, all memory accesses are to "device" memory, which requires aligned accesses. In systems with multi-stage boot loaders a thunk may be required to a later stage before the MMU has been enabled.

As we only want to increase the alignment when the ldr is used we delay the increase in thunk alignment until we know we are going to write an ldr. We also need to account for the ThunkSection alignment increase when this happens.

In some of the test updates, particularly those with shared CHECK lines with position independent thunks it was easier to ensure that the thunks started at an 8-byte aligned address in all cases.

This is a cherry-pick of upstream llvm-project PR 133738 to the arm-software release/20.x branch.

Downstream issue: #243

### DIFF
--- a/arm-software/embedded/patches/llvm-project/0016-Increase-alignment-of-AArch64AbsLongThunk-to-8.patch
+++ b/arm-software/embedded/patches/llvm-project/0016-Increase-alignment-of-AArch64AbsLongThunk-to-8.patch
@@ -1,0 +1,543 @@
+From 061b2de3ac5500d3b018508a826bb1283aef24b6 Mon Sep 17 00:00:00 2001
+From: Peter Smith <peter.smith@arm.com>
+Date: Tue, 1 Apr 2025 09:49:27 +0100
+Subject: [PATCH] Increase alignment of AArch64AbsLongThunk to 8
+
+This permits an AArch64AbsLongThunk to be used in an environment where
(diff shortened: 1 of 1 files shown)


dcandler (approved):
Since the change is now made in a patch file, it no longer needs tracking as a downstream change. That means the tag in the title is no longer necessary, nor the link in the commit message to https://github.com/arm/arm-toolchain/issues/243 which could also be closed now.

The actual patch itself LGTM; the commit message can be tweaked before merging.

smithp35:
> Since the change is now made in a patch file, it no longer needs tracking as a downstream change. That means the tag in the title is no longer necessary, nor the link in the commit message to #243 which could also be closed now.
> 
> The actual patch itself LGTM; the commit message can be tweaked before merging.

Thanks, have edited those parts out of the message.

PR #250, opened 2025-04-03
[ATfE] v8-M library naming and selection issues for strict alignment builds.

simpal01:
We don’t have a -mno-unaligned-access specific variant for the v8M libraries. For v8-M with strict aligned access, it’s more appropriate to use the v7-M variants, as they include a version with strict alignment support. It was a deliberate decision to provide strict-alignment variants only for a subset of architectures.

To support this, both -mno-unaligned-access and -munaligned-access options must 
be passed into the multilib selection logic.  This aligns with the approach in 
https://github.com/llvm/llvm-project/pull/134099/, which also considers the target architecture when determining the alignment access policy. Since Armv6-M does not support unaligned accesses, test changes are needed to reflect the updated behaviour when the target architecture is taken into account.

This patch fixes below issues.

1. The naming of the v8-M libraries should explicitly include ‘unaligned’ to reflect their behaviour.

2. When using -mno-unaligned-access, the multilib system should instead select the v7-M variant.

3. Armv6m test changes are needed to reflect the updated behaviour when the target architecture is taken into account.


arm-software/embedded/test/multilib/armv6m.test
```diff
@@ -1,13 +1,13 @@
 # RUN: %clang -print-multi-directory --target=armv6m-none-eabi -mfpu=none | FileCheck %s --check-prefix=CHECK-NOFP-EXN-RTTI-UNALIGNED
```
vhscampos, line 1: Change CHECK-NO-FP-EXN-RTTI-UNALIGNED to CHECK-NO-FP-EXN-RTTI. That is, remove the "-UNALIGNED" suffix.

arm-software/embedded/test/multilib/armv6m.test
```diff
@@ -1,13 +1,13 @@
...
-# CHECK-NOFP-EXN-RTTI-UNALIGNED: arm-none-eabi/armv6m_soft_nofp_exn_rtti_unaligned{{$}}
+# CHECK-NOFP-EXN-RTTI-UNALIGNED: arm-none-eabi/armv6m_soft_nofp_exn_rtti{{$}}
 # CHECK-NOFP-EXN-RTTI-UNALIGNED-EMPTY:
 
 # RUN: %clang -print-multi-directory --target=armv6m-none-eabi -mfpu=none -mno-unaligned-access | FileCheck %s --check-prefix=CHECK-NOFP-EXN-RTTI
 # CHECK-NOFP-EXN-RTTI: arm-none-eabi/armv6m_soft_nofp_exn_rtti{{$}}
 # CHECK-NOFP-EXN-RTTI-EMPTY:
 
 # RUN: %clang -print-multi-directory --target=armv6m-none-eabi -mfpu=none -fno-exceptions -fno-rtti | FileCheck %s --check-prefix=CHECK-NOFP-UNALIGNED
```
vhscampos, line 9: Change CHECK-NO-FP-UNALIGNED to CHECK-NO-FP. That is, remove the "-UNALIGNED" suffix.

vhscampos (approved):

PR #252, opened 2025-04-04
Add minimal libc++ support with llvmlibc

voltur01:
Particular missing piece is iostreams. libc++ testing remains disabled.


smithp35 (commented):
Code changes LGTM

Could we update https://github.com/arm/arm-toolchain/blob/arm-software/arm-software/embedded/docs/llvmlibc.md with the new status? That page currently says that we provide no C++ library support.

voltur01:
Thanks, I will update the docs now!

PR #253, opened 2025-04-04
Add llvmlibc C++ sample and align with other samples

voltur01:
This adds a sample to demonstrate libc++ support with llvmlibc. It duplicates the startup files, hopefully these will be moved to llvmlibc-support soon or otherwise may be moved out into a shared place for both samples to use.

It also updates the existing C sample to use the relative paths and Makefile variables as other samples - external BIN_PATH did not work previously, only in place builds within the toolchain folder.


vrukesh (approved):
LGTM !

PR #256, opened 2025-04-07
[ATfE] Use normalized 4-field triple for LLVM_DEFAULT_TARGET_TRIPLE

dcandler:
After upstream changes in #122629, omitting the vendor in the default target triple will cause the driver to incorrectly normalize it in certain tests. This patch sets the default to the normalized version to avoid the failures.


pratlucas (approved):
LGTM.

PR #263, opened 2025-04-09
[ATfE] Remove unaligned variants of Armv6-M

vhscampos:
The Armv6-M architecture does not support unaligned memory accesses, so these variants are pretty much identical to their aligned counterparts.

This patch removes the unaligned variants and adjust the multilib flags for compatibility.


simpal01 (approved):
LGTM

Issue #259, opened 2025-04-09
[Downstream Change][LV] wrf in SPEC2017 emits "Verification Error" on Grace when using llvmorg-20.1.0-rc1


pawosm-arm:
This is an issue reported upstream in this bug report [1]. The upstream fix [2] would require pulling-in at least 16 other commits and still would require reautogeneration of some of the test cases. Alternative proposal is to revert or block the affected code.

[1] https://github.com/llvm/llvm-project/issues/126836
[2] https://github.com/llvm/llvm-project/pull/132691

pawosm-arm:
Fixed by PR #260
Alternative available as PR #264

pawosm-arm:
The upstream change is of more significance than our downstream hack. Therefore #260 should be reverted after the relevant upstream change is merged.

PR #264, opened 2025-04-09
[Downstream change][LV] Backported patchset with the fix for the WRF issue

pawosm-arm:
Downstream issue: https://github.com/arm/arm-toolchain/issues/259


pawosm-arm:
It forced upstream to do a proper backport, so I'm closing this one.

PR #270, opened 2025-04-11
Add LLVM libc based toolchain build

voltur01:
Add build script for ATfE variant with LLVM libc as the defaulty C library to simplify testing project compatibility with LLVM libc.

Update CMake file to give the package an informative name.


arm-software/embedded/scripts/build_llvmlibc_toolchain.sh
```diff
@@ -0,0 +1,26 @@
...
+# The script creates a build of the toolchain in the 'build_llvmlibc_toolchain'
+# directory, inside the repository tree.
+
+set -ex
+
+export CC=clang
+export CXX=clang++
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+REPO_ROOT=$( git -C ${SCRIPT_DIR} rev-parse --show-toplevel )
```
statham-arm, line 19: Shell quoting nits: `${SCRIPT_DIR}` in this line should be double-quoted. So should `${BUILD_DIR}` in the `mkdir` and `cd` commands below.

(But it's OK not to quote these `$(...)` constructs on the RHS of an assignment, or the `${REPO_ROOT}` in the assignment to `BUILD_DIR`, because in that context, the shell decides what's the RHS of a variable assignment _before_ expanding those things into stuff that might contain spaces.)

statham-arm (commented):
Since @voltur01 is on holiday this week and I'm on holiday next week, in the interests of moving things along, I'll change my mind and approve this PR, then submit my own followup which fixes the quoting, in this file and several others with the same issue.

statham-arm (approved):

voltur01:
Thank you @statham-arm! Note that we have the same code in other build scripts, so may need to review all of them.

PR #278, opened 2025-04-14
[ATfL] OMPT support should not be built at that stage

pawosm-arm:

It turned out that libomp.a that we are shipping is useless.

It should not support OMPT as it tries to dlopen() libarcher.so

We are already building libomp.so with OMPT support enabled along with libarcher.so in the previous stage.

Fully operational libomp.a is needed by Allinea Smoke Tests.

With OMPT disabled, -ldl is not needed for OpenMP tests to pass, so the config file hack could be removed from build.sh.


kiranchandramohan (approved):
LG

PR #280, opened 2025-04-14
[ATfL] Restore obligatory linking with -ldl. Apparently this is needed for RHEL8

pawosm-arm:
In my previous commit I've misread the reason why those two lines were introduced. Now we cannot build for RHEL8 and this patch restores those two lines.


kiranchandramohan (approved):
LG

PR #282, opened 2025-04-15
[ATfL] Remove Installation.md as we have no control over its content

pawosm-arm:
The URLs in this file point into internal machines and are invalid outside Arm. We will not know the correct URLs until 5mins before the release. As we cannot control it, we should not take responsibility for it.

Also, shipping a document about installing debs/rpms inside of debs/rpms is meaningless.


DaftanoPro (approved):

pawosm-arm:
I've read Kiran's comment on our communicator today and his rationale behind keeping this file in this repo is sound.

The more I think about it, the more I see that we want to fix one wrong with another. It is wrong to remove this file, and we doing it because we can't have correct URL. And this is also wrong, or even worse. At this stage of development, the decision about the URL should be already settled and shared among teams.

PR #286, opened 2025-04-16
[Downstream change][TargetParser] Parse Amazon Linux triple and handle RuntimeLibcalls

MacDue:
This parses the "aarch64-amazon-linux" triple and updates RuntimeLibcalls to mark the sincos* functions as available on Amazon Linux. This allows these functions to be vectorized on Amazon Linux.

Downstream change: #87


MacDue:
I believe this patch may be the correct fix for upstream too (but I've not made a PR there yet as this fix won't be backported to the LLVM 20 release upstream, so is less time sensitive there).

kiranchandramohan:
Is this different in spirit from https://github.com/llvm/llvm-project/pull/109263 ? Is it only in the driver that new triples are not allowed? Is it safe to add in llvm.

Will be good to have this reviewed upstream also.

MacDue:
> Is this different in spirit from [llvm/llvm-project#109263](https://github.com/llvm/llvm-project/pull/109263) ? Is it only in the driver that new triples are not allowed? Is it safe to add in llvm.
> 
> Will be good to have this reviewed upstream also.

AFAIK this is not the same as  [llvm/llvm-project#109263](https://github.com/llvm/llvm-project/pull/109263) , which IIUC is modifying the clang driver to find a GNU install located at a path which contains a `aarch64-amazon-linux` triple automatically. This patch just parses the triple and updates the `RuntimeLibcalls` to acknowledge that Amazon Linux has an implementation of `sincos`.

pawosm-arm (dismissed):

MacDue:
> Will be good to have this reviewed upstream also.

Created an upstream version of this PR: https://github.com/llvm/llvm-project/pull/136114

pawosm-arm (approved):

kiranchandramohan:
Thanks @MacDue for this work. We are going for the ATfL 20 release without this fix. For ATfL and LLVM 21 we will decide on an appropriate fix after discussing with AWS.

PR #295, opened 2025-04-22
[ATfL] Copy top-level LLVM's LICENSE.TXT into our tarball

pawosm-arm:
Apparently, `make install` doesnt't do that


kiranchandramohan (approved):
LG.

PR #298, opened 2025-04-23
[ATfE] Add required strictly aligned variants for Armv8.1-M Main

vhscampos:
Two non-PACBTI variants needed a strictly aligned counterpart:

 - armv8.1m.main_hard_nofp_mve
 - armv8.1m.main_hard_nofp_mve_exn_rtti

because no suitable Armv7-M variant exists for these two as a fallback option. The combination of hard FP ABI and no FPU is not covered by our Armv7-M variants.

On the PACBTI side, all the existing variants needed their strictly aligned counterparts. This is because it's not safe to fallback to Armv7-M variants in this case, as these are not built for PACBTI.

This patch introduces new strictly aligned variants for Armv8.1-M Main. Two of them are for non-PACBTI + hard FP ABI + no FPU, while the rest is for all the PACBTI variants.


simpal01 (approved):
LGTM

PR #299, opened 2025-04-23
[ATfE] Fix string compare

dcandler:
STREQUAL should be used to compare a string instead of EQUAL, otherwise the short package name will never be used.


voltur01 (approved):
Thank you, LGTM.

PR #314, opened 2025-05-02
[ATfE] Re-enable LLVM-libc for nofp variants

saturn691:
After https://github.com/llvm/llvm-project/pull/137592 is merged and the patch is collected downstream, these variants for LLVM-libc can be reenabled.


pratlucas (approved):
LGTM

PR #319, opened 2025-05-07
[ATfE] Accept arm64 as well as aarch64 machine names in FVP scripts

dcandler:
The scripts to download and run the FVPs use `uname` to choose between the Linux x86 and AArch64 packages. These expect `aarch64` as a return, but `arm64` is also possible. The FVP platform should be the same for both.


vhscampos (approved):

dcandler:
Abandoning in favor of https://github.com/arm/arm-toolchain/pull/320

PR #320, opened 2025-05-08
[ATfE] Only check machine name when run_fvp.py runs

dcandler:
Since the run_fvp.py script is imported into other scripts, the machine name check can be carried out on machines where FVP testing is not supported and disabled.

This patch moves the check inside the run function, so that it the machine name only needs to be matched when FVP tests are actually being executed.

### DIFF
--- a/arm-software/embedded/arm-runtimes/test-support/run_fvp.py
+++ b/arm-software/embedded/arm-runtimes/test-support/run_fvp.py
@@ -22,35 +22,49 @@ class FVP:
     cmdline_param: str
 
 
-# The location of files within the FVP install directory can differ
-# between the packages for different platforms.
-uname_machine = uname().machine.lower()
-if uname_machine == "x86_64":
-    platform_dir = "Linux64_GCC-9.3"
-elif uname_machine == "aarch64":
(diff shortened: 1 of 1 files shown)


arm-software/embedded/arm-runtimes/test-support/run_fvp.py
```diff
@@ -69,7 +83,17 @@ def run_fvp(
...
     if fvp_model not in MODELS:
         raise Exception(f"{fvp_model} is not a recognised model name")
-    model = MODELS[fvp_model]
+
+    # The location of files within the FVP install directory can differ
+    # between the packages for different platforms.
+    uname_machine = uname().machine.lower()
+    if uname_machine == "x86_64":
+        fvp_platform = "x86"
+    elif uname_machine == "aarch64":
```
HugoSilvaSantos, line 92: Don't we need to add "arm64" to support the Mac studios?

dcandler: The tests are not expected to be run on that platform, so for now it should not be necessary to match the other machine name. Instead the name is not checked if not running the tests.

HugoSilvaSantos (approved):
LGTM

PR #322, opened 2025-05-12
Remove unused CMake Meson files

voltur01:
These files were moved to the arm-runtimes folder.


voltur01:
This was pointed out in LLVM ET https://github.com/ARM-software/LLVM-embedded-toolchain-for-Arm/pull/642

statham-arm (approved):
Indeed, the cmake scripts that use them are in the `arm-runtimes` directory, and refer to these files relative to `${CMAKE_CURRENT_SOURCE_DIR}`, so it's clear that the versions in that directory will be the ones used, and not these.

PR #324, opened 2025-05-12
[ATfE] Allow for a user specified system name when packaging

dcandler:
If a system name for the package other than the CMake default is required, this can now be set through the `PACKAGE_SYSTEM_NAME` option. If unset, the default remains the CMake system name.

Additionally, the `build.sh` script has been modified so that additional CMake arguments can be passed down to CMake during configuration. `EXTRA_CMAKE_ARGS` can now be read as an environment variable, which will be appended to the arguments in the script.


pratlucas (approved):

HugoSilvaSantos (approved):
LGTM

PR #325, opened 2025-05-13
[ATfE] Fix path to the root of the repo when applying llvm patch

stuij:


pratlucas (dismissed):

arm-software/embedded/docs/building-from-source.md
```diff
@@ -93,7 +93,7 @@ export CXX=clang++
 mkdir repos
 git -C repos clone https://github.com/picolibc/picolibc.git
 git -C repos/picolibc am -k "$PWD"/patches/picolibc/*.patch
-git -C arm-toolchain am -k "$PWD"/patches/llvm-project/*.patch
+git -C ../.. am -k "$PWD"/patches/llvm-project/*.patch
```
simpal01, line 96: The original change looks correct to me. Using relative paths in this way can be confusing, especially for those unfamiliar with the directory structure. A clearer reference could make it easier to follow something like:
git -C /path/to/arm-toolchain

stuij: I have a couple of issues with the way this is done now:

- This is a set of instructions for people to use to create their own toolchain. I use this myself to just copy-paste them in the embedded directory, and everything just works. Putting in `arm-toolchain` or `path/to/arm-toolchain` will break this use case.
- It also confused me, because I reason about `arm-toolchain` from the perspective of someone who expects this snippet to work. So I'm assuming that `arm-toolchain` should be in the path, but isn't. Something must be broken. If I didn't know the toolchain quite well I wouldn't immediately understand what the issue was. `/path/to/arm-toolchain` would be clearer than `arm-toolchain`, but I would still wonder why you wouldn't just supply working commands. And so I'd wonder if I was missing something. Is it another toolchain? Is it a typo?

So yea, to me working code just makes more sense. And `two dirs down` is clear enough I think with the context of what you're doing. And a quick `(cd ../..; pwd)` or equivalent can quickly dispel any questions.

simpal01: ok makes sense. Can you add a line saying "The following commands assume you are in the arm-toolchain/arm-software/embedded directory" at line number 89.

stuij: Added it, although I put it a bit higher and changed the wording slightly to cover the other code block as well.

simpal01: Thank you changes LGTM

simpal01 (approved):
LGTM

PR #326, opened 2025-05-13
[ATFE] Configure ATFE to exclude libc-provided AEABI builtins from compiler-rt

simpal01:
We don't want two implementations of the __aeabi runtime in two different libraries. We could suppress the compiler-rt implementation at build time so there's only the picolibc definition. This patch https://github.com/llvm/llvm-project/pull/137952 introduced a new cmake flag called
COMPILER_RT_EXCLUDE_LIBC_PROVIDED_ARM_AEABI_BUILTINS to support this functionality.

When enabled, this flag excludes the following ARM AEABI memory function implementations from the compiler-rt build: 

- __aeabi_memcmp
- __aeabi_memset
- __aeabi_memcpy
- __aeabi_memmove

These functions are already provided by standard C libraries like newlib, and picolibc, so excluding them avoids duplicate symbol definitions and reduces unnecessary code duplication.

Also we can now remove the downstream patch that changed the library order to put -lc before compiler-rt builtins as there is now this recent upstream addition of the
COMPILER_RT_EXCLUDE_LIBC_PROVIDED_ARM_AEABI_BUILTINS flag.


dcandler (approved):
LGTM

PR #328, opened 2025-05-14
Add an `elf2bin` utility.

statham-arm:
This tool reads ELF image files, paying attention only to the 'segment' view of the file presented by the program header table, and ignoring the section header table completely (including not even noticing whether one exists).

It extracts the contents of the loadable segments of the file, and outputs them in various simple forms: plain binary files, two well-known hex file formats (Intel Hex and Motorola S-records), and the much simpler 'Verilog hex' format which is just a binary file with each byte translated into a two-digit hex number.

Additional features include the ability to restrict which segments are extracted; the choice of whether to output one binary file per segment or a single one that combines all segments at their correct relative offsets with padding in between; and the ability to distribute the output bytes across multiple output files suitable for loading into a banked ROM setup in which (for example) each of four ROMs holds only the bytes living at addresses with a particular residue mod 4.

In other words, this tool provides essentially all the same functionality that Arm Compiler's `fromelf` did, in its binary and hex output modes.

### DIFF
--- a/arm-software/shared/elf2bin/Opts.td
+++ b/arm-software/shared/elf2bin/Opts.td
@@ -1,0 +1,64 @@
+// Command-line option definitions for elf2bin
+//
+// Copyright (c) 2022-2025, Arm Limited and affiliates.
+//
+// Part of the Arm Toolchain project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+
+include "llvm/Option/OptParser.td"
+
+multiclass Value<string longname, string shortname, string metavar,
+                 string help, OptionGroup group> {
+  def NAME # _EQ : Joined<["--"], longname # "=">,
+                   HelpText<help>,
+                   MetaVarName<metavar>,
+                   Group<group>;
+  def : Separate<["--"], longname>,
+        Alias<!cast<Joined>(NAME # "_EQ")>;
+
+  if !not(!empty(shortname)) then {
+    def : JoinedOrSeparate<["-"], shortname>,
+          Alias<!cast<Joined>(NAME # "_EQ")>,
+          HelpText<"Alias for --" # longname>, Group<group>;
+  }
+}
+
+multiclass NoValue<string longname, string shortname, string help,
+                   OptionGroup group> {
+  def NAME : Flag<["--"], longname>,
+             HelpText<help>,
+             Group<group>;
+
+  if !not(!empty(shortname)) then {
+    def : Flag<["-"], shortname>,
+          Alias<!cast<Flag>(NAME)>,
+          HelpText<"Alias for --" # longname>;
+  }
+}
+
+def grp_mode: OptionGroup<"mode">, HelpText<"FORMAT OF OUTPUT">;
+def grp_file: OptionGroup<"output">, HelpText<"LOCATION OF OUTPUT">;
+def grp_opts: OptionGroup<"options">, HelpText<"OPTIONS">;
+
+defm output_file: Value<"output-file", "o", "FILENAME", "Name of the output file. Invalid if more than one output file is to be generated.", grp_file>;
+defm output_pattern: Value<"output-pattern", "O", "PATTERN", "Schema for naming all output files. May contain %f for base name of input file; %F for full name of input file; %a or %A for base address of ELF segment, in hex or in HEX; %b for bank number, if using --banks; %% for a literal % sign.", grp_file>;
+
+defm ihex: NoValue<"ihex", "", "Output Intel Hex files (\":108000\" style)", grp_mode>;
+defm srec: NoValue<"srec", "", "Output Motorola Hex files (\"S3150800\" style)", grp_mode>;
+defm bin: NoValue<"bin", "", "Output a binary file per ELF segment", grp_mode>;
+defm bincombined: NoValue<"bincombined", "", "Output a single binary file, containing all segments, with padding to place them at the correct relative positions", grp_mode>;
+defm vhx: NoValue<"vhx", "", "Output a Verilog Hex file per ELF segment", grp_mode>;
+defm vhxcombined: NoValue<"vhxcombined", "", "Output a single Verilog Hex file, containing all segments, with padding to place them at the correct relative positions", grp_mode>;
+
+defm base: Value<"base", "", "ADDRESS", "Base address of the whole output file, for --bincombined or --vhxcombined", grp_opts>;
+defm banks: Value<"banks", "", "WIDTHxNUM", "Partition the output into subfiles by writing WIDTH bytes in turn to each of NUM files in cyclic order, for binary or Verilog Hex output", grp_opts>;
+defm datareclen: Value<"datareclen", "", "LENGTH", "Number of data bytes to put into each record, for Intel Hex or Motorola Hex output", grp_opts>;
+defm segments: Value<"segments", "", "ADDRESS,ADDRESS,...", "Select only the ELF segments of the input which start at the specified addresses", grp_opts>;
+defm zi: NoValue<"zi", "", "Include the zero-initialized padding at the end of each ELF segment", grp_opts>;
+defm physical: NoValue<"physical", "", "Use the physical addresses (p_paddr) of ELF segments", grp_opts>;
+defm virtual: NoValue<"virtual", "", "Use the virtual addresses (p_vaddr) of ELF segments", grp_opts>;
+
+defm help: NoValue<"help", "", "Display this help", grp_opts>;
+defm version: NoValue<"version", "", "Display the version", grp_opts>;
--- a/arm-software/shared/elf2bin/bin.cpp
+++ b/arm-software/shared/elf2bin/bin.cpp
@@ -1,0 +1,330 @@
+// Code to write binary and VHX output for elf2bin
+//
+// Copyright (c) 2022-2025, Arm Limited and affiliates.
+//
+// Part of the Arm Toolchain project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+
+#include "elf2bin.h"
+
+#include <algorithm>
+#include <cassert>
+#include <memory>
+#include <queue>
+
+#include "llvm/ADT/StringRef.h"
+#include "llvm/ADT/bit.h"
+#include "llvm/Support/raw_ostream.h"
+
+using namespace llvm;
+
+// For classes below that process input data in manageably-sized chunks
+constexpr size_t CHUNKSIZE = 65536;
+
+/*
+ * Abstraction that provides a means of getting binary data from
+ * somewhere. In the simple case this will involve reading data from a
+ * segment in an ELF file. In more complex cases there might also be
+ * zero-byte padding, or one of these stream objects filtering out the
+ * even-index bytes of another.
+ */
+class BinaryDataStream {
+public:
+  // Returns a string of data, in whatever size is convenient. (But
+  // the size should be bounded, so that other streams filtering this
+  // one don't have to swallow a whole file in one go.)
+  //
+  // An empty returned string means EOF.
+  virtual StringRef read() = 0;
+  virtual ~BinaryDataStream() = default;
+};
+
+/*
+ * BinaryDataStream implementation that reads data from an ELF image.
+ */
+class ElfSegment : public BinaryDataStream {
+  InputObject &inobj;
+  size_t position, remaining;
+
+  StringRef read() override {
+    size_t readlen = std::min<size_t>(remaining, CHUNKSIZE);
+    size_t readpos = position;
+    position += readlen;
+    remaining -= readlen;
+    return inobj.membuf->getBuffer().substr(readpos, readlen);
+  }
+
+public:
+  ElfSegment(InputObject &inobj, size_t offset, size_t size)
+      : inobj(inobj), position(offset), remaining(size) {}
+};
+
+/*
+ * BinaryDataStream implementation that generates zero padding.
+ */
+class Padding : public BinaryDataStream {
+  size_t remaining;
+  char buffer[CHUNKSIZE];
+
+  StringRef read() override {
+    size_t readlen = std::min<size_t>(remaining, CHUNKSIZE);
+    remaining -= readlen;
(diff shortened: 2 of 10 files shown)


smithp35 (commented):
Would it be possible to add some information to https://github.com/arm/arm-toolchain/blob/arm-software/arm-software/docs/arm-toolchain-features.md

Which has a list of additional features available in arm-toolchain. I know elf2bin is not strictly a new feature, more of a standalone new program, but I think it counts as something not available upstream.

I think it could be worth a paragraph saying at least what elf2bin is and to run --help for usage information.

statham-arm:
I can do better than that – I found some actual documentation I'd forgotten already having written! Added a reference, and also the relevant parts of the documentation I already had. 

(I cut out material that only made sense when elf2bin lived in an experimental standalone repo, like how to build and test it, because now that's all done by the main toolchain build scripts.)

arm-software/shared/elf2bin/Opts.td
```diff
@@ -0,0 +1,64 @@
...
+defm bin: NoValue<"bin", "", "Output a binary file per ELF segment", grp_mode>;
+defm bincombined: NoValue<"bincombined", "", "Output a single binary file, containing all segments, with padding to place them at the correct relative positions", grp_mode>;
+defm vhx: NoValue<"vhx", "", "Output a Verilog Hex file per ELF segment", grp_mode>;
+defm vhxcombined: NoValue<"vhxcombined", "", "Output a single Verilog Hex file, containing all segments, with padding to place them at the correct relative positions", grp_mode>;
+
+defm base: Value<"base", "", "ADDRESS", "Base address of the whole output file, for --bincombined or --vhxcombined", grp_opts>;
+defm banks: Value<"banks", "", "WIDTHxNUM", "Partition the output into subfiles by writing WIDTH bytes in turn to each of NUM files in cyclic order, for binary or Verilog Hex output", grp_opts>;
+defm datareclen: Value<"datareclen", "", "LENGTH", "Number of data bytes to put into each record, for Intel Hex or Motorola Hex output", grp_opts>;
+defm segments: Value<"segments", "", "ADDRESS,ADDRESS,...", "Select only the ELF segments of the input which start at the specified addresses", grp_opts>;
```
smithp35, line 58: Could mention that we look for p_vaddr or p_paddr depending on physical/virtual

arm-software/shared/elf2bin/Opts.td
```diff
@@ -0,0 +1,64 @@
...
+defm zi: NoValue<"zi", "", "Include the zero-initialized padding at the end of each ELF segment", grp_opts>;
+defm physical: NoValue<"physical", "", "Use the physical addresses (p_paddr) of ELF segments", grp_opts>;
```
smithp35, line 60: Could it be worth a single option that takes physical or virtual as a parameter rather than two options. I'm sure we don't support both at the same time.

arm-software/shared/elf2bin/elf2bin.cpp
```diff
@@ -0,0 +1,468 @@
...
+    if (!elf)
+      fatal(infile, "unable to process: not an ELF file");
+    inobj.elf = std::move(elf);
+  }
+
+  /*
+   * Helper function for listing the segments of a file, paying
+   * attention to the --segments option to restrict to a subset.
+   */
+  auto segments = [&](InputObject &inobj) {
```
smithp35, line 337: Do we check that the ELF file isn't relocatable and error out with something like "ELF file must be type ET_EXEC or ET_DYN"?

arm-software/shared/elf2bin/elf2bin.cpp
```diff
@@ -0,0 +1,468 @@
...
+    }
+    outfiles.insert(conv.outfile);
+  }
+
+  uint64_t bankmod = nbanks * bankwidth;
+
+  for (const auto &conv : convs) {
+    switch (format.value()) {
+    case Format::BinMultifile:
+      bin_write(*conv.inobj, conv.outfile, conv.fileoffset.value(),
```
smithp35, line 432: The functions have a lot of shared parameters, could it be worth passing conv in as a "context" parameter.

Alternatively we should be doing just one of the options in this switch for each of the convs. Could we extend conv to be a transformation object that includes a write() member function?

arm-software/shared/elf2bin/test/makeelf.py
```diff
@@ -0,0 +1,345 @@
+#!/usr/bin/env python3
```
smithp35, line 1: Isn't this very similar to yaml2obj? 

If we're able to guarantee that the llvm tools build then perhaps the test suite can use yaml2obj and not need this?

arm-software/shared/elf2bin/test/test.py
```diff
@@ -0,0 +1,1339 @@
+#!/usr/bin/env python3
```
smithp35, line 1: Apologies for being extremely annoying as it would have been a lot of work to put this together.

Is there a reason we couldn't use llvm-lit, yaml2obj and elf2bin and run these like normal lit tests?

smithp35 (commented):
I've got as far as elf2bin.h on first pass. Mostly concentrating on things that jumped out rather than going through the code in detail looking for problems.

Mostly stylistic, subjective comments so far. I think the overall architecture of the program is fine.

My comment on choice of testing infrastructure is the one with the most impact on the PR. That will likely be worth discussing with the team to see if there's a consensus on method.

arm-software/shared/elf2bin/elf2bin.cpp
statham-arm: Not explicitly, but we check there's a program header table, which surely has the same effect.

arm-software/shared/elf2bin/Opts.td
statham-arm: Indeed not, but I think any other approach involves more typing: `--no-virtual` to contradict `--virtual`, or `--address=physical` vs `--address=virtual`. Brevity is useful.

arm-software/shared/elf2bin/bin.cpp
```diff
@@ -0,0 +1,330 @@
...
+
+/*
+ * BinaryDataStream implementation that reads data from an ELF image.
+ */
+class ElfSegment : public BinaryDataStream {
+  InputObject &inobj;
+  size_t position, remaining;
+
+  StringRef read() override {
+    size_t readlen = std::min<size_t>(remaining, CHUNKSIZE);
```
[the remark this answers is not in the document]
statham-arm: Oh yes. Must have been copy-paste from `hex.cpp`, where the similar construction is needed to disambiguate the output type, because there's a `uint64_t` and a `size_t`.

arm-software/docs/elf2bin.md
```diff
@@ -0,0 +1,172 @@
...
+input file containing segments at addresses 0x8000, 0x20000 and
+0x10000000, then `--segments 0x8000,0x10000000` would skip the middle
+one. This option applies to all output modes.
+
+Normally, `elf2bin` treats each segment as containing only the bytes
+actually stored in the ELF file. That is, the segment is treated as
+having length corresponding to its `p_filesz` field, not its
+`p_memsz`. You can use `--zi`, in any mode, to tell `elf2bin` to
+include zero padding after each segment to bring it up to its
+`p_memsz` length.
```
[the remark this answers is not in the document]
statham-arm: Another good catch! I must have added that option after writing this early version of the documentation.

arm-software/docs/elf2bin.md
```diff
@@ -0,0 +1,172 @@
...
+  how to name multiple output files.
+
+* Optionally, other options to adjust behavior.
+
+### Output modes
+
+The available output-mode options are:
+
+* `--ihex`: Intel hex format (the one where lines begin with '`:`').
+  `elf2bin`'s take on this format is similar to `fromelf`'s: it always
```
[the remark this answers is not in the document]
statham-arm: Good catch. I've added a couple of extra paragraphs saying that.

arm-software/shared/elf2bin/bin.cpp
```diff
@@ -0,0 +1,330 @@
...
+  std::error_code error;
+  llvm::raw_fd_ostream ofs(outfile, error);
+  if (error)
+    fatal(outfile, "unable to open", errorCodeToError(error));
+
+  while (true) {
+    StringRef data = st.read();
+    if (data.empty())
+      return;
+    for (uint8_t c : data) {
```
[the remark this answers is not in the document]
statham-arm: I'm sure it would, but I don't know that it's _necessary_. It's true that there's a small amount of duplicated code here. But it's not _always_ a readability improvement to turn two easy-to-understand functions into a single one with a parameter, because it means you have to understand what _both_ functions do to make sense of it.

If the amount of duplicated code were more like 20 lines than 4, I think there'd be a better case for this.

arm-software/shared/elf2bin/bin.cpp
```diff
@@ -0,0 +1,330 @@
...
+    // Set pos to be (-firstres), so that we'll skip the right
+    // number of bytes before the first one we return.
+    //
+    // Written as (1 + ~firstres) to avoid Visual Studio
+    // complaining about negating an unsigned.
+    pos = (1 + ~firstres) & mask;
+  }
+};
+
+static void bin_write(BinaryDataStream &st, const std::string &outfile) {
```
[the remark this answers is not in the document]
statham-arm: I don't think it's that important. (And I'm one of the stubborn holdouts these days who types all my function names by hand instead of letting language servers autocomplete them for me – surely more up-to-date developers would be even _less_ likely to need to remember the detailed spelling of everything.)

arm-software/shared/elf2bin/bin.cpp
```diff
@@ -0,0 +1,330 @@
...
+combined_prepare(InputObject &inobj, const std::vector<Segment> &segments_orig,
+                 bool include_zi, std::optional<uint64_t> baseaddr) {
+  // Sort the segments by base address, in case they weren't already.
+  struct {
+    bool operator()(const Segment &a, const Segment &b) {
+      return a.baseaddr < b.baseaddr;
+    }
+  } comparator;
+  std::vector<Segment> sorted = segments_orig;
+  std::sort(sorted.begin(), sorted.end(), comparator);
```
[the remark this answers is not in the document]
statham-arm: If you've got two segments with the same base address then surely the overlap detection just below here will reject the input file anyway?

I'll do the lambda, although after clang-format has broken it across 3 lines, it turns out it doesn't shorten the code as much as you'd hoped :-)

arm-software/shared/elf2bin/elf2bin.cpp
```diff
@@ -0,0 +1,468 @@
...
+  if (Arg *A = Args.getLastArg(OPT_output_file_EQ))
+    outfile = A->getValue();
+  if (Arg *A = Args.getLastArg(OPT_output_pattern_EQ))
+    outpattern = A->getValue();
+
+  std::optional<Format> format;
+  if (Arg *A = Args.getLastArg(OPT_ihex, OPT_srec, OPT_bin, OPT_bincombined,
+                               OPT_vhx, OPT_vhxcombined)) {
+    auto Option = A->getOption();
+    if (Option.matches(OPT_ihex))
```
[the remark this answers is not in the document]
statham-arm: I don't understand this comment – there _is_ a use of `getLastArg` in the snippet you've quoted?

statham-arm:
(Sorry for pushing these changes in two batches; I hadn't realised that when github said "9 hidden conversations" they were actually _important_. I assumed it had hidden them because they were obsolete, not just because it felt overwhelmed! So I responded to the top few review comments and the bottom few, and only found the middle ones afterwards. :-)

arm-software/shared/elf2bin/elf2bin.cpp
smithp35: OK, I see that check in a different file. I'll put a comment on that line to see if we can make the error message less ELF specific. Just thinking that a user might not know that relocatable objects don't have program headers. I realise that they are even less likely to understand ET_EXEC and ET_DYN.

arm-software/shared/elf2bin/elf.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+template <typename ELFT>
+static std::vector<Segment>
+get_segments(InputObject &inobj, ELFObjectFile<ELFT> &elfobj, bool physical) {
+  std::vector<Segment> segments;
+
+  Expected<ArrayRef<typename ELFT::Phdr>> phdrs_or_err =
+      elfobj.getELFFile().program_headers();
+  if (!phdrs_or_err) {
+    fatal(inobj, "unable to read program header table",
```
smithp35, line 23: A user might not understand that a relocatable object doesn't have a program header table. Perhaps worth adding a sentence like "elf2bin works on executables or shared libraries."

arm-software/shared/elf2bin/bin.cpp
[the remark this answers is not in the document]
smithp35: I'm guilty of leaving comments while reading top to bottom. Yes it will reject these.

AIUI the segment check will work with overlays as long as the user has selected p_paddr as the baseaddr.

arm-software/shared/elf2bin/elf2bin.cpp
smithp35: I should rephrase to be less specific to conversion. There are quite a few parameters that are shared between all of the functions and these could be grouped together in a struct to shorten the parameter list. Given that some of these functions have upwards of 8 parameters it can be hard to spot differences.

Overall this is just a subjective preference on my part.

arm-software/shared/elf2bin/elf2bin.h
```diff
@@ -0,0 +1,157 @@
...
+// congruent to {4,5} mod 8. (Still two residues, but now starting at 4.)
+//
+// If no bank switching is needed, then modulus=1, firstres=0, nres=1
+// is a combination that indicates 'write all input bytes to the output'.
+//
+// The output is written to the file 'outfile'.
+//
+// These functions will exit the entire program with an error message if
+// anything goes wrong. So callers need not handle the failure case.
+void bin_write(InputObject &inobj, const std::string &outfile,
```
pratlucas, line 84: I couldn't spot any non-const uses of `inobj` on the implementation side of the `*_write` methods (I might have missed something, though). Can they be made a `const &` in these interfaces?

smithp35 (approved):
Thanks for the updates.

LGTM on my side for the initial implementation as it is functionally complete, documented and tested. If we find out that further changes are needed to the implementation or tests we can iterate on it.

I've marked all my comments as resolved.

pratlucas (approved):

PR #260, opened 2025-04-09
[Downstream change][LV] Partially block upstream code causing 'Verification Error' on Grace

pawosm-arm:
Downstream issue: #259

PR #335, opened 2025-05-16
Remove workaround for Grace "Verification Error"

dcandler:
#260 applied a downstream change to fix a Grace bug with LLVM 20.1. The upstream fix for the same bug has now been merged in commit 0e50aeaa8a4720656f6cc8b7e267aca07779bd63. This also added new tests, which currently do not give the expected output due to the downstream modifications.

Since the backported upstream fix makes the downstream workaround unecessary, this patch removes the downstream changes and restores the files to match upstream.

Removes downstream issue: #259

### DIFF
--- a/llvm/lib/Analysis/IVDescriptors.cpp
+++ b/llvm/lib/Analysis/IVDescriptors.cpp
@@ -985,7 +985,6 @@ bool RecurrenceDescriptor::isReductionPHI(PHINode *Phi, Loop *TheLoop,
                       << *Phi << "\n");
     return true;
   }
-#if 0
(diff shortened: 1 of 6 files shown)


pawosm-arm (approved):
The reverted chances to the test files were reautogenerated. I assume you did the same to get the test files which actually pass.

dcandler:
The tests had already been updated for the upstream fix in 91a3f14d9497b761258d9358c31215ac29969dca (merged as 1a8041eff4db78a493556ffb0704588ec90d61a4) so there was no need to generate them again. With the downstream modifications removed, all the tests passed.

PR #339, opened 2025-05-20
[ATfE] Fix and re-enable package test on MacOS

pratlucas:
The package test implementation curretly relies on using the CMake `tar`
command to unpack the toolchain for testing. As this CMake command is
not able to handle .dmg image files, the package test was disabled
temporarily on MacOS platforms.
This change implements an alternative way to unpack the toolchain .dmg
file for testing on MacOS. It introduces a new `extract_dmg.sh` script
that makes use of the `hdiutil` program to extract the contents
correctly. With the fix in place, this change also re-enables the
package tests on MacOS platforms.

### DIFF
--- a/arm-software/embedded/CMakeLists.txt
+++ b/arm-software/embedded/CMakeLists.txt
@@ -887,13 +887,23 @@ add_custom_target(
     COMMAND "${CMAKE_COMMAND}" -E rm -rf unpack
     COMMAND "${CMAKE_COMMAND}" -E make_directory unpack
 )
-add_custom_target(
-    unpack-llvm-toolchain
-    COMMAND "${CMAKE_COMMAND}" -E tar x ${package_filepath}
-    DEPENDS ${package_filepath}
-    USES_TERMINAL
-    WORKING_DIRECTORY unpack
-)
+if(package_filename_extension STREQUAL ".dmg")
+    add_custom_target(
+        unpack-llvm-toolchain
+        COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/scripts/extract_dmg.sh ${package_filepath} ${PACKAGE_FILE_NAME}
+        DEPENDS ${package_filepath}
+        USES_TERMINAL
+        WORKING_DIRECTORY unpack
+    )
+else()
+    add_custom_target(
+        unpack-llvm-toolchain
+        COMMAND "${CMAKE_COMMAND}" -E tar x ${package_filepath}
+        DEPENDS ${package_filepath}
+        USES_TERMINAL
+        WORKING_DIRECTORY unpack
+    )
+endif()
 add_dependencies(
     unpack-llvm-toolchain
     clear-unpack-directory
@@ -905,16 +915,7 @@ add_dependencies(check-llvm-toolchain check-${LLVM_TOOLCHAIN_C_LIBRARY})
 add_dependencies(check-llvm-toolchain check-compiler-rt)
 add_subdirectory(test)
 add_dependencies(check-llvm-toolchain check-llvm-toolchain-lit)
-
-# The package tests currently only support tar and zip, so don't
-# attempt to run them on other formats, such as dmg with macOS.
-# A dummy check-package-llvm-toolchain target is still
-# required for the test.sh script.
-if(package_filename_extension STREQUAL ".tar.xz" OR package_filename_extension STREQUAL ".zip")
-    add_subdirectory(packagetest)
-else()
-    add_custom_target(check-package-llvm-toolchain)
-endif()
+add_subdirectory(packagetest)
 
 if(LLVM_TOOLCHAIN_CROSS_BUILD_MINGW)
     find_program(MINGW_C_EXECUTABLE x86_64-w64-mingw32-gcc-posix REQUIRED)
--- a/arm-software/embedded/CMakeLists.txt
+++ b/arm-software/embedded/CMakeLists.txt
@@ -888,6 +888,8 @@ add_custom_target(
     COMMAND "${CMAKE_COMMAND}" -E make_directory unpack
 )
 if(package_filename_extension STREQUAL ".dmg")
+    # Extract only the toolchain directory from the .dmg file in ${package_filepath},
+    # as the image also includes a symlink to the "Applications" directory.
     add_custom_target(
         unpack-llvm-toolchain
         COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/scripts/extract_dmg.sh ${package_filepath} ${PACKAGE_FILE_NAME}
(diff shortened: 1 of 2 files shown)


arm-software/embedded/CMakeLists.txt
```diff
@@ -887,13 +887,23 @@ add_custom_target(
...
-    unpack-llvm-toolchain
-    COMMAND "${CMAKE_COMMAND}" -E tar x ${package_filepath}
-    DEPENDS ${package_filepath}
-    USES_TERMINAL
-    WORKING_DIRECTORY unpack
-)
+if(package_filename_extension STREQUAL ".dmg")
+    add_custom_target(
+        unpack-llvm-toolchain
+        COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/scripts/extract_dmg.sh ${package_filepath} ${PACKAGE_FILE_NAME}
```
statham-arm, line 893: This command line took me some head-scratching to make sense of. From the script below, `$1` is the .dmg file to unpack, and `$2` is a glob pattern? So we're extracting only one thing from the archive and it's called `${PACKAGE_FILE_NAME}`? I guess that must be a top-level directory inside the .dmg containing the rest of the distribution?

arm-software/embedded/scripts/extract_dmg.sh
```diff
@@ -0,0 +1,40 @@
...
+    exit 1
+fi
+
+DMG_PATH="${1}"
+PATTERN_TO_EXTRACT="${2:-"*"}"
+DST_DIR="${3:-"."}"
+
+if command -v hdiutil > /dev/null; then
+    ATTACH_OUTPUT=$( hdiutil attach -nobrowse -readonly "${DMG_PATH}" )
+    MOUNT_PATH=$( echo "${ATTACH_OUTPUT}" | grep '/Volumes/' | sed -r 's/.*(\/Volumes\/.*)/\1/' )
```
statham-arm, line 30: Tiny sed tip: if your regex wants to match things with `/` in it, it's easier to read if you don't also use `/` as the delimiter. You can write e.g. `sed -r 's!.*(/Volumes/.*)!\1!'` and then you don't have quite so many toothpicks scattered around.

arm-software/embedded/scripts/extract_dmg.sh
```diff
@@ -0,0 +1,40 @@
...
+    echo "usage: ${0} <path_to_dmg> [<glob_pattern_to_extract>] [<dst_dir>]"
+    exit 1
+fi
+
+DMG_PATH="${1}"
+PATTERN_TO_EXTRACT="${2:-"*"}"
+DST_DIR="${3:-"."}"
+
+if command -v hdiutil > /dev/null; then
+    ATTACH_OUTPUT=$( hdiutil attach -nobrowse -readonly "${DMG_PATH}" )
```
statham-arm, line 29: It's difficult to review this without knowing what the output of `hdiutil attach` typically looks like. Can you show an example? (In a reply here will do – I won't insist on it being a comment in the script, unless you think that's a good idea yourself.)

arm-software/embedded/scripts/extract_dmg.sh
```diff
@@ -0,0 +1,40 @@
...
+    MOUNT_PATH=$( echo "${ATTACH_OUTPUT}" | grep '/Volumes/' | sed -r 's/.*(\/Volumes\/.*)/\1/' )
+    if [[ -z "${MOUNT_PATH}" ]]; then
+        >&2 echo "ERROR: Path to mounted volume not found."
+        exit 1
+    fi
+    cp -R "${MOUNT_PATH}"/${PATTERN_TO_EXTRACT} .
```
statham-arm, line 35: What happens if one of these commands fails? You're not checking the return code of any of them. I _presume_ that if `hdiutil attach` fails then it won't print output matching your regex, so the failure will be detected, but if `cp -R` doesn't work (e.g. because the destination disk has filled up), then the following `hdiutil detach` (assuming it succeeds) will mask the failure by resetting `$?` to 0 before exiting the script.

The usual advice is `set -e` so that the first failing command terminates the script. But in this case I think you probably _do_ still want to do the `hdiutil detach`, because mounting a thing and then forgetting to umount it is also bad error-handling behaviour. I think I'd do something like
```
cp -R "${MOUNT_PATH}"/${PATTERN_TO_EXTRACT} .
exitcode=$?
hdiutil detach "${MOUNT_PATH}" && exit $?
```
which will still _run_ the detach, but then propagate the exit code from the cp.

arm-software/embedded/CMakeLists.txt
pratlucas: That's correct. The .dmg file contains an extra symlink to the `Applications` directory, which is there for the drag-and-drop installation experience on Macs, so simply extracting all of the image's contents is probably not the best approach.

arm-software/embedded/scripts/extract_dmg.sh
```diff
@@ -27,13 +27,19 @@ DST_DIR="${3:-"."}"
 
 if command -v hdiutil > /dev/null; then
     ATTACH_OUTPUT=$( hdiutil attach -nobrowse -readonly "${DMG_PATH}" )
-    MOUNT_PATH=$( echo "${ATTACH_OUTPUT}" | grep '/Volumes/' | sed -r 's/.*(\/Volumes\/.*)/\1/' )
+    # The output of `hdiutil attach` includes a description of the mounted
+    # volumes with their corresponding mount path. E.g.:
+    # /dev/disk4     GUID_partition_scheme
+    # /dev/disk4s1   EFI
+    # /dev/disk4s2   Apple_HFS              /Volumes/ATfE-21.0.0-pre-Darwin-universal
```
statham-arm, line 34: Hmmm. So the output doesn't explicitly tell you _which_ mounted volume corresponds to the specific .dmg you mounted using this command? It just lists _all_ currently mounted things, and expects you to be able to guess which is the one you're interested in?

Suppose you had another .dmg mounted already when this script ran. (Maybe because another CI job is running in parallel, or maybe because some failure left a previous .dmg mounted.) It sounds as if you'd find _two_ entries in the output from `hdiutil` with `/Volumes/` in them, and nothing to tell you which one was the right one?

Perhaps that's worth a safety precaution. If the grep+sed that constructs `MOUNT_PATH` gives more than one output, I think it would be safer to fail with an error message than to proceed anyway with what might be totally the wrong path.

pratlucas: No, it only includes the mounted volumes that correspond to the .dmg file included in the command. Because of the Mac's internals it turns out that that corresponds to more than one concrete mount, but only one of them maps to a directory under `/Volumes`.
I'll update the comment to be clearer on this detail.

statham-arm: If all three of those output lines are referring to this .dmg file, then that sounds as if every .dmg used for software distribution has an EFI boot partition that never gets used? That's bizarre!

statham-arm (approved):

PR #340, opened 2025-05-21
elf2bin: fix RTTI-related build failure under clang.

statham-arm:
elf2bin depends on the LLVM supporting libraries, which are built without RTTI. So it must also be built without RTTI, to avoid a link failure when the RTTI for `Elf2BinOptTable` tried to refer to the RTTI for its base classes in `libLLVMOption` and didn't find it.

I had intended to turn off RTTI in all our builds, by writing an if statement that knew the gcc option -fno-rtti and the MSVC option /GR-. But I put the wrong condition in the if, and only enabled -fno-rtti for gcc itself, not for all compilers with gcc-compatible command lines. So when building with clang, RTTI was enabled in elf2bin, and the link failed.

### DIFF
--- a/arm-software/shared/elf2bin/CMakeLists.txt
+++ b/arm-software/shared/elf2bin/CMakeLists.txt
@@ -7,6 +7,10 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 #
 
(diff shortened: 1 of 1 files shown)


dcandler (approved):
LGTM

PR #341, opened 2025-05-21
Add elf2bin to the build via LLVM_EXTERNAL_PROJECTS

statham-arm:
Adding it directly via `add_subdirectory` meant that it was built, but not installed as part of the overall toolchain package, because I hadn't manually included a cmake `install` command.

I'd expected that `add_llvm_tool` would set up installation automatically. But in fact it only does that if the tool is on the list in `LLVM_TOOLCHAIN_TOOLS`. Worse, if the installation targets aren't set up by the time the main LLVM cmake setup finishes running, then adding the tool to `LLVM_DISTRIBUTION_COMPONENTS` causes another error, because the end of the LLVM cmake setup tries to add dependencies to all those targets. So calling `add_subdirectory` after LLVM installation finishes is simply too late. You need to arrange that elf2bin's `CMakeLists.txt` is invoked _during_ LLVM configuration.

`LLVM_EXTERNAL_PROJECTS` seems to be the LLVM cmake setup's escape hatch for downstreams to add extra things that aren't in the upstream tree. Configuring elf2bin in that before involving LLVM's cmake setup, instead of using `add_subdirectory` afterwards, does the right thing.

We still need to add elf2bin to `LLVM_TOOLCHAIN_TOOLS` to make the install targets exist, and add it to `LLVM_DISTRIBUTION_COMPONENTS` to make it one of the tools ATfE includes in the distribution package.

A nice side effect is that now that `elf2bin/CMakeLists.txt` is called _from_ the LLVM main cmake setup, we don't have to manually import as many of LLVM's cmake definitions – they're already in scope automatically.

### DIFF
--- a/arm-software/embedded/CMakeLists.txt
+++ b/arm-software/embedded/CMakeLists.txt
@@ -204,6 +204,7 @@ set(LLVM_DISTRIBUTION_COMPONENTS
     clang-resource-headers
     clang
     dsymutil
+    elf2bin
     lld
     llvm-ar
     llvm-config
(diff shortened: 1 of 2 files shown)


dcandler (approved):
LGTM

PR #351, opened 2025-05-29
Add automated check for downstream changes

dcandler:
In order to follow the project's downstream patch policy (see CONTRIBUTING.md), modifications to the LLVM codebase (all files outside the arm-software sub-directory, with a few exceptions) need to have a issue opened in GitHub in order to track the changes, with the pull request linking to that issue.

This patch implements a Python script and a GitHub workflow to run the script whenever a pull request is opened or modified. The script takes a pull request number, and uses the GitHub command-line interface to get information about which files are changed, and whether the body text of the pull request contains the required linkage. To determine which files need tracking, the file changes are compared against the automerge ignore list, which already contains the rules for which files are expected to be different. The body text of the pull request can then be searched for the expectedly formatted tagging.

Automating the check with a GitHub workflow should ensure downstream changes are properly tracked and that anyone making a downstream change is aware of the downstream patch policy.

This currently does not check whether a pull request which states that it removes a downstream change actually completely removes a downstream change. That will require additional scripting, to follow later.

### DIFF
--- a/.github/workflows/check_downstream_changes.yml
+++ b/.github/workflows/check_downstream_changes.yml
@@ -1,3 +1,16 @@
+# Copyright (c) 2025, Arm Limited and affiliates.
+# Part of the Arm Toolchain project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# This workflow runs the check_downstream_changes.py script in order to check
+# whether a pull request makes a downstream change, and if so contains the
+# appropriate tracking in the text. The script only requires the pull request
+# number, and should be triggered whenever a pull request is editted or
+# synchronized, since both the files changed and the text of the pull request
+# itself are part of the check.
+# The script requires the GitHub CLI to be available, and authentiated.
+
 name: check-downstream-changes
 
 on:
@@ -13,8 +26,6 @@ on:
     branches:
       - arm-software
       - release/arm-software/**
-
-  workflow_dispatch: ~
 
 jobs:
   check-downstream-changes:
--- a/.github/workflows/check_downstream_changes.yml
+++ b/.github/workflows/check_downstream_changes.yml
@@ -1,0 +1,40 @@
+name: check-downstream-changes
+
+on:
+  # Trigger whenever a pull request is opened or changed.
+  # Use pull_request_target since we don't want to checkout the version of
+  # the script in the pull request, which may have been modified.
+  pull_request_target:
+    types:
+      - opened
+      - reopened
+      - edited
+      - synchronize
+    branches:
+      - arm-software
+      - release/arm-software/**
+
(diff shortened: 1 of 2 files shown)


.github/workflows/check_downstream_changes.yml
```diff
@@ -0,0 +1,40 @@
...
+    types:
+      - opened
+      - reopened
+      - edited
+      - synchronize
+    branches:
+      - arm-software
+      - release/arm-software/**
+
+  workflow_dispatch: ~
```
HugoSilvaSantos, line 17: Please add a comment to explain this line allow the workflow to be triggered manually from the GH actions UI

arm-software/ci/check_downstream_changes.py
```diff
@@ -0,0 +1,245 @@
+#!/usr/bin/env python3
+
+"""
+A script to check that a pull request adheres to the downstream patch policy.
+If the pull request modifies file outside the arm-software build directory
+(or any other files excluded from automerge) then the pull request needs to
+contain specific text to link to a downstream tracking issue.
+
+Requires the GitHub CLI tool (gh) to query the repo.
```
HugoSilvaSantos, line 9: Is it available by default on GH runners?

kiranchandramohan:
Can we also add a comment to any downstream PR stating that it can only be submitted if approved by at least one engineer in ATfE and one engineer in ATfL?

.github/workflows/check_downstream_changes.yml
dcandler: Actually I meant to remove that; a manual workflow dispatch serves no purpose now since it requires the parameter of the pull request number.

arm-software/ci/check_downstream_changes.py
dcandler: Yes (GitHub CLI in https://github.com/actions/partner-runner-images/blob/main/images/arm-ubuntu-24-image.md).
The automerge workflow similarly makes use of the GitHub CLI on the same runner.

HugoSilvaSantos (approved):

dcandler:
> > > Can we also add a comment to any downstream PR stating that it can only be submitted if approved by at least one engineer in ATfE and one engineer in ATfL?
> > 
> > 
> > How downstream changes are approved is something I have been discussing separately with @pratlucas , in that when a pull request that makes downstream changes is detected, we can also automatically assign reviewers. What I was going to propose is to creating a group with members from both product teams who can specifically review such changes due to the shared impact, and automatically assign that group. That way the interested people are automatically notified, and approval from the group would be required rather than just an individual.
> > I don't think the branch protection rules can be tailored to block merges until one specific reviewer approves, however leaving a comment telling the author to wait for specific approval was one of possibilities discussed which can also be done through the script via the GitHub CLI.
> 
> Do you intend to do these separately or will you do it in a separate patch?

I think it's better done in a later patch once we've had some discussion between the teams on how approvals for downstream changes should work, since approvals is a separate issue to the tracking that this patch is dealing with.

PR #358, opened 2025-06-03
[ATfE] Enable aarch64 compiler-rt tests

vhscampos:
After recent bug fixes, the AArch64 compiler-rt tests are passing, therefore we are able to enable them all.


pratlucas (approved):
LGTM. Thanks for the work to enable those tests!

PR #363, opened 2025-06-05
[ATfL] build.sh modification to generate flang/armflang manpages

pawosm-arm:
This commit removes disablement of generating the flang docs (man pages). It was disabled as it had been failing. We had to wait until a fix for generating flang manpages has landed upstream. See https://github.com/llvm/llvm-project/pull/141882

### DIFF
--- a/arm-software/linux/build.sh
+++ b/arm-software/linux/build.sh
@@ -82,7 +82,6 @@ COMPILER_CMAKE_FLAGS=(
     -DCMAKE_SKIP_RPATH=No
     -DCMAKE_SKIP_INSTALL_RPATH=No
     -DLLVM_BUILD_DOCS=ON
-    -DFLANG_INCLUDE_DOCS=OFF
     -DLLVM_ENABLE_SPHINX=ON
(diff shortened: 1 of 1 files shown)


pawosm-arm:
This PR had to wait until a fix for generating flang manpages has landed upstream.

arm-software/linux/build.sh
```diff
@@ -82,7 +82,6 @@ COMPILER_CMAKE_FLAGS=(
     -DCMAKE_SKIP_RPATH=No
     -DCMAKE_SKIP_INSTALL_RPATH=No
     -DLLVM_BUILD_DOCS=ON
-    -DFLANG_INCLUDE_DOCS=OFF
```
kiranchandramohan, line 85: Why are we switching this off? This is also not mentioned in the summary.

pawosm-arm: We were switching it OFF as it used to fail before a recent upstream change has merged. Now in this commit I'm removing the line that switched generating the flang pages off.

pawosm-arm: I've added a reference to the relevant upstream PR in this PR commit message.

PR #364, opened 2025-06-05
Update run_fvp and run_qemu scripts, to capture and return the stdout and stderr to the calling function.

vrukesh:
Update run_fvp and run_qemu scripts, to capture and return the stdout and stderr to the calling functions.
This can also be useful for automation, and viewing the error strings.


arm-software/embedded/arm-runtimes/test-support/run_fvp.py
```diff
@@ -212,4 +212,4 @@ def run_fvp(
         result_stdout = result.stdout
 
     sys.stdout.buffer.write(result_stdout)
-    return result.returncode
+    return result.returncode, result.stdout.decode(), result.stderr.decode() if result.stderr != None else ""
```
statham-arm, line 217: In this kind of situation it's a good idea to add the `errors='replace'` argument to the decode calls. Then any bytes that aren't legal UTF-8 will be turned into a Unicode replacement character � instead of causing an exception.

If the program under test does emit invalid UTF-8, we still want to see in the log file what it _was_, so that we can get a clue about what went wrong. A Python exception that stops the script from showing us the useful detail isn't good for anyone.

PR #374, opened 2025-06-11
[ATfE] Build AArch64 libraries with -fPIC

vhscampos:
We have the need to support client code built with Position Independent Code (PIC) for AArch64.

Since PIC libraries are compatible with either PIC and non-PIC code, instead of duplicating all AArch64 libraries for PIC, we decided to transform the existing libraries into PIC.


simpal01 (commented):
Can you document these changes which are part of the PR?

Issue #375, opened 2025-06-12
[Downstream Change] Enable fstream independently of filesystem


voltur01:
**What is the change**
Since C++17 `fstream` is only enabled as part of the full `filesystem` support, however in embedded systems it is typical to redirect standard IO into a serial port or use basic file operations with semihosting, while the rest of `filesystem` (directories and such) is not available.

This change will enable `fstream` independently of the full `filesystem`.

**Why this change cannot be done upstream**
This was discussed with the community:
* The original libc++ change: https://reviews.llvm.org/D152168#4441783
* Embedded work group: https://discourse.llvm.org/t/llvm-embedded-toolchains-working-group-sync-up/63270/80
* libc++ monthly sync on 10-06-2025: https://docs.google.com/document/d/1yC-3bE6pefVizUhfiUs1Shv11EzuL0KOklGSLqto6dk/edit?tab=t.0#heading=h.f01ybrrpnyo2

The conclusion was that long term it is better to provide a porting API to be able to enable `filesystem` as the whole, even if parts of it are stubbed out.

This change will take quite some time, thus we are taking the downstream change to unblock users who want to port their code from pre-C++17, where `fstream` was available, to a later standard.

PR #376, opened 2025-06-12
Enable fstream independently of filesystem

voltur01:
Downstream issue: #375

This enables fstream for embedded systems where basic file operations are available via semihosting, but not additional filesystem operations like managing directories.

### DIFF
--- a/libcxx/include/fstream
+++ b/libcxx/include/fstream
@@ -191,7 +191,8 @@ typedef basic_fstream<wchar_t> wfstream;
 #else
 #  include <__config>
 
-#  if _LIBCPP_HAS_FILESYSTEM && _LIBCPP_HAS_LOCALIZATION
+// Downstream issue: #375 (Enable fstream independently of filesystem)
+//#  if _LIBCPP_HAS_FILESYSTEM && _LIBCPP_HAS_LOCALIZATION
 
 #    include <__algorithm/max.h>
 #    include <__assert>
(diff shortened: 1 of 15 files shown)


vrukesh (approved):
LGTM

voltur01:
@pawosm-arm please confirm you are happy with this one too, thank you!

libcxx/include/fstream
```diff
@@ -191,7 +191,8 @@ typedef basic_fstream<wchar_t> wfstream;
 #else
 #  include <__config>
 
-#  if _LIBCPP_HAS_FILESYSTEM && _LIBCPP_HAS_LOCALIZATION
```
pawosm-arm, line 194: I'm seriously concerned that those ifs are being commented out. E.g., we do fiddle with `_LIBCPP_HAS_LOCALIZATION` on some of the resistant build environments. Instead of commenting this out, could a new variable been introduced and this conditional expression extended to make use of it?

pawosm-arm (changes requested):

libcxx/include/fstream
voltur01: Thank you for the review!

`_LIBCPP_HAS_LOCALIZATION` is ON by default, thus ATfL has been built with it ON anyway - it is usually disabled for embedded systems that do not have any locale support at all (not just small differences in behaviour).

If you are still concerned, I can keep `_LIBCPP_HAS_LOCALIZATION`, but remove `_LIBCPP_HAS_FILESYSTEM` which again is ON by default, thus ON for ATfL builds.

pawosm-arm (approved):

libcxx/include/fstream
voltur01: Replaced with a check for `_NEWLIB_VERSION` here and in other places.

voltur01:
@pratlucas , @pawosm-arm I replaced commented out lines with a check for newlib, which now looks more consistent with other checks as well as less lines changed.

Please confirm you are happy with the update, thanks!

Issue #378, opened 2025-06-12
[Downstream Change] Include newlib.h to define _NEWLIB_VERSION


voltur01:
**What is the change**
This change adds check and include of `newlib.h` next to the same for `picolibc.h` to define `_NEWLIB_VERSION`.

**Why this change cannot be done upstream**
This was discussed in libc++ monthly sync on 10-06-2025 https://docs.google.com/document/d/1yC-3bE6pefVizUhfiUs1Shv11EzuL0KOklGSLqto6dk/edit?tab=t.0 and the conclusion was that it is a bad idea to include "random" header files, we need to make use of the C library definition in the CMake (now only LLVM libc is supported) to get C libarry specific settings.

This will take quite some time to design and implement, thus a downstream change is made to fix C++ library support with newlib.

PR #379, opened 2025-06-12
Check and include newlib.h

voltur01:
Downstream issue: #378

Add a check if newlib.h is available and include it, if yes, to get the definition of _NEWLIB_VERSION similar to picolibc.h


pratlucas (approved):
LGTM.

voltur01:
@pawosm-arm please confirm the patch is OK from ATfL side, thank you!

PR #382, opened 2025-06-16
Enable all headers on LLVM-libc

saturn691:
Follows from https://github.com/llvm/llvm-project/pull/144114 LLVM-libc doesn't provide all functions, but it is reasonable to expect the signature in the header file, consistent with picolibc. This allows the user to implement it later.


voltur01 (approved):
Great job!

PR #383, opened 2025-06-16
[NFC] Rearrange llvmlibc-support

saturn691:
Minor cleaning up of the llvmlibc-support folder.

- Split crt0 and semihost libraries
- Use the crt0 library provided in samples

### DIFF
--- a/arm-software/embedded/llvmlibc-support/semihost/CMakeLists.txt
+++ b/arm-software/embedded/llvmlibc-support/semihost/CMakeLists.txt
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2022-2025, Arm Limited and affiliates.
+# Copyright (c) 2025, Arm Limited and affiliates.
 # Part of the Arm Toolchain project, under the Apache License v2.0 with LLVM Exceptions.
 # See https://llvm.org/LICENSE.txt for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception


arm-software/embedded/samples/Makefile.conf
```diff
@@ -13,5 +13,6 @@ MICROBIT_TARGET=--target=armv6m-none-eabi -march=armv6m -mfpu=none -mfloat-abi=s
 AARCH64_TARGET=--target=aarch64-none-elf
 
 CRT_SEMIHOST=-nostartfiles -lcrt0-semihost -lsemihost
+CRT_SEMIHOST_LLVM=-nostartfiles -lcrt0 -lsemihost
```
voltur01: Some ideas how to avoid this difference:
* picolibc provides `crt0.o` if llvm-libc does so, then it would not need `-nostartfiles` and `-lcrt0`.
* If we expect that LLVM libc will need `crt0-semihost` in the future, then we can create it as a placeholder right now even if it only provides the same implementation.

arm-software/embedded/llvmlibc-support/crt0/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+#
+# Copyright (c) 2022-2025, Arm Limited and affiliates.
```
voltur01: Is `2022-` inherited from another file? Otherwise could be just 2025 (same in the other new file below).

PR #385, opened 2025-06-16
Update test script for sanitizers, reducing the list of covered tests to only M-profile coverage.

vrukesh:
Update test script for sanitizers, reducing the list of covered tests to only M-profile coverage.


pratlucas:
Can you update the description (and commit message when squashing) to cover the reason why this is needed, please?

voltur01 (approved):

PR #386, opened 2025-06-16
Update README.md with Armv9-A support.

smithp35:
We support v9-A code-generation. We map the v9-A family to the v8-A C-libraries.


pratlucas (approved):
LGTM.

PR #392, opened 2025-06-19
Update test script for sanitizers, to handle the failure in check-all tests.

vrukesh:
Update test script for sanitizers, to handle the failure in check-all tests. Now, script does not exit on the non-zero failure. Also, to indicate which command being executed, the commands are separated into individual lines and verbose is enabled.

### DIFF
--- a/arm-software/embedded/scripts/test_atfe_sanitizer.sh
+++ b/arm-software/embedded/scripts/test_atfe_sanitizer.sh
@@ -21,12 +21,22 @@ PROCESSOR_COUNT=$(getconf _NPROCESSORS_ONLN)
 
 cd "${REPO_ROOT}"/build
 
+# If a test fails, lit will ordinarily return a non-zero result,
+# which prevents further testing. Setting the --ignore-fail option
+# will cause testing to continue, so that CI systems can get a
+# full set of results.
+# The check-all target runs the upstream clang and LLVM tests,
+# which do not generate an junit xml results file by default.
+# Additionally setting the --xunit-xml-output option store the
+# results.
+export LIT_OPTS="--ignore-fail --xunit-xml-output=results.xml"
+ninja -j$PROCESSOR_COUNT check-all
+
 # The llvm-toolchain targets already set --xunit-xml-output so
 # only the --ignore-fail option is needed.
 # The picolibc tests do not use lit so do not support this option.
 export LIT_OPTS="--ignore-fail"
-ninja -j$PROCESSOR_COUNT check-all \
-    check-compiler-rt-armv7m_hard_fpv5_d16_exn_rtti_unaligned \
+ninja -j$PROCESSOR_COUNT check-compiler-rt-armv7m_hard_fpv5_d16_exn_rtti_unaligned \
     check-picolibc-armv7m_hard_fpv5_d16_exn_rtti_unaligned \
     check-cxx-armv7m_hard_fpv5_d16_exn_rtti_unaligned \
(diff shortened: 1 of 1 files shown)


arm-software/embedded/scripts/test_atfe_sanitizer.sh
```diff
@@ -21,12 +21,22 @@ PROCESSOR_COUNT=$(getconf _NPROCESSORS_ONLN)
...
+export LIT_OPTS="--ignore-fail --xunit-xml-output=results.xml"
+ninja -j$PROCESSOR_COUNT check-all
+
 # The llvm-toolchain targets already set --xunit-xml-output so
 # only the --ignore-fail option is needed.
 # The picolibc tests do not use lit so do not support this option.
 export LIT_OPTS="--ignore-fail"
-ninja -j$PROCESSOR_COUNT check-all \
-    check-compiler-rt-armv7m_hard_fpv5_d16_exn_rtti_unaligned \
+ninja -j$PROCESSOR_COUNT check-compiler-rt-armv7m_hard_fpv5_d16_exn_rtti_unaligned \
```
vhscampos: Do these test targets here have `-xunit-xml-output=results.xml` implictly? If so, there's not need to extract `check-all` from here. All the test targets would share the same `LIT_OPTS`

vrukesh: @vhscampos 
Apologies for the delay. I was still trying in vain for replicating locally the failures in Github action.

We have tried to keep this script as much as possible consistent with https://github.com/arm/arm-toolchain/blob/arm-software/arm-software/embedded/scripts/test.sh , hence I have retained the flags.

arm-software/embedded/scripts/test_atfe_sanitizer.sh
```diff
@@ -21,13 +21,23 @@ PROCESSOR_COUNT=$(getconf _NPROCESSORS_ONLN)
...
 # only the --ignore-fail option is needed.
 # The picolibc tests do not use lit so do not support this option.
 export LIT_OPTS="--ignore-fail"
-ninja -j$PROCESSOR_COUNT check-all \
-    check-compiler-rt-armv7m_hard_fpv5_d16_exn_rtti_unaligned \
-    check-picolibc-armv7m_hard_fpv5_d16_exn_rtti_unaligned \
-    check-cxx-armv7m_hard_fpv5_d16_exn_rtti_unaligned \
-    check-cxxabi-armv7m_hard_fpv5_d16_exn_rtti_unaligned \
-    check-unwind-armv7m_hard_fpv5_d16_exn_rtti_unaligned
+ninja -j$PROCESSOR_COUNT check-compiler-rt-armv7m_hard_fpv5_d16_exn_rtti_unaligned
```
HugoSilvaSantos, line 42: What is the advantage of using multiple ninja commands instead of one?

vrukesh: Verbose is enabled, so debugging is easier from the generated logs. Also, we are NOT exiting now on failure in any command. Otherwise there is no runtime impact.

HugoSilvaSantos: Can you add a comment to briefly explain the new approach please?

vrukesh: Edited the comment. Please check if its fine !

HugoSilvaSantos (approved):

PR #393, opened 2025-06-19
Change the name of LLVM libc package to be easier to identify

voltur01:
Now `ATfE-*-Linux-x86_64.tar.xz` matches both packages, making one `ATfE_llvmlibc-*-Linux-x86_64.tar.xz` will make it distinct and easier to handle in scripts.


vrukesh (approved):
LGTM.

Issue #352, opened 2025-05-29
[ATfE RFC] Observed Side Effect When Removing -O2 in Newlib Build for ATfE (commit 59ea1b4)


Laur59:
This message is not a request. I am writing to share our experience using ATfE (Arm Toolchain for Embedded), particularly focusing on an issue we encountered with newlib.

Even though newlib support is experimental, we build ATfE using newlib because our project is developed with gcc/newlib. We test our software on various boards with different cores: Cortex-M3, Cortex-M4, Cortex-M7, and Cortex-M33.

With our latest build of ATfE (from revision 2ab47295), we observed that the binary for the Cortex-M33 core dumps at boot. To resolve this, we rebuilt the toolchain by reintroducing the "-O2" optimization flag for newlib in CMakeLists.txt. This change resulted in a working binary for the Cortex-M33.

We acknowledge that this issue is likely related to newlib. However, we wanted to share our findings with the community for awareness and further discussion.

voltur01:
Hello,

Thank you for sharing this!

If I understand correctly, `-O2` removed here https://github.com/arm/arm-toolchain/pull/349 causes the regression, so may be it was there not by accident.

statham-arm:
Hi @Laur59,

Can you give some more information about this core dump? Are you able to share details of exactly how you built your binary (compiler and linker command lines)? In what part of the code did the crash occur?

I understand if you can't share your actual project, but if the crash occurs at startup, are you able to reduce it to a small example that still crashes, not containing any large amounts of your valuable code?

statham-arm:
That's something – the line that says "CFSR = 0x00100000 STKOF" means that the specific type of crash is a stack overflow.

I'm not familiar with this particular fault reporting framework, and I don't know why it lists all those registers as unavailable including pc. But perhaps the list of hex numbers at the bottom of the dump, under "Stack content before the fault", might include an address that matches something in the code segment of your binary? If so, then that address might point to a location in whatever function is recursing out of control, which could be a clue.

statham-arm:
Hi @Laur59,

I haven't been able to reproduce the crash here, because although your reproducer builds successfully, we don't have the development board you mention to run it on. I've been trying to run it on a Fast Model simulating a Cortex-M33, but it doesn't simulate the peripherals your code is reading, so all the I/O routines in your code either do nothing, or read nonsense. But I've stubbed out various parts of ukOS's init code in order to get _near_ the point of failure, and I have a theory.

[...]

PR #395, opened 2025-06-20
Set newlib optimization level to -Os

voltur01:
After -O2 optimization was removed in https://github.com/arm/arm-toolchain/pull/349 it was reported that memory usage increased a lot https://github.com/arm/arm-toolchain/issues/352

The proposed solution is to use -Os which is based on -O2, but takes care of code size that is important in embedded applications.


saturn691:
LGTM!

saturn691 (approved):

PR #397, opened 2025-06-23
[ATfE] Fix bootcode not working for ARMv4T

saturn691:
To fix some of the nightly failing build errors (introduced by the patch https://github.com/arm/arm-toolchain/pull/394) which was caused by the ARMv4T variant, stub functions need to be introduced temporarily. 

Previously, only ARM architectures 'A', 'R' and 'M' were considered, therefore the `#if ... #endif` clause would miss any other variants. When tests are able to run, we can fill in the stub functions to fix any future failed tests, but up until the tests, this acts to suppress the build errors.

### DIFF
--- a/arm-software/embedded/llvmlibc-support/crt0/bootcode.cpp
+++ b/arm-software/embedded/llvmlibc-support/crt0/bootcode.cpp
@@ -22,6 +22,7 @@
 #include "system_registers_m.h"
 #else
 // ARMv4T
+// TODO: fill in stub functions once we can start testing LLVM-libc
 namespace bootcode {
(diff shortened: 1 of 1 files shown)


voltur01:
Minor, this adds some new includes that do not seem to be relevant to the rest of changes.

In general, could you please expand the description a bit, what this gains us, e.g. it will be possible to build something (but not run?). If this is related to running tests, can we disable running tests in the JSON file, thus do not need to stubs?

BTW, if we keep the stubs, then it would be good to add some TODO or FIXME comments and may be even a GitHub issue link with the details to keep better track of remaining work for the future.

PR #400, opened 2025-06-25
[ATfE] Build Armv7-A and Armv7-R library variants for Thumb state

vhscampos:
This change also modifies the FVP config file of AArch32 to boot up in Thumb state.


simpal01 (approved):
LGTM

PR #402, opened 2025-06-26
[ATfL] Unify armclang and armflang manpages

pawosm-arm:
Now with a proper flang manpage available, we can unify the way armclang and armflang mapages are postprocessed.


pawosm-arm:
@ping?

PR #404, opened 2025-06-27
[ATfE] pacret_bti variants shouldn't be built with pacbti

john-brawn-arm:
In multilib.json the flags for these library variants don't have pacbti, so if we compile with e.g. -mcpu=cortex-m55 -mbranch-protection=standard we select the pacret_bti variant and end up executing an undefined bxaut instruction. Fix this by not building with pacbti, which means -mbranch-protection=standard causes instructions to be used that are forward-compatible with pacbti but are nop without it.

### DIFF
--- a/arm-software/embedded/arm-multilib/json/variants/armv8.1m.main_hard_fp_nomve_pacret_bti.json
+++ b/arm-software/embedded/arm-multilib/json/variants/armv8.1m.main_hard_fp_nomve_pacret_bti.json
@@ -3,7 +3,7 @@
         "common": {
             "TARGET_ARCH": "armv8.1m.main",
             "VARIANT": "armv8.1m.main_hard_fp_nomve_pacret_bti",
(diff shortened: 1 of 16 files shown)


voltur01:
Thank you for addressing this!

Originally, we added this variant to test the performance impact of PACBTI. Do we know what would be performance impact of using NOP-space instructions on Cortex-M85 that does implement PACBTI (vs BXAUT)?

If the impact is significant, should we rather look at updating library selection logic in multilib.yaml?

vhscampos:
From the compatibility standpoint it would be a no brainer, but I agree that we ought to be careful with performance considerations. Pinging @smithp35 to have more eyes here.

john-brawn-arm:
I'll take a look but I expect the performance impact to be close to zero: clang currently never emits bxaut (there's nothing in the ARM backend that will cause bxaut to be selected), so the only change as a result of this is in a few compiler-rt function we'll get aut+bx instead of bxaut (as a result of https://github.com/llvm/llvm-project/pull/146057).

voltur01 (approved):

Issue #405, opened 2025-07-01
[Downstream Change][ARM]Disable memcpy LDM/STM inlining


nasherm:
**What is the change**
Patch to disable memcpy LDM/STM inlining for Cortex-V7M

**Why this change cannot be done upstream**
An attempt was made to upstream this patch. Upstream reviewers wanted a better solution that could be more fine-grained for the issue at hand.

pawosm-arm:
What is the link to the upstream PR this issue mentions?

dcandler:
> Upstream would prefer a different solution
And why can't we implement their preferred solution? We should be avoiding downstream changes if at all possible, and if it's not possible we should record the rationale.

Also, please don't close the issue, as it needs to track the change while it exists.

PR #406, opened 2025-07-01
[Downstream Patch][ARM] Disable inline MEMCPY with LDM/STM instructions

nasherm:
This downstream patch is a performance patch for Cortex-V7M cores. An attempt was made to upstream this work to the LLVM project:

https://github.com/llvm/llvm-project/pull/106378

but it wasn't accepted by upstream as they would prefer a different solution.

Downstream issue:#405


pratlucas (approved):
LGTM, but as this is a downstream change to LLVM code please it'll also need an approval from @arm/arm-toolchain-linux .

PR #408, opened 2025-07-02
Check review approvals in downstream changes

dcandler:
Currently, pull requests only require a single approved review in order to merge. For downstream changes, we would prefer both product teams to review the changes first, but GitHub does not have the functionality to set different merge requirements depending on the paths of the changed files.

However, since we have a script that already runs on downstream changes to check it meets certain requirements, this could also check the review state and which teams have approved. This patch adds this additional check, getting the review information from the existing query and using the API to get the members of each team. If a review from a team is missing, the check will fail which will prevent merging. Once both have approved, the check will pass allowing merging.

The script automatically runs on modifications to the pull request, so it is additionally set to trigger when a pull request is reviewed. The `pull_request_review` trigger does not have a `branches` filter, so the job is set to run conditionally instead.

### DIFF
--- a/.github/workflows/check_downstream_changes.yml
+++ b/.github/workflows/check_downstream_changes.yml
@@ -35,9 +35,7 @@ jobs:
   check-downstream-changes:
     runs-on: ubuntu-24.04-arm
 
-    if: github.repository == 'arm/arm-toolchain' &&
-      (github.event.pull_request.base.ref == 'arm-software' ||
-      startsWith(github.event.pull_request.base.ref, 'release/arm-software/')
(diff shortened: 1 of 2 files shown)


dcandler:
While I hope I've made the check script sufficiently verbose that it clearly explains why it doesn't pass, a comment would also be useful to provide the required context.
The issues use the `downstream-change` label, so should we use the same for pull requests? In fact we could also check the issue has the correct label too.

pawosm-arm (approved):
I hope it has been tested thoroughly...

dcandler:
I was able to test the python script on the last downstream change, but GitHub workflow changes are much harder to test before they merge, especially when they require specific active pull requests.

But I did take another look at the yaml file just in case and now I've spotted a typo

pawosm-arm (approved):

Issue #409, opened 2025-07-02
[Downstream Change] [Pipelines] Additional unrolling  in LTO


VladiKrapp-Arm:
**What is the change**
Add extra full unrolling pass to improve performance on cores with branch predictors. It helps produce simplified loops, which can then be SROA'd allowing further simplification.
This change is already present in the codebase in the form of a performance patch file.

**Why this change cannot be done upstream**
The introduced change is in common code with no easy way to demonstrate general usefullness. We know this to be benefitial for some Arm specific cases from internal benchmarking.

PR #410, opened 2025-07-02
[Pipelines] Additional unrolling  in LTO

VladiKrapp-Arm:
Some workloads require specific sequences of events to happen to fully simplify. This adds an extra full unrolling pass to help these cases on the cores with branch predictors. It helps produce simplified loops, which can then be SROA'd allowing further simplification, which can be important for performance.
Feature adds extra compile time to get extra performance and is enabled by the opt flag 'extra-LTO-loop-unroll' (off by default).

Original patch by David Green (david.green@arm.com)

Downstream issue: #409


pawosm-arm (approved):
Approving as this is disabled by default

llvm/lib/Passes/PassBuilderPipelines.cpp
```diff
@@ -2039,6 +2044,19 @@ PassBuilder::buildLTODefaultPipeline(OptimizationLevel Level,
...
+
+// #409
+  if (LTOExtraLoopUnroll) {
+    LoopPassManager OmaxLPM;
+    OmaxLPM.addPass(LoopFullUnrollPass(Level.getSpeedupLevel(),
+                                       /* OnlyWhenForced= */ !PTO.LoopUnrolling,
+                                       PTO.ForgetAllSCEVInLoopUnroll));
+    FPM.addPass(
+        createFunctionToLoopPassAdaptor(std::move(OmaxLPM),
+                                        /*UseMemorySSA=*/false,
```
pawosm-arm, line 2056: nit: sparse use of spaces here and in the next line.

VladiKrapp-Arm: I ran clang-format on this, and this seems to conform.

VladiKrapp-Arm: I've brought both the blocks into same style, which is the prevalent one throughout the file.

pawosm-arm (approved):

PR #411, opened 2025-07-02
Move optimization flags docs into shared features docs

voltur01:
Move optimization flags documentation from embedded folder into the shared features documentation files and include them into the distribution package.

Remove embedded optimization flags file since the downstream chage policy requires adding documentation into the shared docs files.


VladiKrapp-Arm (approved):
LGTM. 

There are still files remaining in the embedded/docs folder, but all seem to be build related, so probably don't need to be distributed.

voltur01:
@VladiKrapp-Arm the whole `embedded/docs` folder is already included into the package - only the shared docs files where missing.

voltur01:
@pawosm-arm @kiranchandramohan are you OK with the changes to the shared `arm-software/docs/arm-toolchain-features.md` file?

arm-software/docs/arm-toolchain-features.md
```diff
@@ -19,4 +19,11 @@ There are no experimental features implemented.
...
 # Features
 
-There are no additional features implemented.
+## Additional loop unroll in the LTO pipeline
+In some cases it is benefitial to perform an additional loop unroll pass so that extra information becomes available to later passes, e.g. SROA.
+Use cases where this could be beneficial - multiple (N>=4) nested loops.
+
+Usage:
+```
+    -Wl,-plugin-opt=-extra-LTO-loop-unroll=true/false
```
pawosm-arm, line 28: Could you point me where is this `extra-LTO-loop-unroll` defined?

voltur01: Please see https://github.com/arm/arm-toolchain/issues/409 - this is one of changes that previously existed as a patch file under https://github.com/arm/arm-toolchain/tree/arm-software/arm-software/embedded/patches/llvm-project-perf

pawosm-arm: I'd postpone my review until the documented change is actually merged.

kiranchandramohan: Do we need to document this option in the common features page. Since this flag is not recommended for ATfL would it make sense only to advertise it in ATfE?

voltur01: The policy suggests that features that impact both toolchains are documented in the shared file to show the difference from upstream.

Whether it is useful or not, is a separate concern. If you think it is harmful in ATfL, then we have two options: document as such here or update the implementation patch to make it unavailable in ATfL (though I am not sure this is justified as it is off by default).

arm-software/docs/arm-toolchain-features.md
```diff
@@ -19,4 +19,11 @@ There are no experimental features implemented.
 
 # Features
 
-There are no additional features implemented.
+## Additional loop unroll in the LTO pipeline
+In some cases it is benefitial to perform an additional loop unroll pass so that extra information becomes available to later passes, e.g. SROA.
```
kiranchandramohan, line 23:
```suggestion
In some cases it is beneficial to perform an additional loop unroll pass so that extra information becomes available to later passes, e.g. SROA.
```

arm-software/docs/arm-toolchain-features.md
kiranchandramohan: We don't know whether it is harmful or not unless we run some performance tests with ATfL.

I am assuming that you have data/experience that shows it is beneficial for ATfE. But the recommendation here seems to be generic, and that is the concern here.

Do you switch this ON in a config file for ATfE? 

Can we retain the information about the availability of this flag here and remove the recommendation from this file and move that to an ATfE-specific file?

voltur01: It is used here https://github.com/arm/arm-toolchain/blob/arm-software/arm-software/embedded/OmaxLTO.cfg

I do not see this wording as a recommendation to use - it says "sometimes" and "could" so it would need to be judged in each use case. To be clear, it is not recommended for every embedded use case either - it is useful for benchmarks - thus I do not want to add a recommendation in ATfE file either.

kiranchandramohan (approved):
LG.

PR #412, opened 2025-07-03
[ATfL] Incorporate build-tests into our build process

pawosm-arm:


arm-software/linux/build.sh
```diff
@@ -252,6 +266,13 @@ apply_patches() {
     echo_bold "Applying patches...done"
 }
 
+prepare_boldtests() {
+    if ! bolttests_present; then
+        echo "The external bolt tests will be cloned now to ${BOLTTESTS_DIR}"
+        git clone "${BOLTTESTS_URL}" "${BOLTTESTS_DIR}"
+    fi
```
paschalis-mpeis: else `git pull` ?

pawosm-arm: the plan was, if the user provided the sources for them tests, we trust they are the ones that should be tested (as is). In a sense, `git clone` is a fallback here.

pawosm-arm: As I realized we can reach no agreement here, I've decided to get rid of the git clone fallback. In the end, we always run this script on the sources already cloned/pulled/checked-out by the CI scripts which populates docker images for building ATfL. Hence it should be up to the user to come with the correct sources which are up to date or whatever.

arm-software/linux/build.sh
```diff
@@ -190,6 +191,8 @@ Environment Variables:
                         (default: ${MKMODULEDIRS_PATH})
     SOURCES_DIR         The directory where all source code will be stored
                         (default: $SOURCES_DIR)
+    BOLTTESTS_DIR       The optionsl directory where the bolt-tests repo has been cloned
```
paschalis-mpeis: nit:
```suggestion
    BOLTTESTS_DIR       The options directory where the bolt-tests repo has been cloned
```

paschalis-mpeis (approved):
From BOLT's perspective the patch looks good, thank you!

arm-software/linux/build.sh
pawosm-arm: Thanks for spotting this, it meant to be `optional`.

PR #416, opened 2025-07-08
[RFC] Set C library for libc++ to avoid including extra headers

voltur01:
This is a prototype (i.e. not intended to be merged directly) to avoid downstream change https://github.com/arm/arm-toolchain/issues/378 and solve the issue of including picolibc.h or newlib.h directly.

It makes use of CMake LIBCXX_LIBC option to set the library at build time and generate C defines to detect the C library in C++ code.

This works with picolibc and newlib. llvm-libc option cannot be enabled yet, because it causes extra command line options to be passed to libcxxabi that is added in ATfE CMake in one step with libcxx itself - needs a refactoring to split these.


voltur01:
Was discussed in the [libc++ monthly sync](https://docs.google.com/document/d/1yC-3bE6pefVizUhfiUs1Shv11EzuL0KOklGSLqto6dk/edit?tab=t.0#heading=h.4sl73a7bnqpw) on 08-07-2025 and agreed that it is an acceptable approach.

PR #418, opened 2025-07-09
[ATfE] New CMake variable `LIBRARY_BUILD_TYPE` to select the build type for libraries

vhscampos:
This patch creates a new CMake variable called `LIBRARY_BUILD_TYPE`. This variable governs the build type to be used in all library builds and can have as value either `minsizerelease` or `release`. The list of options can be extended in the future. It is used to set the build type for the C library, compiler-rt and the C++ library. In addition, it is also used to set the optimization flag for newlib explicitly since it doesn't use neither CMake nor Meson.

With this change, we achieve consistency in the build types across all libraries.

The `PICOLIBC_BUILD_TYPE` is thus deprecated and removed.

As a consequence of this patch, all libraries will follow the same build type instruction. Before this, for instance, the C++ library was always built in MinSizeRel. From now on it can also be built in Release in accordance to what is decided for a particular variant. As a data point, `libc++.a` under the AArch64 A-profile unaligned variant has its size grown by 13% from switching from MinSizeRel to Release.


voltur01 (approved):
Makes sense, thank you!

statham-arm:
Sorry if I'm a bit late with this suggestion, but are `minsize` and `release` the best names?

It seems to me that the reason somebody would compile for `minsize` was because they needed to fit into the very small memory of the device the code is intended to run on. So _they_ are intending to release something based on a `minsize` build, and it seems odd to suggest they aren't, by calling only the other one `release`.

Perhaps `minsize` versus `speed`, or something that says what the `release` build does better?

PR #421, opened 2025-07-10
Allow using external llvm-project sources

voltur01:
This adds an option to point cmake to llvm-project sources outside of arm-toolchain project tree.


kiranchandramohan:
Could you prefix with ATfE if this is only for ATfE?

voltur01:
@kiranchandramohan do you mean to prefix the commit message or the CMake variable name?

kiranchandramohan:
> @kiranchandramohan do you mean to prefix the commit message or the CMake variable name?

The commit message.

PR #422, opened 2025-07-10
[ATfE] Re-enable use of zlib

### LINKED ISSUE #285: [ATfE] 20.1.0-ATfE --compress-debug-sections issue
Hi ARM,
We are using -Wl,--compress-debug-sections=zlib option, because of this option our elf size is significantly less.
This option was working fine with 20.0.0, but with latest version we are getting below error:
"error: --compress-debug-sections: LLVM was not built with LLVM_ENABLE_ZLIB or did not find zlib at build time"

If possible please consider this option for prebuilt binary packages.

pratlucas:
This change re-enables the use of zlib on builds of the Arm Toolchain
for Embedded when the library is available in the build environment. The
library is used by the compiler to compress debug session.

Information about zlib has also been added to the SBOM generation
script.

Fixes #285


HugoSilvaSantos (approved):

vrukesh (approved):
LGTM.

PR #423, opened 2025-07-14
[ATFE] Package common multilib headers into target triple directory.

simpal01:
The current ATFE package  redundantly duplicates identical C and C++ headers across more than 100 ibrary variants, despite the fact that these headers do not vary within a given target triple aside from few. As a result, each variant currently maintains its own include/ and include/c++/v1/ directories, leading to unnecessary duplication.

This patch is to refactor the multilib layout to use a single shared header directory per target triple, eliminating redundant copies of identical headers across library variants, while keeping variant-specific headers within each variant’s own include/ directory.

Changing the layout of the package structure will require corresponding updates to the ATFE build system, as it currently assumes a variant specific directory hierarchy for headers and libraries during installation and packaging.

A new Python script has been introduced that runs after the runtime subproject has been executed for each variant and the non-optimised multilib directories have been generated. This script identifies and extracts common headers from the non-optimised multilibs and creates an optimized multilib directory. In this optimised directory, only the common headers are centralised, while the remaining contents are preserved as-is from the non-optimised layout.

A CMake flag called ENABLE_MULTILIB_HEADER_OPTIMISATION has been added to disable this multilib optimisation phase. When this flag is not set, the optimised multilib directory will not be generated. Additionally, the optimisation step is skipped automatically if there are fewer than two multilib variants to build.

To support this new layout with a centralised include directory, a new field called IncludeDirs has been added to multilib.yaml. This field specifies a base directory for locating header files which gets added to the header search path sequence by clang.

Corresponding changes in Clang:
https://github.com/llvm/llvm-project/pull/146651


arm-software/embedded/arm-multilib/CMakeLists.txt
```diff
@@ -390,3 +399,19 @@ install(
...
 )
+
+# Extract common header files which spans across multilib variant directories.
+if(ENABLE_MULTILIB_HEADER_OPTIMISATION)
+    add_custom_target(generate-multilib-common-headers ALL
+        COMMAND ${Python3_EXECUTABLE} "${CMAKE_CURRENT_SOURCE_DIR}/common-headers-generate.py"
+                "${CMAKE_CURRENT_BINARY_DIR}/multilib"
+                "${CMAKE_CURRENT_BINARY_DIR}/multilib-optimised"
+        COMMENT "Generating common headers"
+	DEPENDS ${all_runtime_targets}
```
dcandler: This should also depend on the `multilib-yaml` target if the script expects the file to exist to be copied.
Also, the line is indented with a tab rather than spaces.

arm-software/embedded/arm-multilib/CMakeLists.txt
```diff
@@ -354,6 +362,7 @@ foreach(lib_idx RANGE ${lib_count_dec})
 
 endforeach()
 
+
```
dcandler: nit: extra newline

arm-software/embedded/arm-multilib/CMakeLists.txt
```diff
@@ -390,3 +399,19 @@ install(
...
+        COMMAND ${Python3_EXECUTABLE} "${CMAKE_CURRENT_SOURCE_DIR}/common-headers-generate.py"
+                "${CMAKE_CURRENT_BINARY_DIR}/multilib"
+                "${CMAKE_CURRENT_BINARY_DIR}/multilib-optimised"
+        COMMENT "Generating common headers"
+	DEPENDS ${all_runtime_targets}
+    )
+
+    install(
+                DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/multilib-optimised/"
+                DESTINATION .
```
dcandler: nit: too much indenting

arm-software/embedded/arm-multilib/common-headers-generate.py
```diff
@@ -0,0 +1,159 @@
...
+                else:
+                    print(f"Warning: {src_dir} does not exist and will be skipped.")
+
+        # Step4: Copy multilib.yaml file as it is from the non-optimised multilib directoy.
+        src_yaml = os.path.join(args.multilib_non_optimised_dir, "multilib.yaml")
+        dst_yaml = os.path.join(args.multilib_optimised_dir, "multilib.yaml")
+        if os.path.exists(src_yaml):
+            shutil.copy2(src_yaml, dst_yaml)
+        else:
+            print(f"Warning: {src_yaml} does not exist.")
```
dcandler: Should this be an error rather than a warning? The yaml file seems like it should be required.

arm-software/embedded/arm-multilib/CMakeLists.txt
simpal01: Done

arm-software/embedded/arm-multilib/CMakeLists.txt
simpal01: Done

arm-software/embedded/arm-multilib/CMakeLists.txt
simpal01: Done

arm-software/embedded/arm-multilib/common-headers-generate.py
```diff
@@ -0,0 +1,159 @@
...
+                f"Skipping extracting the common headers for {target}: input path {input_target_dir} not found"
+            )
+            continue
+
+        variant_includes = collect_variant_include_paths(input_target_dir)
+        if len(variant_includes) < 2:
+            print(
+                f"Skipping extracing the common headers for {target}: not enough variants to compare.At least two variants must be enabled for the multilib header optimisation phase to proceed."
+            )
+            return
```
dcandler, line 83: If only one variant is found the script will exit early, without producing any output. But the CMake script expects something in the output folder. If the optimization can't run, then the input folder should be duplicated in the output.
Also, spelling: extracing

simpal01: I have used  add_custom_target() with a dependency on another targets (${all_runtime_targets}  and multilib-yaml)  in CMake. Since it doesn’t declare any outputs, it reruns whenever the dependency changes, but CMake doesn’t track whether anything was actually generated.

dcandler: There's still an install rule for the generated folder: if the optimization is skipped and nothing is generated, the build will throw an error because it can't find the the `multilib-optimised` folder.

Either the Python script needs to always create the `multilib-optimised` folder (even if no optimization is applied because there's only one variant), or the CMake script needs to be able to recognize that nothing was generated and instead install the non-optimized folder.

arm-software/embedded/arm-multilib/CMakeLists.txt
```diff
@@ -390,3 +400,19 @@ install(
     FILES ${CMAKE_CURRENT_BINARY_DIR}/multilib/multilib.yaml
     DESTINATION .
 )
```
dcandler: `multilib.yaml` will get installed twice, by this rule and as part of the `multilib-optimised` rule below. While the file is going to be the same, this rule can be disabled when optimization is enabled.

simpal01: Done

arm-software/embedded/arm-multilib/common-headers-generate.py
simpal01: Done

arm-software/embedded/CMakeLists.txt
```diff
@@ -180,6 +180,14 @@ option(
     OFF
 )
 
+option(
+    ENABLE_MULTILIB_HEADER_OPTIMISATION "Enable multilib header optimisation phase"
```
dcandler, line 184: The warning would be better in the help text, since if anyone runs the `cmake -LH` command that should explain why the option is there and why is should not be enabled yet. A message printed during configuration is easily missed, and unwanted if an updated clang is actually being used.

Likewise the help text for the option in the `arm-mulitilib` project should match.

dcandler: Since the clang changes will be going in first now, the warning won't be needed any more.

dcandler (approved):

dcandler (approved):

PR #424, opened 2025-07-14
Extend SECURITY.md with Arm Toolchain details

voltur01:
This extends the SECURITY.md inherited from the LLVM Project by adding details on how to report security issues for components not provided by LLVM Project.

SECURITY.md file was added to the automerge ignore list in a separate PR.

### DIFF
--- a/arm-software/ci/.automerge_ignore
+++ b/arm-software/ci/.automerge_ignore
@@ -3,3 +3,4 @@ arm-software
 CONTRIBUTING.md
 README.md
 LICENSE.TXT
+SECURITY.md
--- a/arm-software/ci/.automerge_ignore
+++ b/arm-software/ci/.automerge_ignore
@@ -3,4 +3,3 @@ arm-software
 CONTRIBUTING.md
 README.md
 LICENSE.TXT
-SECURITY.md
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,5 +1,16 @@
-# Reporting LLVM Security Issues
-
-To report security issues in LLVM, please follow the steps outlined on the
-[LLVM Security Group](https://llvm.org/docs/Security.html#how-to-report-a-security-issue)
-page.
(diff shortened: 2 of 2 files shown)


arm-software/ci/.automerge_ignore
```diff
@@ -2,4 +2,5 @@
 arm-software
 CONTRIBUTING.md
 README.md
-LICENSE.TXT
\ No newline at end of file
+LICENSE.TXT
+SECURITY.md
```
pawosm-arm: Please add EOL at EOF, github puts that red (-) icon here for a reason...

voltur01: It was like that before - hope adding an extra line will not affect the script.

@dcandler before I add it, do I need to split this into a separate PR to avoid the downstream script complaining?

pawosm-arm (dismissed):
LGTM in general.
Leaving EOL at EOF concern for your own consideration.

pawosm-arm:
nit: Check failed. Pull request #424 contains downstream changes, but does not have a correctly formatted link to a downstream tracking issue.

arm-software/ci/.automerge_ignore
dcandler: Yes, I think the automerge ignore file needs to be modified first, since that contributes towards determining what constitutes a downstream change. Changes to `.automerge_ignore` are already excluded from the check, so only after that list is modified will changes to `SECURITY.md` be similarly excluded. A bit of a hassle, but for security the check script doesn't run using the changes in the incoming PR.

Adding the extra line won't make a difference to either of the scripts using the file, so there's no harm in fixing the formatting.

vhscampos:
LGTM

arm-software/ci/.automerge_ignore
voltur01: Thank you, I move the change out to https://github.com/arm/arm-toolchain/pull/425 - please check.

PR #429, opened 2025-07-18
[ATfE] Changes in library variants build for code size

vhscampos:
- All the M-profile variants have their build type changed to `minsizerelease`. Exceptions apply to Armv8.1-M Main. See below.
- For Armv8.1-M Main, the following changes apply:
  - All the PACBTI variants have their build type changed to `minsizerelease`.
  - Ditto for all the soft nofp nomve variants.
  - For the `unaligned` variants that don't fall under the previous categories, a new version is created so that two versions now coexist: one with `release` as the build type, and another with `minsizerelease`.
- All variants whose build type is `minsizerelease` are renamed to include the "size" segment in their names. This includes not only M-profile but also Armv4t and Armv5te.
- Update tests and add ones that were missing, for example for Armv4t and Armv5te.

### DIFF
--- a/arm-software/embedded/test/multilib/armv4t.test
+++ b/arm-software/embedded/test/multilib/armv4t.test
@@ -1,7 +1,7 @@
 # RUN: %clang -print-multi-directory --target=arm-none-eabi -mfpu=none | FileCheck %s --check-prefix=ARMV4T-EXN-RTTI
-# ARMV4T-EXN-RTTI: arm-none-eabi/armv4t_exn_rtti_minsize{{$}}
+# ARMV4T-EXN-RTTI: arm-none-eabi/armv4t_exn_rtti_size{{$}}
 # ARMV4T-EXN-RTTI-EMPTY:
 
 # RUN: %clang -print-multi-directory --target=arm-none-eabi -mfpu=none -fno-exceptions -fno-rtti | FileCheck %s --check-prefix=ARMV4T
-# ARMV4T: arm-none-eabi/armv4t_minsize{{$}}
+# ARMV4T: arm-none-eabi/armv4t_size{{$}}
 # ARMV4T-EMPTY:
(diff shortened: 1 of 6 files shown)


dcandler:
The original version of the patch did make downstream changes, so reviews were automatically requested from both teams per the CODEOWNERS file. Those downstream changes have now been removed, but the review requests aren't removed automatically. The request to arm-toolchain-linux can be ignored/manually removed now.

arm-software/embedded/test/multilib/armv4t.test
```diff
@@ -1,3 +1,7 @@
-# RUN: %clang -print-multi-directory --target=arm-none-eabi -mfpu=none | FileCheck %s
-# CHECK: arm-none-eabi/armv4t_exn_rtti{{$}}
-# CHECK-EMPTY:
+# RUN: %clang -print-multi-directory --target=arm-none-eabi -mfpu=none | FileCheck %s --check-prefix=ARMV4T-EXN-RTTI
+# ARMV4T-EXN-RTTI: arm-none-eabi/armv4t_exn_rtti_minsize{{$}}
```
saturn691: Is this supposed to be `_minsize`?

arm-software/embedded/test/multilib/armv4t.test
```diff
@@ -1,3 +1,7 @@
...
+# ARMV4T-EXN-RTTI-EMPTY:
+
+# RUN: %clang -print-multi-directory --target=arm-none-eabi -mfpu=none -fno-exceptions -fno-rtti | FileCheck %s --check-prefix=ARMV4T
+# ARMV4T: arm-none-eabi/armv4t_minsize{{$}}
```
saturn691: Ditto for all the tests

arm-software/embedded/test/multilib/armv4t.test
vhscampos: Good catch. Thanks

PR #430, opened 2025-07-22
[ATFE] Mark tests XFAIL via downstream patching.

simpal01:
Three tests need to be marked as XFAIL due to the following reasons:

1. FVP AND QEMU are unable to handle crash related signals.
2. There is no compatible library available when the tests are run with -frwpi.


arm-software/embedded/patches/llvm-project/0002-libc-tests-with-picolibc-disable-large-tests.patch
```diff
@@ -40,8 +40,9 @@ index b5f9089308d2..0a83e75ceceb 100644
 +set(LIBUNWIND_TEST_PARAMS "long_tests=False;large_tests=False" CACHE STRING "")
```
vhscampos: Why this change?

vhscampos (commented):
Is it necessary to make these changes to the other downstream patches? I mean the git version at the bottom and the commit id at the top

arm-software/embedded/patches/llvm-project/0002-libc-tests-with-picolibc-disable-large-tests.patch
simpal01: It is the same set of lines .. But am not sure why it is highlighting.

pawosm-arm:
I guess this too should be ignored by the linux toolchain reviewers?

dcandler:
> I guess this too should be ignored by the linux toolchain reviewers?

Yes, it looks like the patches were put in the root folder in the first commit by mistake. I've removed the review request for the linux team.

vhscampos (approved):

PR #432, opened 2025-07-24
Include the upstream release branch in the "Sync from Upstream LLVM" workflow.

MarkMurrayARM:
Include the upstream release branch in the "Sync from Upstream LLVM" workflow.
(Step 2 from our branching guide)

### DIFF
--- a/.github/workflows/sync_from_upstream.yml
+++ b/.github/workflows/sync_from_upstream.yml
@@ -12,7 +12,7 @@ jobs:
     if: github.repository == 'arm/arm-toolchain'
     strategy:
       matrix:
-        branch: [main, release/20.x]
+        branch: [main, release/21.x]
     env:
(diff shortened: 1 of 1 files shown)


MarkMurrayARM:
Hi @voltur01 and @pawosm-arm - please review?

pawosm-arm:
Are we absolutely certain we don't want to track changes to the previous release anymore? If @voltur01 is OK with this, then I'm OK with it too.

voltur01:
Yes, we keep the main and one release branch - we would not re-release from 20.x at this point.

pawosm-arm (approved):
Approving it, but note, we'll still be tracking the release/20.x branch for performance, at least until ATfL21 is released. I suppose release/20.x does not receive any performance-critical changes these days, so it is probably ok to just replace it with release/21.x here.

PR #436, opened 2025-07-25
[ATfE] Move errno from semihost to libc and add time utils

saturn691:
Previously, we used an external errno in LLVM-libc, which had to be defined in the semihosting library. However, LLVM-libc does provide an option to use a single-threaded errno implementation on request, so for consistency with other C libraries, we move errno from `libsemihost.a` to `libc.a`.

Also add new functions to allow clock().


statham-arm (commented):
Commit message nits:

This all seems to be specific to llvmlibc, but you don't say so anywhere!

> This made no sense (why is errno a part of semihosting?)

Because libsemihost is the extra library component _we_ provide, built by our own build scripts. So it's easier to put stuff of our own in there than to add it to libc.a, because adding to our own library doesn't involve modifying libc's build scripts and risking merge conflicts.

More to the point, the comment you've deleted shows that we believed LLVM libc didn't provide errno at all – but apparently that was wrong, and it does provide errno if you know how to ask for it.

statham-arm (approved):

PR #438, opened 2025-07-28
[ATfE] Fail with an error message when -frwpi is used. (#428)

MarkMurrayARM:
ATfE doesn't currently provide any library variants built for RWPI. And if you link RWPI with non-RWPI code, nothing good will happen – they have incompatible ABIs. So we ought to give an error message rather than silently choose an incompatible library.

Using https://github.com/llvm/llvm-project/pull/149132 which allowed multilib selection to be aware of the relocation model at all, this commit does the simplest possible thing: adds a catch-all error message that will trigger on _any_ use of `-frwpi`, and report that there's no suitable library variant.

(Another possibility would have been to add `-fno-rwpi` to the validity criteria for each existing library, but that seemed less likely to give a good error message. It would allow RWPI variants to be present for some cases but not all, but at the moment, we don't need that.)

Since the error message works the same everywhere, I've just added one test.

(cherry picked from commit 43cc67f425ec889088f27d2fa06cdea7cb4c588e)


statham-arm:
As far as I can see, https://github.com/llvm/llvm-project/pull/149132 isn't on this release branch, so clang won't be including `-frwpi` in the multiilb flag set for this check to detect. So I'd expect the error message never to trigger at all. I presume the ongoing CI run will fail the test added by this commit.

PR #441, opened 2025-07-28
This implements generic test wrapper, used to execute binaries, such as with QEMU or FVP

vrukesh:
This implements generic test wrapper, used to execute binaries, such as with QEMU or FVP. Underneath implementation is transparent to the user, so user need not worry about the parameters required to pass to execute with underlying model (such as QEMU or FVP).


statham-arm:
Commit message nit: please don't write the whole message on a single line! In standard tools like `gitk` it extends off the right side of the screen and becomes unreadable. Please stick to the usual git convention of having a short (less than about 60 chars) subject line, then a blank line, and _then_ a whole paragraph of explanation if it's needed.

saturn691:
Also prefix it with [ATfE] - makes it easier to search through commits

arm-software/embedded/arm-runtimes/test-support/generic-test-wrapper.py
```diff
@@ -0,0 +1,153 @@
...
+    parser = argparse.ArgumentParser(
+        description="Run a single test using either qemu or an FVP"
+    )
+    # Top-level required argument
+    parser.add_argument("--test_executor", required=True, choices=["fvp", "qemu"],
+                        help="Executor type: fvp or qemu")
+
+    # Common arguments
+    parser.add_argument("--image", help="image file to execute")
+    parser.add_argument("--argv", help="parameters for executable")
```
statham-arm: You've got `--argv` taking a single string here, and `arguments` below taking a whole list of strings. Why both?

arm-software/embedded/arm-runtimes/test-support/generic-test-wrapper.py
```diff
@@ -0,0 +1,153 @@
+#!/usr/bin/env python3
+
```
voltur01: "2023-2024" should be "2023-2025" or even just "2025"?

arm-software/embedded/arm-runtimes/test-support/generic-test-wrapper.py
```diff
@@ -0,0 +1,153 @@
...
+import argparse
+import os
+import pathlib
+import subprocess
+import sys
+
+# https://mesonbuild.com/Unit-tests.html#skipped-tests-and-hard-errors
+EXIT_CODE_SKIP = 77
+
+disabled_tests = [
```
statham-arm: What's the expected form of entries in these "disabled tests" lists, when they stop being empty? Please add a comment explaining it.

It looks as if you're currently checking `args.image` using `.endswith`, presumably in the expectation that you put `"foo"` in `disabled_tests` and it matches `/some/path/to/foo` on the command line. But using `.endswith` on a pathname stored in a single string will also match `foo` at the end of some other filename, such as `newfoo` or `oldfoo`. Was that what you meant to do?

Also, how many different test suites is this wrapper expected to be used for? What happens if two of them use the same file name for unrelated tests?

arm-software/embedded/arm-runtimes/test-support/generic-test-wrapper.py
voltur01: Also do you want to use the banner like in other files, e.g. https://github.com/arm/arm-toolchain/blob/4462d8099aa87f1b746e7ab3f791880721f10710/arm-software/embedded/CMakeLists.txt#L1 lines 1-7

arm-software/embedded/arm-runtimes/test-support/generic-test-wrapper.py
```diff
@@ -0,0 +1,153 @@
+#!/usr/bin/env python3
+
+# SPDX-FileCopyrightText: Copyright 2023-2024 Arm Limited and/or its affiliates <open-source-office@arm.com>
```
statham-arm: Shouldn't there be a license specification as well as the copyright notice?

arm-software/embedded/arm-runtimes/test-support/generic-test-wrapper.py
```diff
@@ -0,0 +1,153 @@
...
+import os
+import pathlib
+import subprocess
+import sys
+
+# https://mesonbuild.com/Unit-tests.html#skipped-tests-and-hard-errors
+EXIT_CODE_SKIP = 77
+
+disabled_tests = [
+]
```
voltur01: Is this expected to be filled in later? or should it be on the previous line for an empty list?

Same below.

arm-software/embedded/arm-runtimes/test-support/generic-test-wrapper.py
```diff
@@ -0,0 +1,153 @@
...
+        help="model name for FVP",
+    )
+    parser.add_argument(
+        "--fvp_config",
+        action="append",
+        help="FVP config file(s) to use",
+    )
+    parser.add_argument(
+        "--tarmac",
+        help="file to wrote tarmac trace to (FVP only)",
```
statham-arm: Typo: "write"

arm-software/embedded/arm-runtimes/test-support/generic-test-wrapper.py
```diff
@@ -0,0 +1,153 @@
...
+    )
+
+    # QEMU arguments
+    parser.add_argument(
+        "--qemu_machine",
+        help="name of the machine to pass to QEMU",
+    )
+    parser.add_argument(
+        "--qemu_command",
+        help="name of the qemu binary used to execute. Must be present in PATH.",
```
statham-arm: Why "must be present in PATH"? What goes wrong if I pass an absolute pathname to a binary that isn't in `PATH`?

arm-software/embedded/arm-runtimes/test-support/generic-test-wrapper.py
```diff
@@ -0,0 +1,153 @@
...
+def main():
+    parser = argparse.ArgumentParser(
+        description="Run a single test using either qemu or an FVP"
+    )
+    # Top-level required argument
+    parser.add_argument("--test_executor", required=True, choices=["fvp", "qemu"],
+                        help="Executor type: fvp or qemu")
+
+    # Common arguments
+    parser.add_argument("--image", help="image file to execute")
```
statham-arm: Why `--image` for the image file? Wouldn't it be more natural to have the image file be the first positional parameter? It's less verbose, and also, then the command line _looks_ like any other command line:
```
generic-test-wrapper.py --option --option myprog arg1 arg2
```

arm-software/embedded/arm-runtimes/test-support/generic-test-wrapper.py
```diff
@@ -0,0 +1,153 @@
...
+import subprocess
+import sys
+
+# https://mesonbuild.com/Unit-tests.html#skipped-tests-and-hard-errors
+EXIT_CODE_SKIP = 77
+
+disabled_tests = [
+]
+
+disabled_tests_fvp = [
```
voltur01: The difference between `disabled_tests` and `disabled_tests_fvp` not clear from the use below, e.g. is `disabled_tests` for EQMU (i.e. opposite to FVP) or it is for both?

Would be nice to have a descriptive enough name or at least a comment.

arm-software/embedded/arm-runtimes/test-support/generic-test-wrapper.py
```diff
@@ -0,0 +1,153 @@
...
+from run_qemu import run_qemu
+from run_fvp import run_fvp
+import argparse
+import os
+import pathlib
+import subprocess
+import sys
+
+# https://mesonbuild.com/Unit-tests.html#skipped-tests-and-hard-errors
+EXIT_CODE_SKIP = 77
```
voltur01: Is this script expected to be run by meson?

arm-software/embedded/arm-runtimes/test-support/generic-test-wrapper.py
```diff
@@ -0,0 +1,153 @@
...
+def run(args):
+    if is_disabled(args.image, args.qemu_command is None):
+        return EXIT_CODE_SKIP
+
+    # Some picolibc tests expect argv[0] to be literally "program-name", not
+    # the actual program name.
+    argv = ["program-name"] + args.arguments
+
+    parent_dir = os.path.dirname(args.image)
+    exeNameString = os.path.basename(args.image)
```
statham-arm: It looks as if `parent_dir` is passed to the underlying run functions to use as the working directory. Can you add a comment explaining why? In general it's perfectly normal to run a command whose executable file is stored in one directory, with the cwd of the process being an unrelated directory. If this wrapper script stops you from doing that, please say why.

arm-software/embedded/arm-runtimes/test-support/generic-test-wrapper.py
```diff
@@ -0,0 +1,153 @@
+#!/usr/bin/env python3
+
+# SPDX-FileCopyrightText: Copyright 2023-2024 Arm Limited and/or its affiliates <open-source-office@arm.com>
+
+# This is generic test wrapper, used to execute binaries either with QEMU or FVP. 
+# Underneath implementation is transparent to the user, so user need not worry
```
voltur01: Since the user has to provide `--test_executor` and the rest of QEMU/FVP specific arguments, this description does not seem to be correct - it appears that the purpose of this script is something different.

arm-software/embedded/arm-runtimes/test-support/generic-test-wrapper.py
```diff
@@ -0,0 +1,153 @@
...
+
+    parser.add_argument(
+        "arguments",
+        nargs=argparse.REMAINDER,
+        default=[],
+        help="optional arguments for the image",
+    )
+    args = parser.parse_args()
+
+    # Conditional validation
```
voltur01: It is not clear what is the point or advantage of having this script with a bunch of executor specific options vs providing the same options to one of the underlying scripts: the caller logic will only differ in the name of the script - the rest will need to be worked out in the same way.

arm-software/embedded/arm-runtimes/test-support/generic-test-wrapper.py
vrukesh: Updated in latest version.

arm-software/embedded/arm-runtimes/test-support/generic-test-wrapper.py
vrukesh: Updated in latest version. 
It was retained as it is earlier from picolibc-wrapper script. Xfails mechanism will not be part of this script.

arm-software/embedded/arm-runtimes/test-support/generic-test-wrapper.py
vrukesh: Updated in latest version.

arm-software/embedded/arm-runtimes/test-support/generic-test-wrapper.py
vrukesh: Updated in latest version.

arm-software/embedded/arm-runtimes/test-support/generic-test-wrapper.py
vrukesh: Updated the text in latest version.

arm-software/embedded/arm-runtimes/test-support/generic-test-wrapper.py
vrukesh: Updated in latest version.

arm-software/embedded/arm-runtimes/test-support/generic-test-wrapper.py
vrukesh: Updated in latest version.

arm-software/embedded/arm-runtimes/test-support/generic-test-wrapper.py
vrukesh: Updated in latest version.

arm-software/embedded/arm-runtimes/test-support/generic-test-wrapper.py
vrukesh: Updated in latest version.

arm-software/embedded/arm-runtimes/test-support/generic-test-wrapper.py
vrukesh: Updated in latest version.

arm-software/embedded/arm-runtimes/test-support/generic-test-wrapper.py
vrukesh: Updated in latest version.

voltur01:
The code changes look good to me, thank you!

arm-software/embedded/arm-runtimes/test-support/generic-test-wrapper.py
vrukesh: This is similar to what we have done in picolibc-test-wrapper. Didn't do too much tinkering here.
Not fixed.

arm-software/embedded/arm-runtimes/test-support/generic-test-wrapper.py
vrukesh: Definitely this is not very best of the possible solutions. Let's revisit at later stage.

PR #442, opened 2025-07-28
[ATfE] Expose compile flags and execution flags from arm-runtimes

vrukesh:
[ATfE]: This patch helps to expose the compile flags and execution flags from arm-runtimes. 
These flags can then be used in generic manner, without need for the compile and execution framework to know if the execution is with QEMU or FVP.


vrukesh:
@saturn691 
1. Updated commit message.
2. There is no target. JSON is created just during configuration command:

rm -Rf build; mkdir build;cd build;
cmake ../arm-runtimes \
  -GNinja \
  -DVARIANT_JSON=../arm-multilib/json/variants/aarch64a_be.json \
  -DC_LIBRARY=picolibc \
  -DFVP_INSTALL_DIR=../fvp/install \
  -DLLVM_BINARY_DIR=../../../ \
  -DCMAKE_INSTALL_PREFIX=./install

vhscampos:
Please make the commit message more direct to the point. "This patch helps" is redundant. Just name it "[ATfE] Expose compile flags and execution flags from arm-runtimes".

Remove the ':' too.

arm-software/embedded/arm-runtimes/CMakeLists.txt
```diff
@@ -673,7 +724,6 @@ if(C_LIBRARY STREQUAL llvmlibc)
         ${common_llvmlibc_cmake_args}
         -DLIBC_TARGET_TRIPLE=${target_triple}
         -DLIBC_CONF_TIME_64BIT=ON
-        -DLIBC_CONF_ERRNO_MODE=LIBC_ERRNO_MODE_SHARED
```
dcandler: Why is this removed?

arm-software/embedded/arm-runtimes/CMakeLists.txt
```diff
@@ -145,6 +145,57 @@ else()
...
+    set (FVP_CONFIG_TEMP "")
+    foreach(cfg ${fvp_config_list})
+        string(APPEND FVP_CONFIG_TEMP " --fvp_config ${cfg}")
+    endforeach()
+
+    get_filename_component(FVP_INSTALL_ABS_PATH "${FVP_INSTALL_DIR}" REALPATH)
+    get_filename_component(FVP_CONFIG_ABS_PATH "${FVP_CONFIG_DIR}" REALPATH)
+
+    if ("${FVP_MODEL}" STREQUAL "aem-r")
+        set(COMPILE_FLAGS "${COMPILE_FLAGS} -mno-unaligned-access -nostartfiles -lcrt0-semihost -lsemihost -T picolibc.ld")
```
dcandler: `COMPILE_FLAGS` is a variable used to hold the flags required to compile the libraries. By writing linker flags into it, this prevents the libraries from being built correctly.

dcandler: `crt0-semihost` is only used for qemu, there's separate `crt0-semihost-fvp` for fvp testing. There's already code for correctly selecting this later.

dcandler (changes requested):
The flags needed to compile tests using the libraries are not necessarily going to be the same as the flags needed to compile the libaries in the first place.

This code duplicates and redefines things that don't need multiple definitions. It would make more sense to use the code that's already there instead of constructing these arguments in parallel.

Also I agree with the others about the commit message.

arm-software/embedded/arm-runtimes/CMakeLists.txt
```diff
@@ -146,6 +146,83 @@ else()
     set(cpu_family arm)
 endif()
 
+# Read the JSON file to load a multilib configuration.
```
dcandler, line 153: I don't think the runtimes project should be tighty coupled to the multilibs project in this way. The runtimes project builds a particular library variant, it shouldn't be concerned with the variant in the context of other variants built as a multilib set.

arm-software/embedded/arm-runtimes/CMakeLists.txt
```diff
@@ -146,6 +146,83 @@ else()
...
+    string(JSON lib_def GET ${multilib_defs} ${lib_idx})
+    string(FIND "${lib_def}" "error" index) 
+    if(index EQUAL -1)
+        string(JSON multilib_variant GET ${lib_def} "variant")
+        if(${VARIANT} STREQUAL ${multilib_variant})
+            message(STATUS "multilib_variant: ${multilib_variant}")
+            string(JSON variant_multilib_flags GET ${lib_def} "flags")
+            message(STATUS "variant_multilib_flags: ${variant_multilib_flags}")
+
+            string(REPLACE "-march=armvX+nofp -march=armvX+nosimd" "-march=armv8-a+nofp+nosimd" variant_multilib_flags "${variant_multilib_flags}")
```
dcandler, line 170: This seems like an especially brittle substitution, since it isn't clear that `armvX` will always be `armv8`.  And that sort of internal implementation of the multilib builder should remain in the scope of the multilibs project.

arm-software/embedded/arm-runtimes/CMakeLists.txt
```diff
@@ -146,6 +146,83 @@ else()
...
+				if ("${FVP_MODEL}" STREQUAL "aem-r")
+					set(LINKER_FLAGS "-mno-unaligned-access -nostartfiles -lcrt0-semihost -lsemihost -T picolibc.ld")
+					set(EXECUTION_FLAGS "--test_executor ${TEST_EXECUTOR} --fvp_model ${FVP_MODEL}${FVP_CONFIG_TEMP} --fvp_install_dir ${FVP_INSTALL_DIR} --fvp_config_dir ${FVP_CONFIG_DIR}")
+				else()
+					set(LINKER_FLAGS "-nostartfiles -lcrt0-semihost -lsemihost -T picolibc.ld")
+					set(EXECUTION_FLAGS "--test_executor ${TEST_EXECUTOR} --fvp_model ${FVP_MODEL}${FVP_CONFIG_TEMP} --fvp_install_dir ${FVP_INSTALL_ABS_PATH} --fvp_config_dir ${FVP_CONFIG_ABS_PATH}")
+				endif()
+
+			elseif("${TEST_EXECUTOR}" STREQUAL "qemu")
+				if(TARGET_ARCH MATCHES "^aarch64")
```
dcandler, line 198: It shouldn't be necessary to find the qemu programs twice in the same script, since this is now done here and later in the next code block (line 231+)

arm-software/embedded/arm-runtimes/CMakeLists.txt
```diff
@@ -146,6 +146,83 @@ else()
...
+					find_program(QEMU_EXECUTABLE qemu-system-aarch64 REQUIRED)
+				else()
+					find_program(QEMU_EXECUTABLE qemu-system-arm REQUIRED)
+				endif()
+
+				set(LINKER_FLAGS "-nostartfiles -lcrt0-semihost -lsemihost -T picolibc.ld")
```
dcandler, line 204: It also shouldn't be necessary to define the same linker flags twice. There was already logic in place to generate them (again in the block starting line 231), and what you've copied hasn't included selecting the correct startup code.

dcandler (changes requested):

PR #451, opened 2025-07-30
[ATFE] Prioritize multilib-specific headers over target-level headers.

simpal01:
Adjust the header search order so that variant-specific directories (e.g  armv6m_soft_nofp_size/include/) are searched before the common target-level directories (e.g include/) in the -internal-isystem list.

This ensures that headers tailored for a specific multilib variant take precedence over the generic headers, allowing correct header resolution in the presence of overrides.

Previously, the common target-level directory could shadow multilib-specific headers if the same file existed in both paths.


saturn691 (approved):
LGTM

vhscampos (approved):

PR #456, opened 2025-08-01
[ATFE] Improve error handling and header optimisation in common header generation logic.

simpal01:
This patch introduces additional error checking and handle edge cases in common header generation script.

1. Adds a check to handle cases where the multilib folder is either empty or does not exist, avoiding unhandled exceptions.
2. Improves logic by allowing header optimisation to proceed for targets that have more than two variants, even if another target lacks sufficient variants for a comparison. These changes ensure the script can still performing valid optimisations where possible.
3. Copying multilib.yaml can be moved outside the loop, as it only needs to be done once after all target folders are generated.

### DIFF
--- a/arm-software/embedded/arm-multilib/common-headers-generate.py
+++ b/arm-software/embedded/arm-multilib/common-headers-generate.py
@@ -57,10 +57,18 @@ def extract_common_headers_for_targets(args):
         shutil.rmtree(args.multilib_optimised_dir)
 
     if os.path.isdir(args.multilib_non_optimised_dir):
-        if not os.listdir(args.multilib_non_optimised_dir):
+        existing_target_dirs = [
+            dir_name
+            for dir_name in MULTILIB_TARGET_DIRS
+            if os.path.isdir(os.path.join(args.multilib_non_optimised_dir, dir_name))
+        ]
+        if not existing_target_dirs:
             raise Exception(
-                f"Error: Expected contents in '{args.multilib_non_optimised_dir}', but folder is empty."
+                f"Error: Expected to find either arm-none-eabi or aarch64-none-elf in '{args.multilib_non_optimised_dir}', but folder is empty."
             )
+        src_yaml = os.path.join(args.multilib_non_optimised_dir, "multilib.yaml")
+        if not os.path.exists(src_yaml):
+            raise FileNotFoundError(f"Source yaml '{src_yaml}' does not exist.")
     else:
         raise FileNotFoundError(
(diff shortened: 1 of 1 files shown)


arm-software/embedded/arm-multilib/common-headers-generate.py
```diff
@@ -80,16 +90,15 @@ def extract_common_headers_for_targets(args):
...
             # optimization is applied. In that case, multilib-optimised will just contain a copy of the
             # single variant from the non-optimised multilib directory.
-            if os.path.exists(args.multilib_non_optimised_dir):
+            if os.path.exists(input_target_dir):
                 shutil.copytree(
-                    args.multilib_non_optimised_dir,
-                    args.multilib_optimised_dir,
-                    dirs_exist_ok=True,
+                    input_target_dir,
+                    output_target_dir,
```
vhscampos: Add `dirs_exist_ok=False` explicitly to avoid future doubts about the value it should take.

arm-software/embedded/arm-multilib/common-headers-generate.py
```diff
@@ -80,16 +90,15 @@ def extract_common_headers_for_targets(args):
...
                 )
-            return
+            continue
 
         # Creating the common include headers for each target
-        os.makedirs(output_include_dir, exist_ok=True)
+        os.makedirs(output_include_dir)
```
vhscampos: Add `exist_ok=False` explicitly for the same reasons.

arm-software/embedded/arm-multilib/common-headers-generate.py
```diff
@@ -56,6 +56,16 @@ def extract_common_headers_for_targets(args):
     if os.path.exists(args.multilib_optimised_dir):
         shutil.rmtree(args.multilib_optimised_dir)
 
+    if os.path.isdir(args.multilib_non_optimised_dir):
+        if not os.listdir(args.multilib_non_optimised_dir):
```
dcandler: I don't think this check would be sufficient. When the toolchain built no variants, it still produced a `multilib.yaml` file but neither target directory. That wouldn't have been caught here since the directory wasn't empty - but it was still unusable by the script.

If this script expects a `multilib.yaml` and at least one of the two target directories to be present, then it should check for those conditions are met.

arm-software/embedded/arm-multilib/common-headers-generate.py

arm-software/embedded/arm-multilib/common-headers-generate.py

arm-software/embedded/arm-multilib/common-headers-generate.py

PR #457, opened 2025-08-01
[ATfE] Throw an error if a requested variant name is not found

dcandler:
By default the toolchain enables all available library variants to be built, but a list of specific variants can instead be enabled through the `-DLLVM_TOOLCHAIN_LIBRARY_VARIANTS` option. Currently, if an invalid name is specified, the configuration will not notice and that variant will silently not be built. This patch instead causes the script to throw an error, so that it is clear that an expected variant is missing if the name is incorrect.


dcandler:
> Wouldn't it be easier to check if ${lib_def} is empty?

`${lib_def}` won't be empty if an invalid name is included in the enable list: the build script is looping through each available variant to check whether it should be enabled or not.

saturn691 (approved):
Ok, fair enough

PR #458, opened 2025-08-01
[ATFE] Add note on header resolution issues when bypassing multilib with --sysroot.

simpal01:


saturn691 (approved):
Edit title to [ATfE]

arm-software/embedded/README.md
```diff
@@ -143,24 +143,11 @@ To display the directory selected by the multilib system, add the flag
...
--lcrt0-semihost \
--lsemihost \
--T picolibc.ld \
--o example example.c
-```
+Multilib headers are distributed across both `target level` and `variant level` directories.
+As a result, using a single `--sysroot` path to override multilib selection is no longer viable.
+Headers have to be pulled from multiple locations including both target and variant specific
+directories which the multilib system handles automatically. Therefore, specifying `--sysroot`
+to bypass multilibs can cause problems with header resolution, and is not recommended in this setup.
```
MarkMurrayARM: Is there any way of directly specifying a multilib? I suspect there is not, in which case this note should say so explicitly, rather than implying it.

MarkMurrayARM (changes requested):

arm-software/embedded/README.md
voltur01: +1

I think it should be very simple and direct statement, maybe even a warning to highlight.

The warning below does not make sense anymore - it was useful when --sysroot was available so there was a chance that someone depended on it.

simpal01: Done

arm-software/embedded/README.md
```diff
@@ -143,30 +143,10 @@ To display the directory selected by the multilib system, add the flag
...
--T picolibc.ld \
--o example example.c
-```
-
 > [!WARNING]
-> The naming convention for the library variant directories is an
-> implementation detail that can change between releases without notice,
-> thus it is best not to depend on specific library variant names in your
-> project build scripts.
+> --sysroot is not a substitute for multilib selection
```
voltur01:
```suggestion
> `--sysroot` is not a substitute for multilib selection.
```

arm-software/embedded/README.md
```diff
@@ -143,30 +143,10 @@ To display the directory selected by the multilib system, add the flag
...
+> The --sysroot option cannot be used to override multilib logic
```
voltur01:
```suggestion
> The `--sysroot` option cannot be used to override multilib logic
```

arm-software/embedded/README.md
```diff
@@ -143,30 +143,10 @@ To display the directory selected by the multilib system, add the flag
...
+> or manually select an arbitrary variant.
```
voltur01:
```suggestion
> or manually select an arbitrary library variant.
```

arm-software/embedded/README.md
simpal01: Done

arm-software/embedded/README.md
simpal01: Done

arm-software/embedded/README.md
simpal01: Done

PR #462, opened 2025-08-01
[ATfE] Fix downstream xfail patch about uses of `atomic`

vhscampos:
One of the xfailed tests that use `atomic` has changed location, so we had to also relocate the corresponding `lit.local.cfg` to the same location.


saturn691 (approved):
Can you edit the description to add the link of the commit that moved this test? If that's too much work, then it's ok, otherwise, LGTM

PR #463, opened 2025-08-04
Revert the upstream picolibc removal of dummyhost library.

MarkMurrayARM:
This is not an approach that is going to be used in later releases, but it does give us time to work on a more complete solution.


voltur01 (approved):
Thank you!

PR #465, opened 2025-08-05
[ATfE] Throw an error if a requested variant name is not found (#457)

dcandler:
By default the toolchain enables all available library variants to be built, but a list of specific variants can instead be enabled through the `-DLLVM_TOOLCHAIN_LIBRARY_VARIANTS` option. Currently, if an invalid name is specified, the configuration will not notice and that variant will silently not be built. This patch instead causes the script to throw an error, so that it is clear that an expected variant is missing if the name is incorrect.

(cherry picked from commit 2bce241944438676ec85a68cfe2cd7b3f0e5baa3)


MarkMurrayARM (approved):
Looks good!

PR #467, opened 2025-08-05
[ATfE] Fix issue with LLVM-libc builds for certain variants

saturn691:
FATAL_ERROR is supposed to be raised when a variant that uses FVPs tries to build tests for LLVM-libc, as they are not supported.

However, this check runs regardless of whether tests are enabled or not, causing build failures. This PR addresses that.

### DIFF
--- a/arm-software/embedded/arm-runtimes/CMakeLists.txt
+++ b/arm-software/embedded/arm-runtimes/CMakeLists.txt
@@ -735,7 +735,7 @@ if(C_LIBRARY STREQUAL llvmlibc)
 
     if(ENABLE_LIBC_TESTS)
         if(NOT(TEST_EXECUTOR STREQUAL qemu))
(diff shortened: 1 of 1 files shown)


arm-software/embedded/arm-runtimes/CMakeLists.txt
```diff
@@ -737,6 +734,10 @@ if(C_LIBRARY STREQUAL llvmlibc)
     )
 
     if(ENABLE_LIBC_TESTS)
+        if(NOT(TEST_EXECUTOR STREQUAL qemu))
```
vhscampos, line 737: If possible, do `if(TEST_EXECUTOR STREQUAL fvp)` instead. We could add other executors in the future.

Or if the tests only work with qemu, the error message should be changed to something like "Using LLVM libc with any executor other than qemu is unsupported".

vhscampos (approved):

PR #469, opened 2025-08-06
[ATfE] Enable compiler-rt tests on Armv8.1-M Main variants

vhscampos:
The compiler-rt tests on Armv8.1-M Main are passing, so it's time we enable them.


saturn691 (approved):
LGTM

PR #471, opened 2025-08-07
[ATfE] Enable picolibc, compiler-rt and libcxx texts for armv4t and armv5te exn_rtti

vhscampos:
The tests are passing now, so we can enable them.


MarkMurrayARM (approved):
LGTM.

PR #472, opened 2025-08-07
[ATfE]: Fix Windows llvmlibc sample to allow exception handling.

vrukesh:
[ATfE]: Fix Windows llvmlibc sample to allow exception handling.


saturn691 (changes requested):
Remove the `:` in the title

saturn691 (approved):

PR #473, opened 2025-08-07
[ATfE] Executor to run a binary with either of QEMU or FVP

vrukesh:
[ATfE] Executor to run a binary with either of QEMU or FVP


arm-software/embedded/arm-runtimes/test-support/generic-test-executor.py
```diff
@@ -0,0 +1,145 @@
...
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+
+# This is a generic executor script to execute tests with the supported models, such as QEMU or FVPs.
+
+from run_qemu import run_qemu
+from run_fvp import run_fvp
+import argparse
+import os
```
dcandler, line 16: `os` is unused, so there is no reason to import it.

arm-software/embedded/arm-runtimes/test-support/generic-test-executor.py
```diff
@@ -0,0 +1,145 @@
...
+import pathlib
+import sys
+
+
+def run(args):
+    # Some tests expect argv[0] to be literally "program-name", not
+    # the actual program name.
+    argv = ["program-name"] + args.arguments
```
dcandler, line 24: > Some tests expect argv[0] to be literally "program-name"

Which tests? What effect will this have on the tests that don't expect it?

arm-software/embedded/arm-runtimes/test-support/generic-test-executor.py
```diff
@@ -0,0 +1,145 @@
...
+        help='list of arguments to pass to qemu, separated with ":"',
+    )
+    parser.add_argument(
+        "arguments",
+        nargs=argparse.REMAINDER,
+        default=[],
+        help="optional arguments for the image",
+    )
+    args = parser.parse_args()
+    print("args: " + str(args))
```
dcandler, line 119: This looks like it should be debug output, not something that is always printed.

arm-software/embedded/arm-runtimes/test-support/generic-test-executor.py
```diff
@@ -0,0 +1,145 @@
...
+    # Top-level required argument
+    parser.add_argument(
+        "--test_executor",
+        required=True,
+        choices=["fvp", "qemu"],
+        help="Executor type: fvp or qemu",
+    )
+
+    # Common arguments
+    parser.add_argument("--image", help="image file to execute")
```
dcandler, line 66: Since `--image` is common and required across both executors, could it not be a positional argument? The image name followed any arguments to be passed to it would be a more natural way to run something.

arm-software/embedded/arm-runtimes/test-support/generic-test-executor.py
```diff
@@ -0,0 +1,145 @@
...
+        help="model name for FVP",
+    )
+    parser.add_argument(
+        "--fvp_config",
+        action="append",
+        help="FVP config file(s) to use",
+    )
+    parser.add_argument(
+        "--tarmac",
+        help="file to wrote tarmac trace to (FVP only)",
```
dcandler: Spelling: write

dcandler (changes requested):
I'm not keen on adding yet another wrapper script which duplicates most of the same functionality which is already available in another script. Especially as there is nothing using this script in our existing testing, which will make it harder to maintain.

Given how much overlap there is with the picolibc test script, it would make more sense to me to consolidate the functionality instead of splitting it further. The picolibc, lit-exec-fvp, lit-exec-qemu and now this generic script all essentially do the same thing, which is juggle around command options and arguments for the underlying script.

arm-software/embedded/arm-runtimes/test-support/generic-test-executor.py
vrukesh: Fixed !

vrukesh:
Planning to use scripts run_fvp or run_qemu directly.

PR #478, opened 2025-08-11
[ATfE] Enable all v7-M, v8-M, and v8.1-M variants for LLVM-libc tests

saturn691:
#420 introduced hermetic testing for LLVM-libc. However, it did not enable all variants. Now we enable all possible variants that compile, to prepare for use in pre-merge scripts.


MarkMurrayARM (approved):
This may be trouble. But break early, break often.

PR #479, opened 2025-08-11
[ATfE] Enable LLVM-libc tests in nightly

saturn691:
LLVM-libc tests can now be run for a lot of variants. This is experimental, however, there should be no reason why these tests won't run.

### DIFF
--- a/arm-software/embedded/scripts/test_llvmlibc.sh
+++ b/arm-software/embedded/scripts/test_llvmlibc.sh
@@ -6,7 +6,6 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 # A bash script to run the LLVM-libc tests from the Arm Toolchain for Embedded.
-
 # This script should be deleted when the generic test.sh script works
 
 set -ex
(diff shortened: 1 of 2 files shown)


arm-software/embedded/scripts/test_llvmlibc.sh
```diff
@@ -0,0 +1,34 @@
...
+# The check-all target runs the upstream clang and LLVM tests,
+# which do not generate an junit xml results file by default.
+# Additionally setting the --xunit-xml-output option store the
+# results.
+export LIT_OPTS="--ignore-fail --xunit-xml-output=results.xml"
+ninja check-all
+
+# The llvm-toolchain targets already set --xunit-xml-output so
+# only the --ignore-fail option is needed.
+# The picolibc tests do not use lit so do not support this option.
```
dcandler: If this script is solely for llvmlibc, then it shouldn't still be referencing picolibc.

arm-software/embedded/scripts/test_llvmlibc.sh
```diff
@@ -0,0 +1,34 @@
...
+export LIT_OPTS="--ignore-fail"
+ninja check-llvm-toolchain
```
dcandler: `check-llvm-toolchain` currently runs the libc tests, the compiler-rt tests, and the ATfE lit tests. The compiler-rt tests are also disabled in llvmlibc builds, and the multilib tests fail. So it may be better just to run `check-llvmlibc` if that's the only suite expected to work.

voltur01 (approved):

PR #481, opened 2025-08-11
[elf2bin] Exclude zero-length segments from --bincombined.

statham-arm:
(And also from --vhxcombined, which uses the same underlying code, but that didn't fit in a sensible-length subject line.)

Last week I found Arm Toolchain had generated an ELF image in which three segments had p_paddr in the bottom few Kb of the address space, but there was also a segment with zero p_filesz (but nonzero p_memsz), with p_paddr > 0x20000000. Running that through `elf2bin --physical --bincombined`, it didn't notice that the segment at a high address had length 0, so it padded the output file to half a Gb in size to reach the start address, and then didn't actually write anything there.

I worked around it at the time by selecting a subset of the segments to put in the --bincombined output file. But I think a better answer is to discard the zero-length segment completely when writing --bincombined output, because the padding in the output file is only there to get the file position to the place where some actually important data needs to be written. If there _is_ no important data to put at that position, there's no need to go there at all.

(Incidentally, I think I know why my image came out like that. The target memory map had Flash at 0, and RAM at 0x20000000. The code segments were in Flash, and the initialized data segment had its p_paddr in Flash but its p_vaddr at 0x20000000, representing respectively where the startup code would memcpy it from and to. The segment with p_filesz==0 represented the program's zero-initialized data, so its p_vaddr was in RAM, representing where the data would be at run time – but since it didn't need to be memcpy()ed from anywhere, the linker had left its p_paddr the same as its p_vaddr, and not bothered to allocate it a different one. In other words, the linker felt that if the segment was 0 bytes long then it didn't need a meaningful p_paddr at all.)

The new behavior adapts appropriately if you use the --zi option: in that situation, a segment with zero p_filesz but nonzero p_memsz doesn't count as empty, because the user _does_ want zero bytes in the --bincombined output file at its address.

### DIFF
--- a/arm-software/shared/elf2bin/bin.cpp
+++ b/arm-software/shared/elf2bin/bin.cpp
@@ -210,8 +210,19 @@ static std::unique_ptr<BinaryDataStream>
 combined_prepare(const InputObject &inobj,
                  const std::vector<Segment> &segments_orig, bool include_zi,
                  std::optional<uint64_t> baseaddr) {
+  // Discard segments that contribute no actual bytes to the file, in case one
+  // of those appears at the highest address in the list and causes trailing
+  // padding of the binary output file.
+  std::vector<Segment> filtered;
+  std::copy_if(segments_orig.begin(), segments_orig.end(),
+               std::back_inserter(filtered),
+               [include_zi](const Segment &seg) {
+                 uint64_t size = include_zi ? seg.memsize : seg.filesize;
+                 return size != 0;
+               });
+
   // Sort the segments by base address, in case they weren't already.
-  std::vector<Segment> sorted = segments_orig;
+  std::vector<Segment> sorted = std::move(filtered);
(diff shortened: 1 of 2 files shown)


smithp35 (approved):
LGTM. 

I think that in the unlikely event of someone trying to take advantage of the old behaviour to create a load of trailing zeroes; they could always replicate it by creating a non-zero sized segment containing zeroes and altering the start address so that it ended in the right place.

statham-arm:
I did try to think of any reason people might have been depending on the old behavior, but couldn't come up with one at all! If I had, I might have put in an option to revert to including zero-length segments. But I think it can wait until (unless) someone asks.

(And if they do ask, maybe there will be a better approach in any case. For example, if someone turns out to need a file _exactly_ the size of their ROM rather than _at most_ that size, then perhaps an option like `--bincombined-size=0x10000` would be more convenient than making a zero-length segment and telling elf2bin not to ignore it.)

PR #486, opened 2025-08-14
[ATfE] Enable v7-A and v7-R variants for LLVM-libc tests

saturn691:
Support added in https://github.com/llvm/llvm-project/pull/153576

Also update bootcode to support v7-A/v7-R outside of hermetic tests.

### DIFF
--- a/arm-software/embedded/llvmlibc-support/crt0/exceptions_7a.h
+++ b/arm-software/embedded/llvmlibc-support/crt0/exceptions_7a.h
@@ -95,7 +95,9 @@ EXFN_ATTR void handle_fiq() {
 // bytes long, and contains code. The whole table must be 32-byte aligned.
 // The table may also be relocated, so we make it position-independent by
 // having a table of handler addresses and loading the address to pc.
-__attribute__((naked, section(".vectors"), aligned(32))) void vector_table() {
+[[gnu::section(".vectors"), gnu::aligned(32), gnu::used, gnu::naked,
+  gnu::target("arm")]]
+void vector_table() {
   asm("LDR pc, [pc, #24]");
   asm("LDR pc, [pc, #24]");
   asm("LDR pc, [pc, #24]");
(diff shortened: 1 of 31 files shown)


arm-software/embedded/llvmlibc-support/crt0/exceptions_7a.h
```diff
@@ -95,7 +95,9 @@ EXFN_ATTR void handle_fiq() {
 // bytes long, and contains code. The whole table must be 32-byte aligned.
 // The table may also be relocated, so we make it position-independent by
 // having a table of handler addresses and loading the address to pc.
-__attribute__((naked, section(".vectors"), aligned(32))) void vector_table() {
+[[gnu::section(".vectors"), gnu::aligned(32), gnu::used, gnu::naked,
```
vhscampos, line 98: Why the change in syntax here?

arm-software/embedded/llvmlibc-support/crt0/exceptions_a.h
```diff
@@ -35,8 +35,12 @@ void setup() {
   // The vector table is always at address 0. The inline assembly is needed
   // here to hide the memcpy to a null pointer from the compiler.
   void *final_vector_table = NULL;
-  asm("" : "+r"(final_vector_table));
-  memcpy(final_vector_table, (void *)&vector_table, 64);
+  // Don't use memcpy as llvmlibc memcpy will fault a copy to address 0.
+  // Size is more important than speed here so don't unroll the loop.
+  _Pragma("clang loop unroll(disable) vectorize(disable)") for (
```
vhscampos, line 40: As far as I can tell, `memcpy` itself has unrolled loops, so I think you shouldn't worry about it.

arm-software/embedded/llvmlibc-support/crt0/exceptions_7a.h
saturn691: Honestly most of the code should use this `[[c++11]]` type syntax. It's more modern.

arm-software/embedded/llvmlibc-support/crt0/exceptions_a.h
saturn691: This is to prevent an internal LLVM-libc issue (see line 38)

vhscampos: No, I mean the pragma to prevent loop unrolling. Since even `memcpy` itself uses unrolling, having this pragma here is unnecessary.

arm-software/embedded/llvmlibc-support/crt0/exceptions_7a.h
vhscampos: But this is C++ syntax then, not C.

I'm not against it, but we must be sure that this will be compatible with all configurations, especially with a C-only compilation.

saturn691: Yes, this attribute syntax has been backported, at least on Clang. Also, this library is compiled on our systems, and this attribute will not affect user code at all, so as long as we don't use very old versions that are unsupported by Clang, we should be OK.

https://stackoverflow.com/questions/74574618/old-style-vs-c11-attribute-syntax

arm-software/embedded/llvmlibc-support/crt0/exceptions_a.h
saturn691: Oh I see what you mean. Here I'm removing the call to `memcpy` and using a manual copy.

PR #487, opened 2025-08-14
[ATfE] Set xfails indirectly instead of modifying test sources

dcandler:
Certain tests will fail when run with ATfE's compiler and libraries. The toolchain currently modifies the test sources to change the expected result, but this requires maintaining downstream patches. The LLVM test driver, lit, provides options to override the expected result set in the file through command line and environment variables, which would remove this burden.

This patch makes use of the `LIT_XFAILS` and `LIT_XFAILS_NOT` environment variables, which will take precedence over what is set in the sources.

A new script is provided which will generate a list of tests for a given project/library variant that are expected to give a different result. This is largely based on the existing set of downstream patches. This also has the advantage of allowing xfails to be set for specific variants, instead of globally. This currently only supports the clang, compiler-rt and libcxx projects.

In order to prepare the correct environment for lit, a wrapper script is also provided which will accept files containing test lists to be loaded into the relevant environment variable. Since the test lists can be quite long, this prevents the command line getting significantly bloated.


arm-software/embedded/arm-runtimes/CMakeLists.txt
```diff
@@ -267,9 +267,24 @@ if(VARIANT STREQUAL "armv6m_soft_nofp")
...
+    # Use the downstream version of lit, which is just a wrapper for the built one.
+    # Pass it the xfail/xfail-not lists.
+    set(compiler_rt_xfails_file_path "${CMAKE_CURRENT_BINARY_DIR}/${VARIANT}_compiler-rt.xfails")
+    set(compiler_rt_xfails_not_file_path "${CMAKE_CURRENT_BINARY_DIR}/${VARIANT}_compiler-rt.upasses")
     set(compiler_rt_test_flags "${lib_compile_flags} ${test_link_flags}")
-    set(compiler_rt_lit_args "${LLVM_LIT_ARGS} --path=${LLVM_BINARY_DIR}/bin --xunit-xml-output=results.junit.xml")
-    set(external_lit_path "${LLVM_BINARY_DIR}/bin/llvm-lit")
+    set(compiler_rt_lit_args "${LLVM_LIT_ARGS} --path=${LLVM_BINARY_DIR}/bin --xunit-xml-output=results.junit.xml --real-lit-path=${LLVM_BINARY_DIR}/bin/llvm-lit --xfails-file=${compiler_rt_xfails_file_path} --xfails-not-file=${compiler_rt_xfails_not_file_path}"
+    )
+    set(external_lit_path "${CMAKE_CURRENT_SOURCE_DIR}/test-support/lit-wrapper.py")
```
vhscampos, line 277: Why is the CMake variable name different here? From `LLVM_DEFAULT_EXTERNAL_LIT`

arm-software/embedded/arm-runtimes/CMakeLists.txt
```diff
@@ -788,7 +804,23 @@ if(ENABLE_CXX_LIBS)
             -DLIBCXX_ENABLE_RTTI=${ENABLE_RTTI}
         )
         if(ENABLE_LIBCXX_TESTS)
-            set(cxxlibs_lit_args "${LLVM_LIT_ARGS} --xunit-xml-output=results.junit.xml")
+            # Use the downstream version of lit, which is just a wrapper for the built one.
+            # Pass it the xfail/xfail-not lists.
+            set(libcxx_xfails_file_path "${CMAKE_CURRENT_BINARY_DIR}/${VARIANT}_libcxx.xfails")
+            set(libcxx_xfails_not_file_path "${CMAKE_CURRENT_BINARY_DIR}/${VARIANT}_libcxx.upasses")
+            set(external_lit_path "${CMAKE_CURRENT_SOURCE_DIR}/test-support/lit-wrapper.py")
```
vhscampos, line 811: Ditto

arm-software/embedded/arm-runtimes/test-support/lit-wrapper.py
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env python3
+
+# SPDX-FileCopyrightText: Copyright 2024-2025 Arm Limited and/or its affiliates <open-source-office@arm.com>
+
+# This is a wrapper script for LLVM's lit, which takes additional options in
+# order to prepare a specific environment for running tests.
+# The new --xfails-file and --xfails-not-file options are functionally similar
```
vhscampos: We should consider upstreaming these new options as they are quite useful

arm-software/embedded/arm-runtimes/test-support/lit-wrapper.py
```diff
@@ -0,0 +1,64 @@
...
+parser.add_argument("--xfails-file")
+parser.add_argument("--xfails-not-file")
+parser.add_argument("--real-lit-path", required=True)
+# Only parse the known args above. The rest should be processed the actual lit.
+known_args, unknown_args = parser.parse_known_args()
+
+# If an xfails file is set, read it.
+if known_args.xfails_file:
+    if os.path.isfile(known_args.xfails_file):
+        with open(known_args.xfails_file, "r") as f:
```
vhscampos: Consider using a linter on this file. I believe pylint would for instance complain about the lack of the `encoding` argument here.

arm-software/embedded/arm-runtimes/test-support/lit-wrapper.py
```diff
@@ -0,0 +1,64 @@
...
+            os.environ["LIT_XFAIL_NOT"] = (
+                os.environ["LIT_XFAIL_NOT"] + ";" + lit_xfails_list
+            )
+        else:
+            os.environ["LIT_XFAIL_NOT"] = lit_xfails_list
+    else:
+        print("file missing?")
+        raise FileNotFoundError(
+            f"Error: Expected file '{known_args.xfails_not_file}' does not exist."
+        )
```
vhscampos: Can you extract the common code into a function?

vhscampos: Making myself clearer, I mean the common code between `if known_args.xfails_not_file:` and `if known_args.xfails_file:`

arm-software/embedded/arm-runtimes/CMakeLists.txt
dcandler: `LLVM_DEFAULT_EXTERNAL_LIT` is a configuration option for the LLVM CMake project. The ATfE project includes the LLVM project as a subdirectory, so setting `LLVM_DEFAULT_EXTERNAL_LIT` as a variable in the same scope before the subdirectory is included has the effect of configuring the LLVM project with that option.

But the compiler-rt and libcxx projects are built as ExternalProjects, so the options can be set directly when the projects are added (though `-DLLVM_EXTERNAL_LIT=`). So the name here is just an internal name, not a name expected by another project.

arm-software/embedded/arm-runtimes/test-support/lit-wrapper.py
dcandler: Yes, that is my intention once we're happy with the mechanism downstream. I designed the wrapper to be easily removed if the options can be upstreamed, so all the would be required is not setting an external lit and removing the path to the upstream one.

arm-software/embedded/arm-runtimes/test-support/lit-wrapper.py
dcandler: Done, plus a few additional suggestions made by the linter.

arm-software/embedded/arm-runtimes/test-support/lit-wrapper.py
dcandler: Done

PR #492, opened 2025-08-20
[ATfE] Fix issue in semihosting with returning errors

saturn691:
This is required for testing, where we need to return a code of 1. Previously, semihosting would always exit with status OK no matter the return code for AArch32.

Add minor fixes with C++ syntax.

### DIFF
--- a/arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
+++ b/arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
@@ -16,9 +16,9 @@ namespace {
 
 void stdio_open(struct __llvm_libc_stdio_cookie *cookie, int mode) {
   size_t args[3];
-  args[0] = (size_t)":tt";
-  args[1] = (size_t)mode;
-  args[2] = (size_t)3; /* name length */
+  args[0] = reinterpret_cast<size_t>(":tt");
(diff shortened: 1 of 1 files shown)


arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
```diff
@@ -33,7 +33,13 @@ void __llvm_libc_exit(int status) {
   block[1] = status;
   semihosting_call(SYS_EXIT, block);
 #else
-  semihosting_call(SYS_EXIT, (const void *)ADP_Stopped_ApplicationExit);
+  if (status == 0) {
+    semihosting_call(
+        SYS_EXIT, reinterpret_cast<const void *>(ADP_Stopped_ApplicationExit));
+  } else {
+    semihosting_call(SYS_EXIT, reinterpret_cast<const void *>(
```
statham-arm, line 40: I had no idea that `ADP_Stopped_RunTimeErrorUnknown` would cause models to exit with a nonzero status. That's much simpler than figuring out whether you can use `SYS_EXIT_EXTENDED`!

It means you don't get full control over the exit status – all nonzero values are normalized to 1. But this is essentially free and better than nothing.

PR #499, opened 2025-08-26
[ATfE][NFC] Modernise llvmlibc-support to use C++ style code

saturn691:
C casts have been changed to `reinterpret_cast` or `static_cast`. Defines have been changed to `inline constexpr`. GNU attributes have been changed to `[[c++11]]` style attributes.

No functional change. Everything still works.

### DIFF
--- a/arm-software/embedded/llvmlibc-support/crt0/bootcode.cpp
+++ b/arm-software/embedded/llvmlibc-support/crt0/bootcode.cpp
@@ -46,12 +46,12 @@ int main(int argc, const char **argv);
 extern "C" void __libc_init_array();
 extern "C" void _platform_init();
 
-extern uintptr_t __data_source[];
-extern uintptr_t __data_start[];
-extern uintptr_t __data_size[];
-extern uintptr_t __bss_start[];
-extern uintptr_t __bss_size[];
-[[gnu::weak]] extern uintptr_t __stack;
+extern char __data_source[];
+extern char __data_start[];
+extern char __data_size[];
+extern char __bss_start[];
+extern char __bss_size[];
+[[gnu::weak]] extern char __stack;
 
 namespace {
 #ifdef __ARM_FEATURE_PAUTH
@@ -65,8 +65,8 @@ void do_start() {
   misc::setup();
 
   // Perform the equivalent of scatterloading
-  memcpy(__data_start, __data_source, reinterpret_cast<uintptr_t>(__data_size));
-  memset(__bss_start, '\0', reinterpret_cast<uintptr_t>(__bss_size));
+  memcpy(__data_start, __data_source, reinterpret_cast<size_t>(__data_size));
+  memset(__bss_start, '\0', reinterpret_cast<size_t>(__bss_size));
 
   memory::enable_cache();
   __libc_init_array();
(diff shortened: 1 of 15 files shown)


arm-software/embedded/llvmlibc-support/crt0/bootcode.cpp
```diff
@@ -46,49 +46,47 @@ int main(int argc, const char **argv);
...
-// TODO: This is temporary, remove once implemented upstream
-extern "C" [[gnu::weak]] void __cxa_finalize(void *) {}
-
-extern char __data_source[];
-extern char __data_start[];
-extern char __data_size[];
-extern char __bss_start[];
-extern char __bss_size[];
-extern char __stack __attribute__((weak));
+extern uintptr_t __data_source[];
```
statham-arm: Why change the type from `char[]` to `uintptr_t[]`?

The main difference seems to me that the new type is promising to the compiler that all these symbol values are multiples of 4 or 8. But I don't think we can guarantee that, so I think `char[]` was better.

arm-software/embedded/llvmlibc-support/semihost/semihost.h
```diff
@@ -52,56 +53,56 @@ static long semihosting_call(long val, const void *ptr) {
...
-#define SYS_RENAME 0x0f
-#define SYS_SEEK 0x0a
-#define SYS_SYSTEM 0x12
-#define SYS_TICKFREQ 0x31
-#define SYS_TIME 0x11
-#define SYS_TMPNAM 0x0d
-#define SYS_WRITE0 0x04
-#define SYS_WRITE 0x05
-#define SYS_WRITEC 0x03
+inline constexpr uint32_t SYS_CLOCK = 0x10;
```
statham-arm, line 57: Are you sure this means those symbols _must not_ be defined in object files including this header? If it had `static` as well then I'd believe it, but I have the feeling that `inline` and `constexpr` are both hints rather than mandatory.

arm-software/embedded/llvmlibc-support/crt0/memory_a.h
```diff
@@ -85,8 +85,9 @@ void setup() {
   // Put the page table in the .init section so it doesn't later get
   // zero-initialized.
   static_assert(sizeof(unsigned long) == PAGE_TABLE_ENTRY_SIZE);
-  static volatile unsigned long pagetable[PAGE_TABLE_ENTRY_COUNT]
-      __attribute__((section(".init"), aligned(PAGE_TABLE_ALIGNMENT)));
+  [[gnu::section(".init"),
+    gnu::aligned(PAGE_TABLE_ALIGNMENT)]] static volatile unsigned long
+      pagetable[PAGE_TABLE_ENTRY_COUNT];
```
statham-arm:
```suggestion
  [[gnu::section(".init"), gnu::aligned(PAGE_TABLE_ALIGNMENT)]]
  static volatile unsigned long pagetable[PAGE_TABLE_ENTRY_COUNT];
```

statham-arm: (Optional formatting nit: this is one line shorter _and_ seems more legible to me. But don't worry about it if you can't persuade clang-format to see it this way!)

arm-software/embedded/llvmlibc-support/semihost/semihost.h
statham-arm: OK, I'm happy that adding the anonymous namespace deals with my concern. Now these identifiers definitely won't end up in the object file under names that the application might want to use.

statham-arm (approved):
LGTM this time, thanks. I've left one more nitpick, but it's only formatting, and I won't worry about it if it's too awkward.

statham-arm (approved):

PR #507, opened 2025-09-01
[ATfE] rename VERSION.TXT in overlay packages

vrukesh:
[ATfE] rename VERSION.txt in overlay packages

The file currently has the same name in an overlay package as it does in the base toolchain that the overlay is installed on top of, so when unpacking the overlay package, the two files collide and the base toolchain's version is overwritten.

### DIFF
--- a/arm-software/embedded/CMakeLists.txt
+++ b/arm-software/embedded/CMakeLists.txt
@@ -569,25 +569,18 @@ add_custom_target(
     -P ${CMAKE_CURRENT_SOURCE_DIR}/cmake/generate_version_txt.cmake
     BYPRODUCTS ${CMAKE_CURRENT_BINARY_DIR}/VERSION.txt
 )
-if(LLVM_TOOLCHAIN_C_LIBRARY MATCHES "^llvmlibc")
+if(NOT (LLVM_TOOLCHAIN_C_LIBRARY STREQUAL picolibc))
     install(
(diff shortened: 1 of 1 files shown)


statham-arm (commented):
Commit message critique:

The subject line mentions `VERSION.TXT`, which doesn't exist. The file you're renaming is called `VERSION.txt`. File names are case sensitive on some platforms, so those aren't the same.

It would be somewhat useful to give a little more explanation following the subject line, just so that it's easy to see _why_ this file has to be renamed. (Maybe we have an internal bug tracker ticket that gives the reason? But that's not visible to everyone who can see this commit.)

How about:

> [ATfE] rename VERSION.txt in overlay packages
>
> The file currently has the same name in an overlay package as it does in the base toolchain that the overlay is installed on top of, so when unpacking the overlay package, the two files collide and the base toolchain's version is overwritten.

dcandler:
Or indeed update the script so that it generates the file with the correct name in the first place instead of needing to rename it later.

vrukesh:
@statham-arm 
Updated the description.

vrukesh:
@dcandler 
I attempted to do the same initially, it might have needed multiple changes in both generate_version_txt.cmake and CMakeLists.txt files.

Hence, we went with most simple and maintainable solution. Kindly let know if this approach is fine.

PR #508, opened 2025-09-01
[ATfE] Add reference to llvmlibc.ld in the LLVM-libc docs

saturn691:
Also change the command to something that will compile, as well as remove the `errno` comment, which is no longer true as of #436


smithp35 (approved):
I think this is fine to merge as is.

If there's a chance to refactor the section I'd prefer that we describe the musts as individual requirements. Then we can supply some command line options that satisfy the requirements. That might mean we can remove a lot of the (or you can write your own implementation). However that work shouldn't prevent this from merging.

PR #509, opened 2025-09-01
[ATfE] Disable parallel multilib configuration steps

MarkMurrayARM:
Overlay builds are being killed at config time due to an unknown reason. Disabling parallel config is a workaround that gets us green builds for now. This will be reverted when the underlying cause is fixed.


jagssa (dismissed):
Looks good. Matches comment from Jira request

saturn691 (changes requested):
Commit title can be shortened to 
> [ATfE] Disable multilib overlay builds

saturn691 (dismissed):

voltur01:
+1 to David's comment: this _is_ an llvmlibc overlay specific problem that needs a workaround.

PR #511, opened 2025-09-01
[multilib-generate.py] Fix parsing of AArch32 extensions

### LINKED ISSUE #484: multilib-generate.py failed to parse output of clang
**Summary**
In arm-software/embedded/arm-multilib/multilib-generate.py , function `get_extension_list()` assumes all features have prefix "FEAT_" but it is not true for arm-none-eabi . 
(aarch64-none-eabi seems to be so)

One can check by executing

```
ATfE/clang --target=arm-none-eabi --print-supported-extensions
ATfE/clang --target=aarch64-none-eabi --print-supported-extensions
```

Hence all arm-none-eabi features are not parsed.

Besides, if `LLVM_TARGETS_TO_BUILD` is set to only "ARM" (excluding AArch64) then 
`get_extension_list()` will get error while trying to query aarch64 features.

**To Reproduce**

The following feature flags are not collected into multilib-fpus.yaml
```
ATfE/clang --target=arm-none-eabi --print-supported-extensions
```

Also, building ATfE with `cmake -DLLVM_TARGETS_TO_BUILD=ARM ...` failed. Unless AArch64 is also appended to LLVM_TARGETS_TO_BUILD .

**Environment**

branch arm-software, commit 3314b3c575d7c7255cdffe3b4316e32bfbe0ccae
host: wsl-ubuntu

statham-arm:
The function `get_extension_list()` in this script parses the output of `clang --target=[...] --print-supported-extensions` to make a list of architecture extension names. It's supposed to work for both AArch32 and AArch64, but in fact it was ignoring all AArch32 extensions, because clang gives differently formatted output for the two architectures, the parsing code was specific to the AArch64 version, and it silently ignored any unrecognised line instead of reporting an error.

I've replaced the previous parser with one that can handle both formats. Also, it reads the output more thoroughly, and objects to anything it doesn't completely understand. So if clang changes the format of this output in future, we'll have to fix this script, but we'll at least know that, and won't find out that it's been quietly doing the wrong thing.

The new code is complicated enough that I've included a couple of unit tests, which you can run with the command
```
python3 -m unittest arm-software/embedded/arm-multilib/multilib-generate.py
```
But I haven't made that test run automatically; it's there to be run manually the next time someone edits the affected code. (A more important test is to ensure the script continues to be happy with clang's actual output, rather than the hardcoded example output in the unit tests.)

Fixes #484, I hope.

### DIFF
--- a/arm-software/embedded/arm-multilib/multilib-generate.py
+++ b/arm-software/embedded/arm-multilib/multilib-generate.py
@@ -41,8 +41,10 @@ which can be matched separately if a library requires them.
 import argparse
 import json
 import os
+import re
 import shlex
 import subprocess
+import unittest
 from dataclasses import dataclass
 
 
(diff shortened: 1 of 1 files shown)


MarkMurrayARM (approved):
Looks good!

PR #519, opened 2025-09-05
[ATfE] Fix LLVM-libc testing issue with CI

saturn691:
The Python script for running QEMU has some additional logic for `virt` machines, which cannot be easily written in CMake. Therefore, just use that, hopefully that will fix a bug in CI. Also pass `-k 0` directly to the underlying CMake, which will allow all tests to run.


vrukesh (approved):
LGTM

PR #521, opened 2025-09-08
[ATfE] Install ipxe-qemu for testing LLVM-libc

saturn691:
This is observed in CI, this patch should get the required files.

```
qemu-system-aarch64: failed to find romfile "efi-virtio.rom"
```

### DIFF
--- a/arm-software/embedded/scripts/install_dependencies.sh
+++ b/arm-software/embedded/scripts/install_dependencies.sh
@@ -17,7 +17,7 @@ sudo apt-get update && sudo apt-get install -y --no-install-recommends \
     python3-pip \
     python3-setuptools \
     qemu-system-arm=1:6.2+dfsg-2ubuntu6.26 \
-    ipxe-qemu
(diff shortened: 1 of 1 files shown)


arm-software/embedded/scripts/install_dependencies.sh
```diff
@@ -16,7 +16,8 @@ sudo apt-get update && sudo apt-get install -y --no-install-recommends \
     ninja-build=1.10.1-1 \
     python3-pip \
     python3-setuptools \
-    qemu-system-arm=1:6.2+dfsg-2ubuntu6.26
+    qemu-system-arm=1:6.2+dfsg-2ubuntu6.26 \
```
dcandler, line 19: Since a specific version of qemu is being installed, we may want to use a specific version for `ipxe-qemu` too for consistency in the builds.

This script appears to be targeting Ubuntu 22, so `1.21.1+git-20220113.fbbdc3926-0ubuntu1` would seem appropriate:
https://launchpad.net/ubuntu/jammy/amd64/qemu-system-arm
https://launchpad.net/ubuntu/jammy/amd64/ipxe-qemu

PR #523, opened 2025-09-10
[ATfE] Run picolibc tests via lit

dcandler:
The picolibc project is built and tested through meson, which differs from the other projects which use CMake to build and lit to run the tests. It would be useful to run all the tests using the same infrastructure, so that they can all be controlled in the same way. In particular, when tests are expected to fail, meson does not provide a mechanism to handle this, whereas lit does.

This patch allows the picolibc test suite to be executed through lit, by generating a lit test for every meson test. The existing lit controls for expected failures can then be used when running the generated test suite.

The overriding of return codes has been removed from the picolibc test wrapper, and instead added to the new list of expected failures. Only the FVP test failures were still relevant: the QEMU failures referenced variants that no longer exist, and the tests themselves actually pass now. Since the tests were treated as skipped, this has gone unnoticed. In future, if the state of a test changes, lit will flag that the result was unexpected.

A CMake target has been retained to optionally run the picolibc tests through meson for debugging purposes, since lit's interpretation of the results will differ from the unmodified results from meson.


arm-software/embedded/arm-runtimes/CMakeLists.txt
```diff
@@ -103,6 +103,10 @@ set(LLVM_BINARY_DIR "" CACHE PATH "Path to LLVM toolchain root to build librarie
 set(PROJECT_PREFIX "subproj" CACHE STRING "Directory to build subprojects in.")
 set(VARIANT_BUILD_ID "${VARIANT}" CACHE STRING "Name to use to identify build directories." )
 
+# A '.' character is used in junit xml to split classes/groups.
+# Variants such as armv8m.main need to be renamed.
+string(REPLACE "." "_" escaped_variant_name ${VARIANT})
```
saturn691: This variable is used once and doesn't have to be in the global scope.

arm-software/embedded/arm-runtimes/test-support/lit-wrapper.py
saturn691, line 1: What's going on here? Is this a chmod?

arm-software/embedded/arm-runtimes/test-support/gen-picolibc-lit-tests.py
```diff
@@ -0,0 +1,97 @@
+#!/usr/bin/env python3
+
+# SPDX-FileCopyrightText: Copyright 2024-2025 Arm Limited and/or its affiliates <open-source-office@arm.com>
```
saturn691: 2025

arm-software/embedded/arm-runtimes/test-support/gen-picolibc-lit-tests.py
```diff
@@ -0,0 +1,97 @@
...
+        # e.g. picolibc:semihost / semihost-argv
+        # The testnames should be unique, so only the name is needed.
+        subproj, full_testname = line.split(":", maxsplit=1)
+        testname = full_testname.split(" / ")[-1]
+        with open(
+            os.path.join(args.output, "tests", testname + ".test"),
+            "w",
+            encoding="utf-8",
+        ) as f:
+            # Invoke meson to rub the test.
```
voltur01:
```suggestion
            # Invoke meson to run the test.
```

arm-software/embedded/arm-runtimes/CMakeLists.txt
dcandler: My reasoning was that the variant name is in the global scope, therefore the escaped name should be as well. Plus it increases visibility that the escaped name may be needed.

Indeed, it shouldn't only be used once: my intention was that the same name should be used in the compiler-rt xml script as well to be consistent. Both scripts should use the same name instead of trying to independantly arrive at the same name. However it looks like I didn't commit that extra bit of code before pushing, so well spotted. I will include it in the fixup patch.

arm-software/embedded/arm-runtimes/test-support/lit-wrapper.py
dcandler: Correct, this adds execute permissions so that it matches llvm-lit and other python scripts in the repo.

arm-software/embedded/arm-runtimes/test-support/gen-picolibc-lit-tests.py
dcandler: Done

arm-software/embedded/arm-runtimes/test-support/gen-picolibc-lit-tests.py
dcandler: Done

arm-software/embedded/arm-runtimes/CMakeLists.txt
```diff
@@ -493,26 +499,66 @@ if(C_LIBRARY STREQUAL picolibc)
...
         ExternalProject_Add_Step(
             picolibc
-            check
-            COMMAND ${CMAKE_COMMAND} -P ${CMAKE_CURRENT_BINARY_DIR}/check-picolibc.cmake
+            gen-lit-tests
+            COMMAND ${Python3_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/test-support/gen-picolibc-lit-tests.py
+                --meson ${MESON_EXECUTABLE}
+                --build <BINARY_DIR>
+                --output ${picolibc_lit_dir}
+                --name picolibc-${escaped_variant_name}
```
vhscampos: Why is the escaped version of the variant name needed here?

arm-software/embedded/arm-runtimes/test-support/gen-picolibc-lit-tests.py
```diff
@@ -0,0 +1,97 @@
...
+
+"""Helper script to generate a lit test suite from a meson project's tests.
+
+The lit testing infratructure has a number of features that it would
+be useful to apply to meson test suite, such as filtering tests,
+setting particular tests to be expected to fail, and controlling
+whether or not testing should continue on a failure.
+Since meson can run tests individually and provide a list of all
+tests, it is possible to run the same tests through lit, with each
+individual test simply invoking meson
```
vhscampos: Missind '.' at the end of the paragraph.

arm-software/embedded/arm-runtimes/CMakeLists.txt
dcandler: The name for the testsuite is set here, and using a name with a `.` in it confuses the XML formatting of the results that are eventually produced because that character is used to split groups. E.g. instead of being treated as a single name, `picolibc-armv8.1m.main_hard_nofp_mve_size` is interpreted as `main_hard_nofp_mve_size` being a test suite in `1m` suite, within the `picolibc-armv8` suite.

In fact the escaped name really should be used for all testing where XML results are produced, but it hasn't been a problem yet for the libcxx tests because the armv8.1 variants aren't included there yet.

arm-software/embedded/arm-runtimes/test-support/gen-picolibc-lit-tests.py
dcandler: Done

arm-software/embedded/arm-runtimes/CMakeLists.txt
dcandler: I've renamed the variable to make it more self-evident why it is being used: the escaped name is there for when an xml-friendly name is requied.

PR #526, opened 2025-09-12
[Clang][Driver][ARM] Forward `-Os` and `-Oz` as multilib flags

vhscampos:
Pass along `-Os` and `-Oz` as multilib flags under ARM and AArch64 so
that they can be used as criteria for multilib selection.

This is a cherry-pick of 579a80784d43f4c21ac4ee2df221ba4f3b98fd70

Reason for downstream patch: this commit was merged in upstream LLVM
after the branching for 21 had been made. The patch is not a bug fix and
is very specific to Arm's interests, therefore I don't see this commit
being accepted for a backport.

Downstream issue:#528

### DIFF
--- a/clang/lib/Driver/ToolChain.cpp
+++ b/clang/lib/Driver/ToolChain.cpp
@@ -255,6 +255,20 @@ static void getAArch64MultilibFlags(const Driver &D,
     Result.push_back(ABIArg->getAsString(Args));
   }
 
+  // Start downstream change. Issue #528
+  if (const Arg *A = Args.getLastArg(options::OPT_O_Group);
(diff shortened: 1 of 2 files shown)


clang/lib/Driver/ToolChain.cpp
```diff
@@ -255,6 +255,20 @@ static void getAArch64MultilibFlags(const Driver &D,
     Result.push_back(ABIArg->getAsString(Args));
   }
 
+  // Start downstream change. Issue #258
```
dcandler: The issue number should be 528 (same elsewhere)

vhscampos: Fixed. Thanks

Issue #533, opened 2025-09-15
[Downstream Change] Amazon Linux still not recognized correctly


pawosm-arm:
**What is the change**
Proper recognition of Amazon Linux

**Why this change cannot be done upstream**
Political reasons I suppose

PR #535, opened 2025-09-15
[Downstream change] Cherry-pick of [TargetParser] Normalize "aarch64-amazon-linux" to "aarch64-amazon-linux-gnu"

pawosm-arm:
Original PR under review: llvm/llvm-project#140070

This is an one off change for the ATfL21 release.

Downstream issue: #533


pawosm-arm:
> I guess I'd be slightly concerned that the upstream patch is never actually going to get merged?

We're introducing it to the release branch only (following @MacDue recommendation). We decided this is how this should be addressed for this release.

pawosm-arm:
> > It's unclear
> 
> Now this is on the release branch only, thus is easier to accept as it will not cause us merge conflicts. i.e. maintenance, in the future.
> 
> We will need a separate issue and PR to implement this same or similar change on the main `arm-software` branch if it will be needed later. Do not have to decide now.

Fun fact, we (me and @kiranchandramohan) haven't been even discussing porting this change to the main `arm-software` branch. This is something we want for this release.

llvm/include/llvm/TargetParser/Triple.h
```diff
@@ -197,7 +197,8 @@ class Triple {
     SUSE,
     OpenEmbedded,
     Intel,
-    LastVendorType = Intel
+    Amazon,
```
simpal01, line 201: As per https://github.com/arm/arm-toolchain/blob/arm-software/CONTRIBUTING.md#downstream-patch-policy

`"The source change should be annotated with a comment including the #issue number>, if there are multiple lines changed a single comment for the block can be used." `

This is applicable to all of the following changes as well.”

pawosm-arm: Good point. Seems this PR needs some additional work, I'll take it later.

pawosm-arm: Done.

PR #542, opened 2025-09-23
[ATfE] Ensure folder exists when writing xfail files

dcandler:
If the destination folder does not exist yet, it can be safely created. Additionally, this removes a couple of lines that appear to be in the wrong place, attempting to write to an unopened file instead of printing the test names.

### DIFF
--- a/arm-software/embedded/arm-runtimes/test-support/xfails.py
+++ b/arm-software/embedded/arm-runtimes/test-support/xfails.py
@@ -507,7 +507,7 @@ def main():
     else:
         print("LIT_XFAIL=" + ";".join(tests_to_xfail))
     if args.xfails_not_file:
-        os.makedirs(os.path.dirname(args.xfails_file), exist_ok=True)
+        os.makedirs(os.path.dirname(args.xfails_not_file), exist_ok=True)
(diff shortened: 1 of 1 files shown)


arm-software/embedded/arm-runtimes/test-support/xfails.py
```diff
@@ -499,14 +500,14 @@ def check_r52_warning():
...
+        os.makedirs(os.path.dirname(args.xfails_file), exist_ok=True)
         with open(args.xfails_file, "w", encoding="utf-8") as f:
             for testname in tests_to_xfail:
                 f.write(testname + "\n")
     else:
         print("LIT_XFAIL=" + ";".join(tests_to_xfail))
-        for testname in tests_to_xfail:
-            f.write(testname + "\n")
     if args.xfails_not_file:
+        os.makedirs(os.path.dirname(args.xfails_file), exist_ok=True)
```
vhscampos: I guess it should be `args.xfails_not_file` as the argument to `os.path.dirname` here.

dcandler: Oops, well spotted. Fixed.

PR #544, opened 2025-09-23

pawosm-arm:

This PR reverts the commit introduced by #535 as it resulted in a faulty C++ compiler which cannot see system-provided standard C++ library. An alternative solution is being cherry-picked here instead which appears to work with Amazon Linux 2023.

This is an one off change for the ATfL21 release.

Downstream issue: #533


llvm/lib/TargetParser/Triple.cpp
```diff
@@ -1306,14 +1306,6 @@ std::string Triple::normalize(StringRef Str, CanonicalForm Form) {
   if (Vendor == Triple::SUSE && Environment == llvm::Triple::GNUEABI)
     Components[3] = "gnueabihf";
 
-  // Downstream issue: #533 (Amazon Linux still not recognized correctly)
```
pawosm-arm, line 1309: This is the part which breaks the frontend driver!

MacDue (approved):
I have no strong opinion on this change, this should also allow vectorizing sincos on Amazon Linux.

kiranchandramohan:
@pawosm-arm Could you clarify whether you are partially reverting https://github.com/llvm/llvm-project/pull/140070 and including changes from https://github.com/llvm/llvm-project/pull/136114?

Would it be possible to revert https://github.com/llvm/llvm-project/pull/140070 first and then create another PR for cherry-picking https://github.com/llvm/llvm-project/pull/136114?

At the moment there do not seem to be tests for this patch.

PR #547, opened 2025-09-25
Update llvmlibc.ld to match picolibc.ld

voltur01:
Update llvmlibc.ld to match picolibc.ld to simplify migration:
- Update the bootcode to use the same section name as picolibc (.text.init.enter) to simplify migration.
- Keep the old name (.vectors) for compatibility with upstream bootcode used in testing.
- Update the linker script template from latest picolibcpp.ld
- Remove -lcrt0-semihost from the llvmlibc test command line since upstream bootcode is used there.
- Highlight in the docs that LLVM libc linker script is licensed under a different license compared to LLVM libc itself.

### DIFF
--- a/arm-software/embedded/docs/llvmlibc.md
+++ b/arm-software/embedded/docs/llvmlibc.md
@@ -51,6 +51,11 @@ following command line options, in addition to `--target`, `-march` or
   or write your own linker script defining `__stack`, and
   `__llvm_libc_heap_limit` if you are using the heap
 
+> [!IMPORTANT]
+> The default `llvmlibc.ld` is provided for testing and is derived from the
+> `picolibc.ld` licensed under the BSD 3 Clause license. This may cause
+> licensing obligations if used in real projects.
+
 For example:
 
 ```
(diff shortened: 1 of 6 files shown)


saturn691 (approved):
LGTM, good catches as well!

arm-software/embedded/docs/llvmlibc.md
```diff
@@ -51,6 +51,11 @@ following command line options, in addition to `--target`, `-march` or
   or write your own linker script defining `__stack`, and
   `__llvm_libc_heap_limit` if you are using the heap
 
+> [!IMPORTANT]
+> The default `llvmlibc.ld` is provided for testing and is derived from the
+> `picolibc.ld` licensed under the BSD 3 Clause license. This may cause
+> licensing obligations if used in real projects.
```
HugoSilvaSantos, line 57: I think using the terms "commercial" or "production" projects might read better.

voltur01: Good point, I thought about it, however it does not have to be commercial - anything distributed as a binary - so I kept it vague on purpose. People need to read and decide for themselves if they are doing something "real", not just playing with a new toolchain.

PR #548, opened 2025-09-25
[ATfE] Create custom timeout mechanism for picolibc test runs

vhscampos:
picolibc defines timeout values for each test. In general these are enough for our test runs, however some variants can be slower due to less intensive code optimizations or security features. One example are the Armv8.1-M Main PACBTI variants optimized for code size.

For such variants, we need a way to specify timeouts values different than the ones set by the picolibc tests definitions.

This patch introduces a mechanism to set timeout multipliers for library variants. In the arm-runtimes' `CMakeLists.txt`, there's a new section where one can specify conditions based on library build options to determine different timeout multipliers. These multipliers are then used in the `gen-picolibc-lit-tests.py` script to generate the `meson test` command line to be used in the test runs. `meson test` takes the `-t` argument, which is the timeout multiplier.

For a limited number of custom timeout conditions, this approach scales without problems. But if in the future the list of custom conditions grows, it might be worth extracting the logic out of `CMakeLists.txt` into a new file in order to have a separation between build system's code and data.


arm-software/embedded/arm-runtimes/CMakeLists.txt
```diff
@@ -526,6 +545,7 @@ if(C_LIBRARY STREQUAL picolibc)
                 --build <BINARY_DIR>
                 --output ${picolibc_lit_dir}
                 --name picolibc-${variant_xml_name}
+                --timeout-multiplier ${timeout_multiplier}
```
dcandler, line 537: The multiplier should also be set in the meson command on line 507 above.

arm-software/embedded/arm-runtimes/test-support/gen-picolibc-lit-tests.py
```diff
@@ -44,6 +44,9 @@ def main():
         required=True,
         help="Name to give the lit suite.",
     )
+    arg_parser.add_argument(
+        "--timeout-multiplier", type=int, help="Timeout multiplier (integer)."
```
dcandler: I think meson accepts a float for the multiplier, so it would make sense if the script did too.

arm-software/embedded/arm-runtimes/CMakeLists.txt
```diff
@@ -518,6 +518,25 @@ if(C_LIBRARY STREQUAL picolibc)
         set(picolibc_xfails_not_file_path "${picolibc_lit_dir}/${VARIANT}_picolibc.upasses")
         set(picolibc_lit_args -sv --real-lit-path=${LLVM_BINARY_DIR}/bin/llvm-lit --xfails-file=${picolibc_xfails_file_path} --xfails-not-file=${picolibc_xfails_not_file_path})
 
+        # Query the timeouts file to see if a timeout multipler is needed.
```
dcandler: I'm not the biggest fan of CMake's json processing. Rather than creating a list of the variants and their timeouts, would it be better to do something rule based?

E.g. in this case a higher timeout is needed for the combination of pacbti and size optimized builds, which can be checked by the `COMPILE_FLAGS` and `LIBRARY_MESON_BUILD_TYPE` variables respectively:

```
if(LIBRARY_MESON_BUILD_TYPE STREQUAL "minsize" AND COMPILE_FLAGS MATCHES "-mbranch-protection=pac-ret\\+bti")
	set(timeout_multiplier "2.0")
else()
	set(timeout_multiplier "1.0")
endif()
```

That should make it a bit more self-explanatory what the conditions are for requiring the extra timeout, as well as making it more robust against changes to the variant names in future (or new variants/etc).

arm-software/embedded/arm-runtimes/CMakeLists.txt
vhscampos: Is the `check-meson` target/step used anywhere? I was even thinking of removing it in another patch.

dcandler: I left it in there for utility, since the results summary will be different from running the tests via lit compared to running meson directly. The translation between the two test frameworks is not perfect, so I felt it better to retain the ability to run the tests through meson so the additional information is still accessible and not hidden beneath an additional layer of scripts.

arm-software/embedded/arm-runtimes/CMakeLists.txt
vhscampos: My first intent was to use a yaml file, but that would have been even worse as CMake doesn't have a yaml parser.

A timeouts mechanism based on rules looks good. Done.

For just a single entry I believe that it can live inside CMakeLists.txt. But if in the future this list grows, I'd really rather have it in a separate place, and split build system's code from data. I suggest we keep that in mind.

arm-software/embedded/arm-runtimes/CMakeLists.txt
```diff
@@ -500,11 +500,19 @@ if(C_LIBRARY STREQUAL picolibc)
             compiler_rt-install
         )
 
+        # Timeout multipliers. The actual timeout value used in testing is:
+        # Timeout in seconds (defined in the picolibc's testing specs) x Timeout multiplier
+        if(LIBRARY_MESON_BUILD_TYPE STREQUAL "minsizerelease" AND COMPILE_FLAGS MATCHES "-mbranch-protection=pacret\\+bti")
```
vhscampos: FIXME: "pacret" should be "pac-ret"

vhscampos: And "minsizerelease" should be "minsize"

PR #549, opened 2025-09-25
Tep 1233 garcar02

garyjcarroll:
Reduce markdown heading levels in arm-software/linux/docs, all currently start at level 2

### DIFF
--- a/arm-software/linux/docs/GettingStarted.md
+++ b/arm-software/linux/docs/GettingStarted.md
@@ -1,10 +1,10 @@
-## Getting Started
+# Getting Started
 
 This document describes how to get started with the already installed Arm
 Toolchain for Linux. Here you will find out how to use it to compile a source
 code into an executable binary.
 
-### Looking at the toolchain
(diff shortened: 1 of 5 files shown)


arm-software/linux/docs/GettingStarted.md
```diff
@@ -191,12 +191,12 @@ $ armclang -c source2.c
 $ armclang -o binary source1.o source2.o
 ```
 
-#### Note
```
pawosm-arm, line 194: The plan was to annotate notes using some syntax extension, for example, github accepts this:
```
**_NOTE:_**  The note content.
```
When in doubt, please ask Philip Gee.

This applies to other `Note` places too.

garyjcarroll: I will add this change to the markdown here and update the internal processing to convert to the required publishing syntax.

garyjcarroll:
Agree with @pawosm-arm that he will make such structural changes to the markdown docs separately.

PR #550, opened 2025-09-26
[ATfE] Mark __stack declaration as extern "C"

voltur01:
The declaration is inside a C++ namespace, thus needs to be explicitly extern "C" to avoid name mangling. Further, all declarations of __start need to be updated to match.


smithp35 (approved):
LGTM. 

It looks like __heap_start may also need the same treatment. Anything coming from a linker script ought to be extern "C" unless we've intentially C++ mangled it in the script.

PR #551, opened 2025-09-26
[ATfE] Add QEMU tracing option

voltur01:
Add command line option to QEMU execution wrappers to enable execution tracing: per instruction disassembly and CPU state.

### DIFF
--- a/arm-software/embedded/arm-runtimes/test-support/run_qemu.py
+++ b/arm-software/embedded/arm-runtimes/test-support/run_qemu.py
@@ -16,6 +16,7 @@ def run_qemu(
     timeout,
     working_directory,
     verbose,
+    trace,
 ):
     """Execute the program using QEMU and return the subprocess return code."""
     qemu_params = ["-M", qemu_machine]
@@ -43,6 +44,13 @@ def run_qemu(
         qemu_params += ["-kernel", image]
     else:
         qemu_params += ["-device", f"loader,file={image},cpu-num=0"]
+
+    # Enable tracing: disassembly, CPU state, interrupts and guest errors like
+    # invalid instructions. One instruction at a time.
+    if trace:
+        qemu_params += ["-singlestep"]
+        qemu_params += ["-d", "in_asm,nochain,cpu,int,guest_errors"]
+        qemu_params += ["-D", trace]
 
     command = [qemu_command] + qemu_params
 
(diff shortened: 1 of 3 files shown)


arm-software/embedded/arm-runtimes/test-support/run_qemu.py
```diff
@@ -44,6 +45,13 @@ def run_qemu(
     else:
         qemu_params += ["-device", f"loader,file={image},cpu-num=0"]
 
+    # Enable tracing: disassembly, CPU state, interrupts and guest errors like
+    # invalid instructions. One instruction at a time.
+    if trace:
+        qemu_params += ["-singlestep"]
```
vhscampos: `-singlestep` is [deprecated](https://www.qemu.org/docs/master/about/removed-features.html#singlestep-removed-in-9-0).

voltur01: Thanks! A great spot and a bit of a problem: on Linux that we support QEMU is still v6 so it does not know the new option, but on macOS QEMU is already v10 so will not have the old option :(

Let me see.

voltur01: Looks like there is no good way to work this around without detecting QEMU version, which is rather awkward.

Probably best to remove for now. Maybe add as a comment to save people some searching later.

vhscampos: I think the singlestep feature is very important for debugging. I understand that detecting the version can feel awkward, but I'd still do it.

voltur01:
This follows the example of `--tarmac` for FVP - I do not see this passed centrally, I understand this is for debugging, i.e. when someone runs individual tests using the command line from the test run log.

Did I miss anything?

arm-software/embedded/arm-runtimes/test-support/run_qemu.py
voltur01: I gave it a try, please have a look, thanks!

arm-software/embedded/arm-runtimes/test-support/run_qemu.py
```diff
@@ -4,7 +4,15 @@
 
 import subprocess
 import sys
+import re
 
+def get_qemu_version(qemu_command):
```
vhscampos: I suggest you change the name to `get_qemu_major_version`.

voltur01: Good suggestion! Done.

dcandler:
LGTM, I was just clarifying the usage ahead of some other changes.

PR #557, opened 2025-09-30
Fix Mingw-w64 build

voltur01:
Fix Mingw-w64 build by adding missing elf2bin external project to Mingw-w64 build of LLVM Project.

Update the minimal version requirement, Ubuntu 24.04 version of Mingw-w64 is required to build LLVM now: afunix.h is reported missing in Ubuntu 22.04 version.


HugoSilvaSantos (approved):
LGTM!

PR #566, opened 2025-10-03
[ATfL] Enable sanitizers for testing ATfL

vrukesh:
[ATfL] Enable sanitizers for testing ATfL

### DIFF
--- a/arm-software/linux/scripts/build_atfl_sanitizer.sh
+++ b/arm-software/linux/scripts/build_atfl_sanitizer.sh
@@ -1,0 +1,85 @@
+#!/bin/bash
+
+# Copyright (c) 2025, Arm Limited and affiliates.
+# Part of the Arm Toolchain project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# A bash script to build the Arm Toolchain for Linux, with address and undefined sanitizer enabled.
+
+# Script implements 2-stage pipeline: first clang is built using arm-toolchain sources.
+# Then this clang is used to compile ATfL sanitizer build.
+#
+# The script creates a build of the toolchain in the 'build' directory, inside
+# the repository tree.
+
+set -vex
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+REPO_ROOT=$( git -C "${SCRIPT_DIR}" rev-parse --show-toplevel )
+
+clang --version
+
+export CC=clang
+export CXX=clang++
+
+# Stage 1: Compile clang
+echo "==> Stage 1: Starting clang build"
+
+mkdir -p "${REPO_ROOT}"/build_llvm
+cd "${REPO_ROOT}"/build_llvm
+
+cmake -G Ninja ../llvm \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ \
+    -DLLVM_LIT_ARGS="-v" \
+    -DCMAKE_INSTALL_PREFIX=../stage1.install \
+    -DLLVM_TARGETS_TO_BUILD=AArch64 \
+    -DLLVM_ENABLE_PROJECTS="clang-tools-extra;clang;llvm" \
+    -DLLVM_ENABLE_RUNTIMES="libcxx;libcxxabi;libunwind;compiler-rt"
+
+ninja
+
+echo "==> Stage 1: Completed clang build"
+
+# Stage 2: Compile library
+echo "==> Stage 2: Starting library build"
+
+mkdir -p "${REPO_ROOT}"/build_libcxx
+cd "${REPO_ROOT}"/build_libcxx
+
+export CC="${REPO_ROOT}/build_llvm/bin/clang"
+export CXX="${REPO_ROOT}/build_llvm/bin/clang++"
+
+cmake -G Ninja ../runtimes -DLLVM_TARGETS_TO_BUILD=AArch64 -DLLVM_USE_SANITIZER="Address;Undefined" -DLLVM_ENABLE_ASSERTIONS=True -DLLVM_ENABLE_RUNTIMES="libcxx;libcxxabi;libunwind;compiler-rt" -DCMAKE_C_COMPILER=$CC -DCMAKE_CXX_COMPILER=$CXX
+
+ninja
+
+echo "==> Stage 2: Completed library build"
+
+# Stage 3: Compile sanitizer build
+echo "==> Stage 3: Starting sanitizer build"
+
+mkdir -p "${REPO_ROOT}"/build_clang_with_sanitizer
+cd "${REPO_ROOT}"/build_clang_with_sanitizer
+
+export CC="${REPO_ROOT}/build_llvm/bin/clang"
+export CXX="${REPO_ROOT}/build_llvm/bin/clang++"
+export LD_LIBRARY_PATH="${REPO_ROOT}"/build_libcxx/lib:$LD_LIBRARY_PATH
+
+# Flag to disable memory leaks detection of LeakSanitizer.
+export ASAN_OPTIONS=detect_leaks=0
+
+ASAN_OPTIONS=detect_leaks=0 cmake -G Ninja ../llvm \
+
+    -DCMAKE_BUILD_TYPE=Release \
+    -DLLVM_USE_SANITIZER="Address;Undefined" -DLLVM_ENABLE_ASSERTIONS=True \
+    -DCMAKE_C_COMPILER=$CC -DCMAKE_CXX_COMPILER=$CXX \
+    -DLLVM_LIT_ARGS="-v" \
+    -DCMAKE_INSTALL_PREFIX=../stage1.install \
+    -DLLVM_TARGETS_TO_BUILD=AArch64 \
+    -DLLVM_ENABLE_PROJECTS="clang-tools-extra;clang;llvm" \
+    -DLLVM_ENABLE_RUNTIMES="libcxx;libcxxabi;libunwind;compiler-rt"
+
+echo "==> Stage 3: Completed sanitizer build"
+
--- a/arm-software/linux/scripts/test_atfl_sanitizer.sh
+++ b/arm-software/linux/scripts/test_atfl_sanitizer.sh
@@ -1,0 +1,37 @@
+#!/bin/bash
+
+# Copyright (c) 2025, Arm Limited and affiliates.
+# Part of the Arm Toolchain project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# A bash script to run the tests from the Arm Toolchain for Embedded, when sanitizers
+# are enabled.
+
+# The script assumes a successful build of the toolchain exists in the 'build'
+# directory inside the repository tree.
+
+# Script does not exit, when the command in the script exits with a non-zero status.
+# Also, print the line of the script being executed.
+set -vx
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+REPO_ROOT=$( git -C "${SCRIPT_DIR}" rev-parse --show-toplevel )
+
+cd "${REPO_ROOT}"/build_clang_with_sanitizer
+
+# If a test fails, lit will ordinarily return a non-zero result,
+# which prevents further testing. Setting the --ignore-fail option
+# will cause testing to continue, so that CI systems can get a
+# full set of results.
+# The check-all target runs the upstream clang and LLVM tests,
+# which do not generate an junit xml results file by default.
+# Additionally setting the --xunit-xml-output option store the
+# results.
+export LIT_OPTS="--ignore-fail --xunit-xml-output=results.xml"
+ninja
+ninja check-llvm
+ninja check-clang
+ninja check-cxx
+ninja check-cxxabi
+


arm-software/linux/scripts/build_atfl_sanitizer.sh
```diff
@@ -0,0 +1,85 @@
...
+# Stage 2: Compile library
+echo "==> Stage 2: Starting library build"
+
+mkdir -p "${REPO_ROOT}"/build_libcxx
+cd "${REPO_ROOT}"/build_libcxx
+
+export CC="${REPO_ROOT}/build_llvm/bin/clang"
+export CXX="${REPO_ROOT}/build_llvm/bin/clang++"
+
+cmake -G Ninja ../runtimes -DLLVM_TARGETS_TO_BUILD=AArch64 -DLLVM_USE_SANITIZER="Address;Undefined" -DLLVM_ENABLE_ASSERTIONS=True -DLLVM_ENABLE_RUNTIMES="libcxx;libcxxabi;libunwind;compiler-rt" -DCMAKE_C_COMPILER=$CC -DCMAKE_CXX_COMPILER=$CXX
```
pawosm-arm: Please split it into separate lines as it was done to the other `cmake` invocations.

arm-software/linux/scripts/build_atfl_sanitizer.sh
```diff
@@ -0,0 +1,89 @@
+#!/bin/bash
+
+# Copyright (c) 2025, Arm Limited and affiliates.
```
kiranchandramohan, line 3: Is adding the Copyright what we usually do? @voltur01 is this OK?

arm-software/linux/scripts/build_atfl_sanitizer.sh
```diff
@@ -0,0 +1,89 @@
...
+#
+# The script creates a build of the toolchain in the 'build' directory, inside
+# the repository tree.
+
+set -vex
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+REPO_ROOT=$( git -C "${SCRIPT_DIR}" rev-parse --show-toplevel )
+
+clang --version
```
kiranchandramohan: Is this left over code?

arm-software/linux/scripts/build_atfl_sanitizer.sh
```diff
@@ -0,0 +1,89 @@
...
+
+# Copyright (c) 2025, Arm Limited and affiliates.
+# Part of the Arm Toolchain project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# A bash script to build the Arm Toolchain for Linux, with address and undefined sanitizer enabled.
+
+# Script implements 2-stage pipeline: first clang is built using arm-toolchain sources.
+# Then this clang is used to compile ATfL sanitizer build.
```
kiranchandramohan, line 11: What about Flang? Are you concentrating on Clang only to start with?

arm-software/linux/scripts/build_atfl_sanitizer.sh
vrukesh: Yes, please refer other scripts in https://github.com/arm/arm-toolchain/tree/arm-software/arm-software/embedded/scripts

arm-software/linux/scripts/build_atfl_sanitizer.sh
vrukesh: Yes, idea is to get started with testing clang first in context of ATfL. We will add flang later.

arm-software/linux/scripts/build_atfl_sanitizer.sh
vrukesh: This is intentional implementation.

pawosm-arm: I presume printing clang version used when running this script can be useful, especially if this is started in a Jenkins pipeline where the output is being logged.

vrukesh:
@DavidSpickett 

Please review.

arm-software/linux/scripts/build_atfl_sanitizer.sh
```diff
@@ -0,0 +1,91 @@
+#!/bin/bash
+
+# Copyright (c) 2025, Arm Limited and affiliates.
+# Part of the Arm Toolchain project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# A bash script to build the Arm Toolchain for Linux, with address and undefined sanitizer enabled.
```
kiranchandramohan:
```suggestion
# A bash script to build the Arm Toolchain for Linux, with Address (ASan) and Undefined Behaviour Sanitizer (UBSan) enabled.
```

arm-software/linux/scripts/build_atfl_sanitizer.sh
```diff
@@ -0,0 +1,91 @@
...
+echo "==> Stage 3: Starting sanitizer build"
+
+mkdir -p "${REPO_ROOT}"/build_clang_with_sanitizer
+cd "${REPO_ROOT}"/build_clang_with_sanitizer
+
+export CC="${REPO_ROOT}/build_llvm/bin/clang"
+export CXX="${REPO_ROOT}/build_llvm/bin/clang++"
+export LD_LIBRARY_PATH="${REPO_ROOT}"/build_libcxx/lib:$LD_LIBRARY_PATH
+
+# Flag to disable memory leaks detection of LeakSanitizer.
```
kiranchandramohan:
```suggestion
# Flag to disable LeakSanitizer (memory leaks detection) in AddressSanitizer.
```

arm-software/linux/scripts/test_atfl_sanitizer.sh
```diff
@@ -0,0 +1,45 @@
...
+echo "check-llvm exit code: $?"
+
+ninja -v check-clang
+echo "check-clang exit code: $?"
+
+ninja -v check-cxx
+echo "check-cxx exit code: $?"
+
+ninja -v check-cxxabi
+echo "check-cxxabi exit code: $?"
```
kiranchandramohan, line 48: What about `clang-rt` and `libunwind`?

arm-software/linux/scripts/build_atfl_sanitizer.sh
```diff
@@ -0,0 +1,91 @@
...
+mkdir -p "${REPO_ROOT}"/build_llvm
+cd "${REPO_ROOT}"/build_llvm
+
+cmake -G Ninja ../llvm \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ \
+    -DLLVM_LIT_ARGS="-v" \
+    -DCMAKE_INSTALL_PREFIX=../stage1.install \
+    -DLLVM_TARGETS_TO_BUILD=AArch64 \
+    -DLLVM_ENABLE_PROJECTS="clang-tools-extra;clang;llvm" \
```
DavidSpickett: You don't need clang-tools-extra just for a toolchain used to build another toolchain. extra is stuff like the clang language server.

arm-software/linux/scripts/build_atfl_sanitizer.sh
```diff
@@ -0,0 +1,91 @@
...
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+REPO_ROOT=$( git -C "${SCRIPT_DIR}" rev-parse --show-toplevel )
+
+clang --version
+
+export CC=clang
+export CXX=clang++
+
+# Stage 1: Compile clang
+echo "==> Stage 1: Starting clang build"
```
DavidSpickett: Is this the syntax to create sections in the logs? IIRC GitHub has a format to do that. Fine if so, just asking in case you're reinventing that wheel.

I would call this something more specific, so that we know whether this part is sanitised or not. Hard to say what, it's not really the "host clang" because that's the clang that compiles this clang.

I might go with "Starting clang build (unsanitized)".

arm-software/linux/scripts/build_atfl_sanitizer.sh
```diff
@@ -0,0 +1,91 @@
...
+    -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ \
+    -DLLVM_LIT_ARGS="-v" \
+    -DCMAKE_INSTALL_PREFIX=../stage1.install \
+    -DLLVM_TARGETS_TO_BUILD=AArch64 \
+    -DLLVM_ENABLE_PROJECTS="clang-tools-extra;clang;llvm" \
+    -DLLVM_ENABLE_RUNTIMES="libcxx;libcxxabi;libunwind;compiler-rt"
+
+ninja
+
+echo "==> Stage 1: Completed clang build"
```
DavidSpickett: I don't think you need the "Completed..." bits. We know stage N has completed because it prints "stage N+1".

arm-software/linux/scripts/build_atfl_sanitizer.sh
DavidSpickett: Or whatever fits the later stages. Basically I want to be able to see where the santiziers are added and to what.
* Building clang (unsanitized)
* Building libraries (sanitized)
* etc etc

arm-software/linux/scripts/build_atfl_sanitizer.sh
```diff
@@ -0,0 +1,91 @@
...
+
+# Stage 2: Compile library
+echo "==> Stage 2: Starting library build"
+
+mkdir -p "${REPO_ROOT}"/build_libcxx
+cd "${REPO_ROOT}"/build_libcxx
+
+export CC="${REPO_ROOT}/build_llvm/bin/clang"
+export CXX="${REPO_ROOT}/build_llvm/bin/clang++"
```
DavidSpickett: This will work but if you know exactly where the compilers will be, pass the path to cmake instead.

Also, CMake will pick up $CC and $CXX without you setting -DCMAKE_C_COMPILER.... That's one advantage to environment variables but the drawback is that environment variables are global state and therefore more fragile.

Also imagine reading a log file where it prints the environment first then a million lines later the cmake config fails. It would be so much nicer if you didn't have to scroll back up. Doesn't really apply here but in general, that's a problem with environment variables.

arm-software/linux/scripts/build_atfl_sanitizer.sh
```diff
@@ -0,0 +1,91 @@
...
+
+cmake -G Ninja ../runtimes \
+    -DLLVM_TARGETS_TO_BUILD=AArch64 \
+    -DLLVM_USE_SANITIZER="Address;Undefined" \
```
DavidSpickett, line 46: This needs a comment to explain why the chosen sanitizers are run and in this combination. Add this same justification to the PR.

"because I haven't tried it" is a perfectly good reason to not enable something btw. If someone comes back to this in a year's time, they know it's on them to investigate it. Rather than wondering if they've missed some justification.

In this case, I think I told you to focus on undefined as the cheapest one. And ASAN because it's A: popular and B: seems to get most of the out of bounds stuff without being a massive pain to enable, as you found with MSAN.

arm-software/linux/scripts/build_atfl_sanitizer.sh
```diff
@@ -0,0 +1,91 @@
...
+    -DLLVM_ENABLE_ASSERTIONS=True \
+    -DLLVM_ENABLE_RUNTIMES="libcxx;libcxxabi;libunwind;compiler-rt" \
+    -DCMAKE_C_COMPILER="$CC" -DCMAKE_CXX_COMPILER="$CXX"
+
+ninja
+
+echo "==> Stage 2: Completed library build"
+
+# Stage 3: Compile sanitizer build
+echo "==> Stage 3: Starting sanitizer build"
```
DavidSpickett: And this one needs to be specific. Building a sanitized clang. Because "santizer" to me is `ninja compiler-rt`, which this is not.

You have way more than clang enabled, so you can say "sanitized toolchain" if that's more accurate here.

arm-software/linux/scripts/build_atfl_sanitizer.sh
```diff
@@ -0,0 +1,91 @@
...
+
+mkdir -p "${REPO_ROOT}"/build_clang_with_sanitizer
+cd "${REPO_ROOT}"/build_clang_with_sanitizer
```
DavidSpickett, line 58: Same for the folder path. Make the name correspond more precisely to what you're building.

This bootstrap process is confusing enough without us being accidentally vague.

arm-software/linux/scripts/build_atfl_sanitizer.sh
```diff
@@ -0,0 +1,91 @@
...
+
+export CC="${REPO_ROOT}/build_llvm/bin/clang"
+export CXX="${REPO_ROOT}/build_llvm/bin/clang++"
```
DavidSpickett: Again prefer passing the actual path to cmake.

arm-software/linux/scripts/build_atfl_sanitizer.sh
```diff
@@ -0,0 +1,91 @@
...
+export LD_LIBRARY_PATH="${REPO_ROOT}"/build_libcxx/lib:$LD_LIBRARY_PATH
```
DavidSpickett, line 60: This one potentially you can pass it via cmake, but I wouldn't object to it remaining an environment variable.

arm-software/linux/scripts/build_atfl_sanitizer.sh
```diff
@@ -0,0 +1,91 @@
...
+
+ninja
+
+echo "==> Stage 1: Completed clang build"
+
+# Stage 2: Compile library
+echo "==> Stage 2: Starting library build"
+
+mkdir -p "${REPO_ROOT}"/build_libcxx
+cd "${REPO_ROOT}"/build_libcxx
```
DavidSpickett, line 42: build_runtimes or build_libraries

This is not just building libcxx.

arm-software/linux/scripts/build_atfl_sanitizer.sh
```diff
@@ -0,0 +1,91 @@
...
+    -DLLVM_TARGETS_TO_BUILD=AArch64 \
+    -DLLVM_ENABLE_PROJECTS="clang-tools-extra;clang;llvm" \
+    -DLLVM_ENABLE_RUNTIMES="libcxx;libcxxabi;libunwind;compiler-rt"
+
+ninja
+
+echo "==> Stage 1: Completed clang build"
+
+# Stage 2: Compile library
+echo "==> Stage 2: Starting library build"
```
DavidSpickett: "libraries build (sanitized)" or something like that.

arm-software/linux/scripts/build_atfl_sanitizer.sh
```diff
@@ -0,0 +1,91 @@
...
+echo "==> Stage 3: Starting sanitizer build"
+
+mkdir -p "${REPO_ROOT}"/build_clang_with_sanitizer
+cd "${REPO_ROOT}"/build_clang_with_sanitizer
+
+export CC="${REPO_ROOT}/build_llvm/bin/clang"
+export CXX="${REPO_ROOT}/build_llvm/bin/clang++"
+export LD_LIBRARY_PATH="${REPO_ROOT}"/build_libcxx/lib:$LD_LIBRARY_PATH
+
+# Flag to disable LeakSanitizer (memory leaks detection) in AddressSanitizer.
```
DavidSpickett, line 62: because...

Don't leave it to future people, aka you, because you are now the expert, to wonder about this.

Again "it was too much hassle and created a bunch of spam" is ok, as long as there's something documented. Future people may disagree but at least they'll have something to disagree with.

DavidSpickett: Also you need to say where the issues come from. Do they come from the sanitized clang, the objdumper, the assembler, what?

DavidSpickett: Also cite any upstream issues if they exist.

arm-software/linux/scripts/build_atfl_sanitizer.sh
```diff
@@ -0,0 +1,91 @@
...
+export ASAN_OPTIONS=detect_leaks=0
+
+cmake -G Ninja ../llvm \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DLLVM_USE_SANITIZER="Address;Undefined" \
+    -DLLVM_ENABLE_ASSERTIONS=True \
+    -DCMAKE_C_COMPILER="$CC" -DCMAKE_CXX_COMPILER="$CXX" \
+    -DLLVM_LIT_ARGS="-v --param=env='ASAN_OPTIONS=detect_leaks=0'" \
```
DavidSpickett: Comment somewhere why this is repeated.

arm-software/linux/scripts/build_atfl_sanitizer.sh
```diff
@@ -0,0 +1,91 @@
...
+    -DCMAKE_INSTALL_PREFIX=../stage1.install \
+    -DLLVM_TARGETS_TO_BUILD=AArch64 \
+    -DLLVM_ENABLE_PROJECTS="clang-tools-extra;clang;llvm" \
```
DavidSpickett: Here clang-tools-extra is fine because it's something you include in the Arm Toolchain, I assume. Earlier, you weren't going to use it.

arm-software/linux/scripts/test_atfl_sanitizer.sh
```diff
@@ -0,0 +1,45 @@
...
+# directory inside the repository tree.
+
+# Script does not exit, when the command in the script exits with a non-zero status.
+# Also, print the line of the script being executed.
+set -vx
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+REPO_ROOT=$( git -C "${SCRIPT_DIR}" rev-parse --show-toplevel )
+
+cd "${REPO_ROOT}"/build_clang_with_sanitizer || exit
```
DavidSpickett, line 21: Make this a decent error rather than an abrupt exit.

if folder does not exist:
  stderr << "expected to find folder bla bla bla"
  exit 1

Also I think you need to exit 1, exit will have a zero return code and make this very mysterious when it does happen :)

arm-software/linux/scripts/test_atfl_sanitizer.sh
```diff
@@ -0,0 +1,45 @@
...
+export ASAN_OPTIONS=detect_leaks=0
+
+# If a test fails, lit will ordinarily return a non-zero result,
+# which prevents further testing. Setting the --ignore-fail option
+# will cause testing to continue, so that CI systems can get a
+# full set of results.
+# Upstream clang and LLVM tests do not generate the junit xml results file by default.
+# Additionally setting the --xunit-xml-output option store the
+# results.
+export LIT_ARGS="--ignore-fail --xunit-xml-output=lit_results.junit.xml --param=env='ASAN_OPTIONS=detect_leaks=0'"
```
DavidSpickett, line 33: If you were to put this in the configure line of the final sanitized toolchain build, you wouldn't need to repeat it here. `-DLLVM_LIT_ARGS=...`.

arm-software/linux/scripts/test_atfl_sanitizer.sh
```diff
@@ -0,0 +1,45 @@
...
+
+# Flag to disable memory leaks detection of LeakSanitizer.
+export ASAN_OPTIONS=detect_leaks=0
+
+# If a test fails, lit will ordinarily return a non-zero result,
+# which prevents further testing. Setting the --ignore-fail option
+# will cause testing to continue, so that CI systems can get a
+# full set of results.
+# Upstream clang and LLVM tests do not generate the junit xml results file by default.
+# Additionally setting the --xunit-xml-output option store the
```
DavidSpickett, line 31: Two things here.

1. Lit will overwrite this file every time it is run. We hit this upstream in CI, each "check" will run lit at least once, sometimes more often. For reasons I won't go into, we couldn't do `check-all` upstream.

Now there is a hidden option to lit which will create a unique name each time. However, it is hidden for a reason and I promised that it would likely remain so and be removed when we refactored the pipelines later (and as usual, later will probably be after I retire).

Nevertheless, it does exist.

That said, I'm not sure if you actually need them here. It is nice in Jenkins because it parses the files but idk about GitHub. Upstream we/I wrote a bunch of custom stuff to make test reports and you won't have that here (though, it is in the monorepo so in a way you do).

You could use check-all since you want to run all tests anyway, though this may be 2 lit invocations, one for projects and one for runtimes.

The other way to do this is what I proposed upstream at first:
```
ninja check-whatever
mv results.junit.xml <some unique prefix>.results.junit.xml
```
Then you can collect them later with `*.results.junit.xml`.

Since that's fiddly, I would suggest dropping the junit output from the first version of this script altogether. Maybe you don't even need it, and if you do, it's easier to add later.

2. You do this at the top:
```
# Script does not exit, when the command in the script exits with a non-zero status.
# Also, print the line of the script being executed.
set -vx
```
Then you also tell lit not to return non-zero exit codes. Surely you only need one?

arm-software/linux/scripts/test_atfl_sanitizer.sh
```diff
@@ -0,0 +1,45 @@
...
+export LIT_ARGS="--ignore-fail --xunit-xml-output=lit_results.junit.xml --param=env='ASAN_OPTIONS=detect_leaks=0'"
+
+ninja -v check-llvm
+echo "check-llvm exit code: $?"
+
+ninja -v check-clang
+echo "check-clang exit code: $?"
+
+ninja -v check-cxx
+echo "check-cxx exit code: $?"
```
DavidSpickett, line 45: If you're gonna split out the testing into stages, make it obvious somehow. Maybe with the `==>` stuff.

I say this because if `ninja check-bla` fails to build pre-requisites, the way this script is written, it'll get buried in the log and you'd never notice that it failed to run anything.

DavidSpickett: Also the way you report failure is not going to show up well. I'm not going to look for a print in the middle of a massive log, but I will look at the very last lines of a log.

(the junit would help too of course but if compilation fails you won't have junit results)

So I suggest that if you want to keep this multiple check- stage approach, accumulate the return code (or rather, latch it) and at the very end, set the script's return code to that value.

This way in the UI we'll get the failure theming for this step, and I can see right away oh, something failed. What failed? Well we have to dig for that but it's at least a starting point.

Basically:
* Declare some global variable return code, set it to 0
* After every command, if it returns a non-zero exit code, set the variable to 1 (in theory we could have many non-zero codes but we have to read the log anyway so it's not that important)
* At the end of the script `exit <that code>`.

arm-software/linux/scripts/build_atfl_sanitizer.sh
```diff
@@ -0,0 +1,91 @@
...
+
+# Copyright (c) 2025, Arm Limited and affiliates.
+# Part of the Arm Toolchain project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# A bash script to build the Arm Toolchain for Linux, with Address (ASan) and Undefined Behaviour Sanitizer (UBSan) enabled.
+
+# Script implements 2-stage pipeline: first clang is built using arm-toolchain sources.
+# Then this clang is used to compile ATfL sanitizer build.
```
DavidSpickett, line 11: This needs to be expanded to include something like:
```
* Latest clang (without sanitizers) built with host clang, then -
* libraries (with sanitizers) built with latest clang
```
And so on.

The two sentences you have are accurate, but are not enough to match to the code below.

I was also going to ask you to elaborate on this in the PR description but if you do something more comprehensive here, you can cite that instead.

arm-software/linux/scripts/test_atfl_sanitizer.sh
```diff
@@ -0,0 +1,45 @@
...
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+REPO_ROOT=$( git -C "${SCRIPT_DIR}" rev-parse --show-toplevel )
+
+cd "${REPO_ROOT}"/build_clang_with_sanitizer || exit
+
+# Flag to disable memory leaks detection of LeakSanitizer.
+export ASAN_OPTIONS=detect_leaks=0
+
+# If a test fails, lit will ordinarily return a non-zero result,
```
DavidSpickett, line 26: Well it would if you hadn't set the bash option above :)

arm-software/linux/scripts/build_atfl_sanitizer.sh
```diff
@@ -0,0 +1,91 @@
...
+#
+# The script creates a build of the toolchain in the 'build_clang_with_sanitizer' directory, inside
+# the repository tree.
+
+set -vex
```
DavidSpickett, line 16: And you set these because...?

arm-software/linux/scripts/build_atfl_sanitizer.sh
```diff
@@ -0,0 +1,91 @@
...
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+REPO_ROOT=$( git -C "${SCRIPT_DIR}" rev-parse --show-toplevel )
+
+clang --version
```
DavidSpickett: Cmake's configure step will print compiler versions for you, that's what I always look for if I need to know. You don't need an extra print here.

arm-software/linux/scripts/build_atfl_sanitizer.sh
```diff
@@ -0,0 +1,91 @@
...
+
+export CC=clang
+export CXX=clang++
```
DavidSpickett: Pass these via -DCMAKE_C_COMPILER.... instead.

arm-software/linux/scripts/build_atfl_sanitizer.sh
DavidSpickett: Another thing is, if I'm not using the script, an env var is one more thing I have to know to set. If I can copy paste a cmake command, so much easier.

DavidSpickett: And doing so doesn't leave my "favourite" terminal session in a weird state. People have enough trouble using the right compiler as it is :)

arm-software/linux/scripts/build_atfl_sanitizer.sh
```diff
@@ -0,0 +1,91 @@
...
+    -DLLVM_USE_SANITIZER="Address;Undefined" \
+    -DLLVM_ENABLE_ASSERTIONS=True \
+    -DLLVM_ENABLE_RUNTIMES="libcxx;libcxxabi;libunwind;compiler-rt" \
+    -DCMAKE_C_COMPILER="$CC" -DCMAKE_CXX_COMPILER="$CXX"
+
+ninja
+
+echo "==> Stage 2: Completed library build"
+
+# Stage 3: Compile sanitizer build
```
DavidSpickett: No need for a comment when I can read the next line.

arm-software/linux/scripts/test_atfl_sanitizer.sh
DavidSpickett: This is one problem with using `check-subproject`.

Your components will be pretty static, so if you do keep this strategy, test it by adding a test failure to each one that you care about and making sure it's reported.

vrukesh: @kiranchandramohan 
Added both.
No issues observed in pre-merge testing.

@DavidSpickett 
Since ATfL does not have xfails yet, it could be tricky to handle the failures. We can revisit this later.

arm-software/linux/scripts/build_atfl_sanitizer.sh
vrukesh: Updated in latest version.

arm-software/linux/scripts/build_atfl_sanitizer.sh
vrukesh: Updated in latest version.

arm-software/linux/scripts/build_atfl_sanitizer.sh
vrukesh: Updated in latest version.

arm-software/linux/scripts/test_atfl_sanitizer.sh
vrukesh: Got it. Thanks !

arm-software/linux/scripts/build_atfl_sanitizer.sh
vrukesh: Useful foe debugging.

-v (Verbose) : Print each line of the script before executing it.
-e (Exit on error) : Exit the script immediately if any command returns a non-zero exit code
-x (xtrace) : Print each command as it is executed, with variable expansions applied.

arm-software/linux/scripts/build_atfl_sanitizer.sh
vrukesh: Updated in latest version.

arm-software/linux/scripts/build_atfl_sanitizer.sh
vrukesh: Updated in latest version.

arm-software/linux/scripts/build_atfl_sanitizer.sh
vrukesh: Updated in latest version.

arm-software/linux/scripts/build_atfl_sanitizer.sh
vrukesh: Updated in latest version.

arm-software/linux/scripts/build_atfl_sanitizer.sh
vrukesh: Updated in latest version.

arm-software/linux/scripts/build_atfl_sanitizer.sh
DavidSpickett: Ok cool just standard stuff then. Though it might be something more nuanced, like in the testing script.

kiranchandramohan (approved):
LGTM.

@pawosm-arm hope this is OK with you.

PR #567, opened 2025-10-03
[ATfE] Fix llvmlibc C++ sample crash by increasing the stack

voltur01:
Fix llvmlibc C++ sample crash by increasing the stack size and align the picolibc sample linker script to use the same.


vrukesh (approved):
LGTM.

PR #568, opened 2025-10-03
[ATfL] These AArch64 definitions are available since the release 21.x

pawosm-arm:


kiranchandramohan (approved):
LG. Thanks @pawosm-arm

PR #570, opened 2025-10-03
[ATfE] Fix llvmlibc AArch64 sample

voltur01:
Fix llvmlibc AArch64 sample by providing the vector table a name that ensures correct placement following the latest (picolibc based) linker script. Without this, the vector table got before _start.

Increase the size of boot_flash to fit.


vrukesh (approved):
LGTM

PR #573, opened 2025-10-03
[ATfL] Try to get rid of an extra line rendered in PDF

pawosm-arm:
The extra line is added for some reason in PDF in case of all the items that span into more than one line.


pawosm-arm:
For experiments only, may not be merged.

kiranchandramohan (approved):
LG

PR #585, opened 2025-10-09
[ATfL][docs] Describe non-root installation procedure

pawosm-arm:


arm-software/linux/docs/Installation.md
```diff
@@ -87,3 +95,37 @@ $ sudo dnf install arm-toolchain-for-linux
...
+
+You can install the Arm Toolchain for Linux, along with the Arm Performance
+Libraries in any directory that you wish to specify. For this, the auxiliary
+installation script can be used. This script does not require system
+administrator (`root`) assistance:
+
+```
+$ bash <(curl -L https://developer.arm.com/-/cdn-downloads/permalink/Arm-Toolchain-for-Linux/Package/user_install.sh) --yes <installation_directory>
+```
+
```
schander-arm, line 109: I don't think we need anything after this. Given that the script outputs

`Installation complete. To use ATfL in your shell, run:
  # Add modulefiles to your MODULEPATH
  module use /home/satcha02/code/atlf/opt/arm/modulefiles

  # Load the ATfL compiler and Arm Performance Libraries
  module load atfl/20.1
  module load arm-performance-libraries

  # Verify the compiler is available
  armclang --version
  armclang++ --version`

pawosm-arm: Ok, so I've removed the overhead information.

PR #614, opened 2025-11-05
[ATfE] Fix invalid register error `ELR_hyp` in exception handling on Armv7-A

vhscampos:
The exception handling functions print out the value of the link register for logging purposes. The problem with this is in reading the `ELR_hyp` register.

This register is only accessible if the processor is in Hypervisor mode. This mode is always available on Armv8-A, but it is optional on Armv7-A, conditioned on the presence of the Virtualization Extension. Attempts to read/write to this register on the latter, if the Virtualization Extension is not in, cause a compiler error, and possibly a fault at runtime if it hadn't been caught by the compiler.

This patch fixes this by reading from the correct link register depending on the configuration. If the processor is not in Hypervisor mode, the code must read from `LR` instead of `ELR_hyp`.

In one of the handlers, the instruction pointed by the link register is printed out. In Thumb mode, the link register might contain an address that is not 4-byte aligned, therefore the load that retrieves the instruction is now done 2 bytes at a time to avoid misaligned access. We could have code that selectively loads either 4 bytes at once or 2 times 2 bytes depending on the ISA state, but it doesn't seem worth it.

### DIFF
--- a/arm-software/embedded/llvmlibc-support/crt0/exceptions_7a.h
+++ b/arm-software/embedded/llvmlibc-support/crt0/exceptions_7a.h
@@ -35,13 +35,8 @@ static bool is_in_hyp_mode() {
   return mode_bits == 0x1A;
 }
 
-#define GET_LR_VALUE_INTO(var)                                                 \
-  do {                                                                         \
-    if (is_in_hyp_mode())                                                      \
-      var = read_elr_hyp();                                                    \
-    else                                                                       \
-      var = (uint32_t)__builtin_return_address(0);                             \
-  } while (0)
+#define GET_LR_VALUE()                                                         \
+  (is_in_hyp_mode() ? read_elr_hyp() : (uintptr_t)__builtin_return_address(0))
 
 EXFN_ATTR void handle_reset() {
   print_str("CPU Exception: Reset\n");
@@ -49,8 +44,7 @@ EXFN_ATTR void handle_reset() {
 }
 
 EXFN_ATTR void handle_undefined() {
-  uintptr_t lr_val;
-  GET_LR_VALUE_INTO(lr_val);
+  uintptr_t lr_val = GET_LR_VALUE();
   print_str("CPU Exception: Undefined Instruction\n");
   print_str("  PC = ");
   print_hex(lr_val);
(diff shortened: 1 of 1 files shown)


arm-software/embedded/llvmlibc-support/crt0/exceptions_7a.h
```diff
@@ -20,36 +20,67 @@ namespace exceptions {
...
+  return mode_bits == 0x1A;
+}
+
+#define GET_LR_VALUE_INTO(var)                                                 \
+  do {                                                                         \
+    if (is_in_hyp_mode())                                                      \
+      var = read_elr_hyp();                                                    \
+    else                                                                       \
+      var = (uint32_t)__builtin_return_address(0);                             \
+  } while (0)
```
statham-arm: Is there a reason you didn't define this macro as an expression rather than a statement, along the lines of this?
```c
#define GET_LR_VALUE() (is_in_hyp_mode() ? read_elr_hyp() : (uintptr_t)__builtin_return_address(0))
```

Then you wouldn't have had to change all those one-line call sites `print_hex(...)` into three lines with a variable declaration.

arm-software/embedded/llvmlibc-support/crt0/exceptions_7a.h
```diff
@@ -20,36 +20,67 @@ namespace exceptions {
...
+  GET_LR_VALUE_INTO(lr_val);
   print_str("CPU Exception: Undefined Instruction\n");
   print_str("  PC = ");
-  print_hex(pc_val);
+  print_hex(lr_val);
   print_str("\n");
-  print_str("  Instruction = ");
-  print_hex(instr);
+  const uint16_t half_word0 = *(uint16_t *)lr_val;
+  const uint16_t half_word1 = *((uint16_t *)lr_val + 1);
```
statham-arm, line 53: This rewrite as two halfword loads instead of a single word load isn't mentioned in the commit message. I guess the idea is to avoid an unaligned load in Thumb mode?

arm-software/embedded/llvmlibc-support/crt0/exceptions_7a.h
vhscampos: Excellent suggestion. Much better now.

arm-software/embedded/llvmlibc-support/crt0/exceptions_7a.h
vhscampos: Added a paragraph to the commit msg. Thank you

statham-arm (approved):

PR #620, opened 2025-11-19
[ATFE][ARM] remove -enable-inline-memcpy-ld-st patch

nasherm:
Downstream patch #406 caused build failures due to an upstream change to the ARM backend. This patch removes the support enabled by this downstream patch as I no longer think it's worth maintaining for performance improvements. Further work should be done upstream to regain any lost performance.

Removes downstream issue: #405


stuij (dismissed):
LGTM

PR #642, opened 2025-12-10
[ATfE] Provide clang-scan-deps utility

voltur01:
Include clang-scan-deps utility into the package to enable C++20 module support.


kiranchandramohan:
@pawosm-arm @vrukesh  Is this required for ATfL as well?

PR #646, opened 2025-12-17
Make samples compatible with LLVM libc

voltur01:
Make samples compatible with LLVM libc by adding the LIBC option to set which library to use.

Remove llvmlibc-samples as they do not show anything extra now.

Note that LLVM libc iostream support is still pending, so only the very basic C++ sample was made compatible with LLVM libc - the rest should be enabled with the LLVM libc stdio work in progress.


arm-software/embedded/samples/Makefile.conf
```diff
@@ -15,3 +19,15 @@ AARCH64_TARGET=--target=aarch64-none-elf
...
+    CFG_FILE=
+    MATH_LIB=
+    LIBC_LD_FILE=picolibcpp.ld
+else ifeq ($(LIBC),llvmlibc)
+    CFG_FILE=--config=llvmlibc.cfg
+    MATH_LIB=-lm
+    LIBC_LD_FILE=llvmlibc.ld
+else
+    $(error Unsupported LIBC=$(LIBC))
+endif
```
saturn691: missing newline

saturn691:
```suggestion
endif

```

arm-software/embedded/samples/src/baremetal-uart/hello.c
```diff
@@ -51,11 +51,41 @@ int uart_putc(char ch, FILE* file)
...
 }
 
-/* Redirect sdtio as per https://github.com/picolibc/picolibc/blob/main/doc/os.md */
-static FILE __stdio = FDEV_SETUP_STREAM(uart_putc, NULL, NULL, _FDEV_SETUP_WRITE);
-FILE *const stdin = &__stdio;
-__strong_reference(stdin, stdout);
-__strong_reference(stdin, stderr);
+#if defined(__PICOLIBC_VERSION__) // built with picolibc
+
+  /* Redirect sdtio as per https://github.com/picolibc/picolibc/blob/main/doc/os.md */
```
saturn691:
```suggestion
  /* Redirect stdio as per https://github.com/picolibc/picolibc/blob/main/doc/os.md */
```

arm-software/embedded/samples/Makefile.conf
voltur01: Thank you, fixed!

PR #650, opened 2026-01-02
[ATfE] Enable iostream with llvm libc

voltur01:
After https://github.com/llvm/llvm-project/pull/168931 was merged with basic stdio implementation, it is now possible to enable iostream with llvm libc.

_LIBCPP_PROVIDES_DEFAULT_RUNE_TABLE seems to be defined here https://github.com/llvm/llvm-project/blob/7976ac990000a58a7474269a3ca95e16aed8c35b/libcxx/include/__config#L720 with __LLVM_LIBC__ however it is not enough - there may be a cleaner upstream fix for this.

### DIFF
--- a/arm-software/embedded/arm-runtimes/CMakeLists.txt
+++ b/arm-software/embedded/arm-runtimes/CMakeLists.txt
@@ -910,7 +910,7 @@ if(ENABLE_CXX_LIBS)
             -DLIBCXX_ENABLE_RTTI=${ENABLE_RTTI}
             -DRUNTIMES_USE_LIBC=system # llvm-libc logic in HandleLibC.cmake is not compatible with ATfE now
             )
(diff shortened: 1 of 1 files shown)


voltur01:
Turned out that problem is that `__LLVM_LIBC__` is not defined early enough to trigger the above mentioned code to add `_LIBCPP_PROVIDES_DEFAULT_RUNE_TABLE` - I replaced the macro defined since `__LLVM_LIBC__` is more generic and clear. 

In the future we should investigate how to enable `__LLVM_LIBC__` via `RUNTIMES_USE_LIBC` option that does not work now: https://github.com/arm/arm-toolchain/blob/a55b188cd82a12190d99310e0caacd5915cb904a/arm-software/embedded/arm-runtimes/CMakeLists.txt#L911

saturn691 (approved):

PR #657, opened 2026-01-07
[ATfE] Enable RUNTIMES_USE_LIBC=llvm-libc

voltur01:
With RUNTIMES_USE_LIBC=llvm-libc standard library includes are disabled, thus the sysroot provided is skipped - we need to add the libc include folder directly to build C++ libraries.


saturn691 (approved):
Happy if CI passes

PR #667, opened 2026-01-12
[ATfE] Use semihosting SYS_READC for stdin with llvm libc

voltur01:
Semihosting SYS_READ call on stdin just returns immediately indicating no data was read, thus rendering stdin unusable with QEMU.

SYS_READC blocks and waits for one symbol to be read from the terminal, however Enter key specifically is returned as '\r', thus the special case (that may not generalize well beyond QEMU).


arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
```diff
@@ -45,17 +45,22 @@ void __llvm_libc_exit(int status) {
...
-  size_t args[4];
-  args[0] = static_cast<size_t>(cookie->handle);
-  args[1] = reinterpret_cast<size_t>(buf);
-  args[2] = size;
-  args[3] = 0;
-  ssize_t retval = semihosting_call(SYS_READ, args);
-  if (retval >= 0)
-    retval = size - retval;
-  return retval;
+                               char *buf, size_t size) {
```
smithp35, line 72: I think it will be worth a comment explaining why we are using SYS_READC. Something like:
```
Currently only supports reading from stdin. We use SYS_READC for reading from stdin as QEMUs SYS_READ does not block. For other files SYS_READ should be used as SYS_READC may block indefinitely in QEMU.
```

smithp35 (commented):
Based on the Picolibc/QEMU discussion in https://gitlab.com/qemu-project/qemu/-/issues/1963, this looks good to me. It is mentioned that newlib only uses SYS_READ, which does indeed look like the case for libgloss/arm/syscalls.c however there may be higher-level code that is able to handle non-blocking IO in newlib, or it may just not work at all with QEMU!

arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
voltur01: Thank you! A comment added including a TODO for when there is regular file support in llvm libc.

PR #678, opened 2026-01-16
[ATfE] Replace call to abort with __llvm_libc_exit in libc startup code

voltur01:
LLVM libc does not provide a way to redefine abort(), thus the local abort() definition in LLVM libc semihosting has to be removed.

Calls to abort() in LLVM libc crt0 replaced with __llvm_libc_exit() which is the LLVM libc defined platform specific exit handler.

This prevents exception handlers from crashing because of the trap instruction in LLVM libc implementation of abort().

### DIFF
--- a/arm-software/embedded/llvmlibc-support/crt0/exceptions_7a.h
+++ b/arm-software/embedded/llvmlibc-support/crt0/exceptions_7a.h
@@ -40,7 +40,7 @@ static bool is_in_hyp_mode() {
 
 EXFN_ATTR void handle_reset() {
   print_str("CPU Exception: Reset\n");
-  abort();
+  __llvm_libc_exit(1);
 }
 
 EXFN_ATTR void handle_undefined() {
@@ -56,7 +56,7 @@ EXFN_ATTR void handle_undefined() {
   print_str("  Instruction (second half word) = ");
   print_hex(half_word1);
   print_str("\n");
-  abort();
+  __llvm_libc_exit(1);
 }
 
 EXFN_ATTR void handle_svc_hyp_smc() {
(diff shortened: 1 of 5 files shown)


vrukesh (approved):
LGTM

arm-software/embedded/llvmlibc-support/crt0/exceptions_7a.h
```diff
@@ -40,7 +40,7 @@ static bool is_in_hyp_mode() {
 
 EXFN_ATTR void handle_reset() {
   print_str("CPU Exception: Reset\n");
-  abort();
+  __llvm_libc_exit(1);
```
smithp35, line 43: From what I can see, we only have a definition of `__llvm_libc_exit()` in libsemihosting.a if we make this change then we will require a user of the support code to use `-lsemihosting`. If it is going to be a usage requirement we should update the user documentation for llvm libc.

I think `__llvm_libc_exit()` should be defined in `libcrt0.a` and `libcrt0-semihost.a` (not sure what `libcrt0-none.a` is for at the moment). The definition in libcrt0-semihost.a should exit via semihosting and the other definitions should either infinite loop or trap. 

Ideally libsemihosting.a should just be an opt in library contain default implementations of the functions needing retargeting to semihosting.

voltur01: Now we have at least two symbols required by `crt0` https://github.com/arm/arm-toolchain/blob/fd9008ab498e94a60facea56fbdfa11caea6c586/arm-software/embedded/samples/src/baremetal-uart/hello.c#L62
* `_platform_init`
* `__llvm_libc_exit`
both are defined by LLVM libc itself so have to be provided by either ATfE startup code or the user code like in the unhosted sample above.

Let me address this as a separate PR to also update the unhosted sample accordingly.

PR #683, opened 2026-01-19
[ATfE] Provide nohost init and exit in llvmlibc startup code

voltur01:
Provide init and exit handlers for not hosted environments in llvm libc crt0.

Update the documentation accordingly.

Update the UART sample to make use of the symbols provided in crt0 and refine the comments for the remaining symbols related to IO retargeting.

### DIFF
--- a/arm-software/embedded/llvmlibc-support/crt0/crt0-nohost.cpp
+++ b/arm-software/embedded/llvmlibc-support/crt0/crt0-nohost.cpp
@@ -11,31 +11,11 @@ extern "C" {
 // Do nothing for platform initialization
 [[gnu::weak]] void _platform_init() {}
 
-// Define architecture specific idle mode
-#if defined(__aarch64__)
-static inline void mask_irqs(void) {
-  __asm__ volatile("msr daifset, #0xf \n isb \n" ::: "memory");
-}
-#elif defined(__arm__) && (__ARM_ARCH_PROFILE == 'M')
-static inline void mask_irqs(void) {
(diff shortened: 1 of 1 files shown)


arm-software/embedded/llvmlibc-support/crt0/crt0-nohost.cpp
```diff
@@ -0,0 +1,41 @@
...
+}
+#elif defined(__arm__)
+static inline void mask_irqs(void) {
+  __asm__ volatile("cpsid if \n isb \n" ::: "memory");
+}
+#else
+#error "Unsupported architecture"
+#endif
+
+static inline void idle_once(void) { __asm__ volatile("wfi" ::: "memory"); }
```
smithp35: There is an intrinsic in the ACLE for this https://arm-software.github.io/acle/main/acle.html#hints 

I think `wfi` is also not supported in v4t, I think it needs 5te.

arm-software/embedded/llvmlibc-support/crt0/crt0-nohost.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+// Go into low power infinite loop on exit
+[[gnu::weak, noreturn]] void __llvm_libc_exit(int status) {
+  (void)status;
+  mask_irqs();
```
smithp35: This will need a privileged instruction and we may (likely) be running at user-mode or EL0 so this will generate an abort.

smithp35 (approved):
LGTM. Thanks for the update.

PR #685, opened 2026-01-19
[ATfL] Fix for a problem with the tests unable to find libc++.so.1

pawosm-arm:
This should fix the problems with the `unittests/CAS/CASTests` and `unittests/Support/SupportTests` test cases being unable to find libc++.so.1.

### DIFF
--- a/arm-software/linux/build.sh
+++ b/arm-software/linux/build.sh
@@ -272,6 +272,11 @@ apply_patches() {
     echo_bold "Applying patches...done"
 }
 
+bootstrap_compiler_default_config() {
+    echo "-fuse-ld=lld" >${BUILD_DIR}/bootstrap_compiler/bin/clang.cfg
+    echo "-fuse-ld=lld" >${BUILD_DIR}/bootstrap_compiler/bin/clang++.cfg
+}
+
 bootstrap_compiler_build() {
     mkdir -p "${BUILD_DIR}/stage/bootstrap_compiler"
     cd "${BUILD_DIR}/stage/bootstrap_compiler"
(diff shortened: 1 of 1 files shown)


arm-software/linux/build.sh
```diff
@@ -272,6 +272,11 @@ apply_patches() {
     echo_bold "Applying patches...done"
 }
 
+bootstrap_compiler_config_restore_sanity() {
```
kiranchandramohan: This is a strange name for the function compared to what it seems to be doing. Why does it restore sanity? What would have happened otherwise?

arm-software/linux/build.sh
```diff
@@ -272,6 +272,11 @@ apply_patches() {
...
+    echo "-fuse-ld=lld" >${BUILD_DIR}/bootstrap_compiler/bin/clang.cfg
+    echo "-fuse-ld=lld" >${BUILD_DIR}/bootstrap_compiler/bin/clang++.cfg
+}
```
kiranchandramohan, line 278: Are these only for building c/c++ libraries and hence the corresponding Fortran entry is missing?

arm-software/linux/build.sh
```diff
@@ -423,7 +432,9 @@ shared_lib_build() {
     rm -r "${ATFL_DIR}.libs"
     echo '-L<CFGDIR>/../runtimes/runtimes-bins/openmp/runtime/src $-Wl,--push-state $-Wl,--as-needed $-lomp $-ldl $-Wl,--pop-state' >bin/clang.cfg
     echo '-L<CFGDIR>/../runtimes/runtimes-bins/openmp/runtime/src $-Wl,--push-state $-Wl,--as-needed $-lomp $-ldl $-Wl,--pop-state' >bin/clang++.cfg
+    echo "-Wl,-rpath=${ATFL_DIR}/lib" >> ${BUILD_DIR}/bootstrap_compiler/bin/clang++.cfg
```
kiranchandramohan, line 435: Why is this only done for clang++ here?

arm-software/linux/build.sh
pawosm-arm: I had some trouble with coming up with a better name, so I'm aware it may seem awkward.

With this patch, we assume the bootstrap compiler settings can diverge (I can't rule out that they will never diverge even more with any future changes, hence I decided to extract this to a separate convenient function). There are places where we want exactly this settings to be used (hence I call them 'sane'), so this function ensures this 'sanity'.

arm-software/linux/build.sh
pawosm-arm: This is to configure the bootstrap compiler. It has no Fortran. We're using bootstrap compiler to buld the (proper) toolchain which includes the Fortran compiler.

arm-software/linux/build.sh
pawosm-arm: As suggested on the slack LLVM channel,  this sets the runtime path for the C++ programs to the place where libc++.so.1 built for this toolchain is being kept.

arm-software/linux/build.sh
pawosm-arm: Renamed to something that sounds neutral.

PR #686, opened 2026-01-19
[ATfL] Simplify the building process

pawosm-arm:
There is no point in separating static and shared lib builds. In the past it helped ensuring that the Fortran runtime library is being built both shared and static. Recently the ability to build and test both (static and shared) at one go has emerged, the building process can be simplified. Still, the static version of the OpenMP runtime library needs to be built separately, but this can be easily handled.

### DIFF
--- a/arm-software/linux/build.sh
+++ b/arm-software/linux/build.sh
@@ -55,7 +55,7 @@ STAGES=(
     "bootstrap_compiler_build"
     "libcpp_build"
     "product_build"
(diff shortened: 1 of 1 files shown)


garyjcarroll (dismissed):

pawosm-arm:
I'll merge this when I see our CI all green to make sure all previous issues were addressed.

garyjcarroll (approved):

PR #696, opened 2026-01-22
[ATfE] Changelog for 22.1.0 release - initial draft

voltur01:


arm-software/embedded/CHANGELOG.md
```diff
@@ -8,13 +8,41 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+(Fixed Virtual Platform).
+
+### Changed
+- LLVM libc added to the main ATfE package.
+- LLVM libc startup code and default linker script provided.
+- Many LLVM libc++ support limitations removed when used with LLVM libc.
+C++ console I/O can be used now, but not file I/O yet.
+- ATfE samples updated to work with LLVM libc, the separate set of LLVM libc
+samples removed.
+- LLVM compiler_rt profile library provided for each library variant.
```
vhscampos: Nit: I believe compiler-rt is spelled with a dash

voltur01: Thanks and a good point: the library itself is called `libclang_rt.profile.a` with underscore, but apparently not `compiler_rt`, while the LLVM project directory is `compiler-rt` and it is referred to in here https://github.com/arm/arm-toolchain/blob/arm-software/compiler-rt/README.txt as `Compiler-RT` even.

I agree that `compiler-rt` would be the best, let me update.

vhscampos (approved):

PR #698, opened 2026-01-23
[ATfE] Add a sample for using FVP

voltur01:
Add a new sample to show how to build and run a trivial application with an IoT FVP.

This sample downloads and uses wrapper scripts from ATfE repository, which are not provided in the binary package. It needs just a few files, so this seems to be better than requiring users to checkout the whole ATfE tree. It is also easier to see what, out of the whole repository, is actually needed.

The sample assumes required FVP is installed and gives pointers how to do so.

### DIFF
--- a/arm-software/embedded/samples/src/baremetal-semihosting-fvp/README.md
+++ b/arm-software/embedded/samples/src/baremetal-semihosting-fvp/README.md
@@ -8,9 +8,10 @@ installed, you can use the
 [`get_fvps.sh`](https://github.com/arm/arm-toolchain/blob/arm-software/arm-software/embedded/fvp/get_fvps.sh)
 script from the ATfE repository to install this and other FVPs.
 
-The sample requires `FVP_INSTALL_DIR` environment variable to be set to find
-the FVP. By default, it assumes that ATfE source tree is checked out and the
-FVPs are installed using the `get_fvps.sh` script above.
+The sample requires the `FVP_INSTALL_DIR` variable to be set in the environment
+or on the make command line, to find the FVP. By default, it assumes that the
+ATfE source tree is checked out and the FVPs are installed using the
+`get_fvps.sh` script above.
 
 The sample makes use of ATfE multilib variants
 [JSON files](https://github.com/arm/arm-toolchain/tree/arm-software/arm-software/embedded/arm-multilib/json/variants)


arm-software/embedded/samples/src/baremetal-semihosting-fvp/Makefile
```diff
@@ -0,0 +1,62 @@
...
+include ../../Makefile.conf
+
+# Needs FVPs installed, see https://github.com/arm/arm-toolchain/blob/arm-software/arm-software/embedded/fvp/get_fvps.sh
+# By default, assume the script is running in ATfE source tree with get_fvps.sh script run to install FVPs
+# in the default location.
+ifndef FVP_INSTALL_DIR
+    FVP_INSTALL_DIR=../../../fvp/install
+endif
+
+# From https://github.com/arm/arm-toolchain/blob/arm-software/arm-software/embedded/arm-multilib/json/variants/armebv6m_soft_nofp_size.json#L12
```
statham-arm: It's not futureproof to quote a Github URL that gives a branch name rather than a specific commit hash (as here, `arm-software`) _and_ a specific line number or range in the fragment id (`#L12`). You're promising that line 12 of that file will continue to contain the same thing in all future commits!

arm-software/embedded/samples/src/baremetal-semihosting-fvp/Makefile
```diff
@@ -0,0 +1,62 @@
...
+-Wl,--defsym=__stack_size=4K
+
+build: hello.elf
+
+hello.elf: *.c
+	$(BIN_PATH)/clang $(CFG_FILE) $(MICROBIT_OPTIONS) $(CRT_SEMIHOST) $(MATH_LIB) -g $(CORSTONE_FVP_MEMORY_MAP) -T $(LIBC_LD_FILE) -o hello.elf $^
+
+setup:
+	# Download helper files and scripts if missing
+	@if [ ! -f lit-exec-fvp.py ]; then \
```
statham-arm: I understand that you're aiming to allow a user to run this sample without having the whole git repo available, but would it not make sense to use the local `lit-exec-fvp.py` out of the git repo if it _is_ there? That way someone can easily test changes to that script.

arm-software/embedded/samples/src/baremetal-semihosting-fvp/Makefile
voltur01: Good point, thanks! I will replace line number with property name(s).

arm-software/embedded/samples/src/baremetal-semihosting-fvp/Makefile
voltur01: Oh, that is a great idea! Let me see how to add this.

arm-software/embedded/samples/src/baremetal-semihosting-fvp/Makefile
voltur01: Done

arm-software/embedded/samples/src/baremetal-semihosting-fvp/Makefile
voltur01: Done

arm-software/embedded/samples/src/baremetal-semihosting-fvp/README.md
```diff
@@ -0,0 +1,22 @@
...
+[Corstone IoT FVP (Fixed Virtual Platform)](https://developer.arm.com/Tools%20and%20Software/Fixed%20Virtual%20Platforms/IoT%20FVPs).
+
+This sample is supported on Linux only. It relies on Corstone-310 FVP to be
+installed, you can use the
+[`get_fvps.sh`](https://github.com/arm/arm-toolchain/blob/arm-software/arm-software/embedded/fvp/get_fvps.sh)
+script from the ATfE repository to install this and other FVPs.
+
+The sample requires `FVP_INSTALL_DIR` environment variable to be set to find
+the FVP. By default, it assumes that ATfE source tree is checked out and the
+FVPs are installed using the `get_fvps.sh` script above.
```
statham-arm:
```suggestion
The sample requires the `FVP_INSTALL_DIR` variable to be set in the environment
or on the make command line, to find the FVP. By default, it assumes that the
ATfE source tree is checked out and the FVPs are installed using the
`get_fvps.sh` script above.
```

voltur01: Agreed, updated the doc - thank you!

PR #705, opened 2026-01-30
[ATfE] Support command line options in LLVM libc

voltur01:
Add support in LLVM libc startup code to get command line options (argc/argv) from semihosting.

Provide an empty handler that can be overridden for no-host environment.

Semihosting passes the command line options as one string, thus basic logic to split it into individual arguments implemented:
- Dummy program name is provided as argv[0].
- Arguments are split by whitespace.
- Quoted text is copied as-is: "a b c " or 'a b c ' will keep all spaces. Not closed quote will run till the end of the provided line.
- Escape sequences: \\, \' , \" and \  to put \, ', " and space respectively.

### DIFF
--- a/arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
+++ b/arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
@@ -9,10 +9,11 @@
 #include "semihost.h"
 #include "platform.h"
 
-#include <stddef.h>
-#include <time.h>
-#include <stdlib.h>
 #include <ctype.h>
+#include <stddef.h>
+#include <stdlib.h>
+#include <string.h>
+#include <time.h>
 
 namespace {
 
@@ -137,10 +138,8 @@ static inline void skip_spaces(const char *&p) {
     ++p;
 }
 
-static int parse_cmdline_buf(char *buf, const char **argv, int max_argv) {
+static int parse_cmdline_buf(char *buf) {
   if (!buf)
-    return -1;
-  if (argv && max_argv <= 0)
     return -1;
 
   int argc = 0;
@@ -148,11 +147,8 @@ static int parse_cmdline_buf(char *buf, const char **argv, int max_argv) {
   char *w = buf;
   skip_spaces(p);
 
-  while (*p != '\0' && (argc < max_argv - 1 || argv == nullptr)) {
+  while (*p != '\0') {
     // Start of token
-    // If argv == nullptr count arguments only, do not change buf or argv
-    if (argv)
-      argv[argc] = w;
     argc++;
 
     char quote = '\0';
@@ -173,24 +169,32 @@ static int parse_cmdline_buf(char *buf, const char **argv, int max_argv) {
         break; // End of token
       }
 
-      if (argv)
-        *w++ = c;
+      *w++ = c;
     }
 
-    if (argv)
-      *w++ = '\0'; // Null-terminate token
+    *w++ = '\0'; // Null-terminate token
 
     skip_spaces(p);
   }
 
-  if (argv)
-    argv[argc] = nullptr;
-
   return argc;
 }
 
-// Parse the command line into argc/argv for the main function
-// Return estimated number of arguments if argv == nullptr
+static int fill_argv_from_parsed_buf(const char *buf, const char **argv,
+                                     int argc) {
+  for (int i = 0; i < argc; i++) {
+    argv[i] = buf;
+    buf += strlen(buf) + 1;
+  }
+  argv[argc] = nullptr;
+  return argc;
+}
+
+// Parse the command line into argc/argv for the main function.
+//
+// Must be called with argv == nullptr first to get the number of arguments,
+// then with allocated argv and the number of arguments from the first call
+// to fill in the argv.
 int _platform_get_argv(char *cmdline, int max_cmdline, const char **argv,
                        int max_argv) {
   if (!cmdline || max_cmdline <= 0)
@@ -198,6 +202,9 @@ int _platform_get_argv(char *cmdline, int max_cmdline, const char **argv,
 
   if (argv && max_argv <= 0)
     return -1;
+
+  if (argv)
+    return fill_argv_from_parsed_buf(cmdline, argv, max_argv - 1);
 
   struct {
     char *buf;
@@ -207,6 +214,6 @@ int _platform_get_argv(char *cmdline, int max_cmdline, const char **argv,
   if (semihosting_call(SYS_GET_CMDLINE, &get_cmdline_args) != 0)
     return -1;
 
-  return parse_cmdline_buf(cmdline, argv, max_argv);
+  return parse_cmdline_buf(cmdline);
 }
 } // extern "C"


arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
```diff
@@ -121,4 +121,95 @@ void _platform_debug_putc(int c) {
...
   __llvm_libc_stdio_write(&__llvm_libc_stderr_cookie, (const char *)&ch, 1);
 }
+
+// Provide command line options (argc/argv) for the main function
+
+// Supported features:
+// - Dummy program name is provided as argv[0].
+// - Arguments are split by whitespace.
+// - Quoted text is copied as-is: "a b c " or 'a b c ' will keep all spaces.
+//   Not closed quote will run till the end of the provided line.
```
statham-arm, line 132: This is not just a code comment, it's a specification for the syntax expected by the program you end up compiling. Someone who wants to run the program with a particular complicated argv list will need to understand these rules in order to unparse their intended argument list so that it will be parsed into the right thing. So it should live in the documentation, not just here.

Also, I think it's worth considering these rules at least a little bit, because changing them later will break existing command lines. Why _these_ rules? They don't exactly match any existing scheme:
* POSIX `sh` uses both single and double quotes, but treats them differently (e.g. `\` inside single quotes isn't special, whereas inside double quotes it is)
* Windows (really MSVC crt0) command line splitting doesn't treat `'` as a quote character at all, and treats `\\` as an escaped literal `\` only in some circumstances

arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
```diff
@@ -121,4 +121,95 @@ void _platform_debug_putc(int c) {
...
+static int is_space(char c) {
+  return (c == ' ' || c == '\t' || c == '\r' || c == '\n');
+}
+
+static int parse_cmdline_buf(char *buf, const char **argv, int max_args) {
+  if (!buf || !argv || max_args <= 0)
+    return 0;
+
+  // Provide a dummy program name
+  argv[0] = "program";
```
statham-arm: ACfE's handling of the string returned from `SYS_GET_CMDLINE` considers it to include the program name, but here you're considering it not to. So running semihosting binaries compiled by both toolchains, you'd have to pass them different command line strings.

The [specification](https://github.com/ARM-software/abi-aa/blob/main/semihosting/semihosting.rst#sys-get-cmdline-0x15) of `SYS_GET_CMDLINE` is vague about the meaning of the command line string. Nonetheless, given that there's an existing convention, wouldn't it be better to stick to it?

(Also, argv is supposed to contain a list of _writable_ strings, so it's not advisable to put a string literal directly into it.)

arm-software/embedded/llvmlibc-support/crt0/bootcode.cpp
```diff
@@ -71,7 +73,13 @@ void do_start() {
   memory::enable_cache();
   __libc_init_array();
   _platform_init();
-  exit(main(0, 0));
+
+  char cmdline[256];
+  const char *argv[8];
```
statham-arm: If we're about to call `main`, then that presumably means the rest of libc has already been set up, including the heap? So we could malloc the space for the command line, and then malloc the space for `argv` too – perhaps by walking along the command line string twice, first deciding how many argument pointers you need, then generating them on the second pass.

That would avoid at least one of these arbitrary limits (I think 7 arguments other than the command name is too few _even_ for a 256-byte limit on the command line itself – I was dealing with a benchmark program just this week which has a lot of short command line options). It also shifts large blocks of memory from the stack to the heap, which I think is preferable, since stacks are often small.

arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
```diff
@@ -121,4 +121,95 @@ void _platform_debug_putc(int c) {
...
+      *w++ = c;
+    }
+
+    *w = '\0'; // Null-terminate token
+
+    while (is_space(*p))
+      ++p;
+  }
+
+  argv[argc] = NULL;
```
statham-arm: Odd that you use `NULL` here, and `nullptr` in the next function 🙂

arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
voltur01: Good point, thanks! I will document in `llvmlibc.md` when we conclude the logic.

I thought about shell-like behavior however I had really difficult time trying to pass anything like that through QEMU - most of the special symbols are handled by the shell that is invoking QEMU before it reaches QEMU or the emulated program.

So I decided that a better reference would be existing embedded libraries:
* picolibc seems to only split by space https://github.com/picolibc/picolibc/blob/15f0b685ee1effe218528e994ab60b4cf600fa8a/picocrt/crt0.h#L141
* Arm libc, however, handles quotations, so that it is at least possible to pass an argument with a space in it.

So Arm libc was the bar for the proposed implementation.

Happy to consider if there is any other good reference or self-consistent model of what it makes sense to handle.

arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
voltur01: Hm, looks like there is a difference in behavior between ACfE and picolibc.

I would prefer not to have the dummy name, indeed, however that by default (without any args passed to QEMU) argc will always be 0, which is different from hosted behavior where the program name is always (?) available.

If this seems a better choice, happy to remove the dummy name.

arm-software/embedded/llvmlibc-support/crt0/bootcode.cpp
voltur01: Good point, let me see if we always have the heap, if yes, then it may be a better strategy, agreed.

VladiKrapp-Arm: Same point, but for helper functions like is_space - could we use the llvmlibc ones at this point? And maybe strtok?

I agree with using malloc for the arguments array - we've been bitten by a fixed length buffer used for this exact purpose. 
We should probably consider error handling - exit with a nice error message if the semihosting or malloc fails? Otherwise the program runs not as the user intended...

arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
voltur01: I documented the behavior which is a simple three statements: args are split by space, if you need a space use quotations, if you need a quotation mark escape it.

arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
voltur01: I removed the dummy program name.

arm-software/embedded/llvmlibc-support/crt0/bootcode.cpp
voltur01: I added use of malloc() for memory. Unfortunately, I cannot find any API to get the actual size of the command line from semihosting, so kept one hardcoded buffer size.

I replaced local is_space() with C standard one, however strtok() seems to be less compelling because of the need to handle quotations and escape sequences.

arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
voltur01: Fixed.

arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
statham-arm: I think if we're _going_ to support two quote characters, they ought to do something different, to justify having both of them at all. Each one imposes inconvenience on the user, by requiring them to quote it if they want to use it literally, so each one ought to provide some benefit in return for that inconvenience.

How about taking _one_ step further towards POSIX sh syntax, by saying that `\` is not special inside single quotes, although it is inside double quotes? That means you can use single quotes for a situation where you have a lot of backslashes and don't want to have to escape them all:
```
myprogram "c:\\foo\\bar\\baz\\quux"      # ugh, what a pain
myprogram 'c:\foo\bar\baz\quux'          # ah, much nicer
```

This means that if you want a literal single quote inside your single-quoted word it's a bit awkward: you have to close the single quotes, then write a `\'`, then reopen them:
```
myprogram 'here'\''s a single quote'
```
But POSIX sh users know this already and it will be familiar to them. And Windows users will probably prefer to use `"` in any case, and then `'` inside that doesn't need escaping at all.

arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
```diff
@@ -125,29 +126,47 @@ void _platform_debug_putc(int c) {
...
+    while (*p != '\0' && isspace(*p))
+      ++p;
+
+    if (*p == '\0')
+      break;
+
+    // Found start of an argument
+    ++argc;
+
+    // Skip non-whitespace until next whitespace separator
```
statham-arm: Surely this will get the wrong answer, because it's not paying attention to whether the whitespace is in quotes? I don't really understand why this new piece of code is _separate_ from `parse_cmdline_buf` at all, let alone separate _and completely different_.

Wouldn't it be better to use the same code for both passes, with writing of output conditionalized? Then you could trust them to agree with each other.

voltur01: Yes, this function overestimates by using very simple logic that does not track quotes.

The problem with `parse_cmdline_buf` is that it modifies `cmdline` in place to create null-terminated arguments, thus if run second time over the same `cmdline` buffer it would stop after the first, now null-terminated, argument.

So a few trade-off here between how complex the parsing logic is vs using 2x memory for the buffer to avoid updating in pace vs using a bit more memory for `argv`. I guess, most of the time quotes will not be used, e.g. in testing.

voltur01: Let me give a try at implementing single quotes, otherwise stick with only one type of quotes.

statham-arm: Yes, by "writing of output conditionalized" I meant conditionalizing _both_ writing to the `argv` array and overwriting the output string.

Although another possibility is that pass 1 processes the output string in place to leave the arguments dequoted and packed together at the start of the string separated by `'\0` (which requires no allocation because it guarantees to use at most as much space as the original string did), and it returns the total number of arguments it found. Then pass 2 just has to find that many NUL-terminated strings and write pointers to them into the argv array. Then the two passes split the work between them, without having to duplicate the same analysis as each other.

arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
voltur01: Thank you for the idea - I removed the estimation function and simplified the escaping logic: now single quotes do not treat `\` as a special escape symbol and in general they support escaping usual shell symbols like $ and ?, however this is not very useful since semihosting or embedded targets are not supposed to interpret them anyway.

arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
```diff
@@ -121,4 +122,93 @@ void _platform_debug_putc(int c) {
...
+}
+
+static int parse_cmdline_buf(char *buf, const char **argv, int max_argv) {
+  if (!buf)
+    return -1;
+  if (argv && max_argv <= 0)
+    return -1;
+
+  int argc = 0;
+  char *p = buf;
```
statham-arm: Style suggestion:

```suggestion
  const char *p = buf;
  char *w = buf;
```

If you make `p` a `const char *`, then the compiler would catch accidental writes through what's supposed to be the read pointer. That means you can't initialize `w` straight from `p` in the loop handling each argument, so instead you could initialize it to the start of `buf` here.

(You would also have to increment `w` when writing the trailing NUL on line 181; assign `argv[argc] = w` rather than `= p`; and `skip_spaces` would now take a `const char *&`.)

This way, you end up with the arguments packed at the start of the buffer, instead of having dead space between them. But that makes no difference to what the strings _are_.

(This is also the basis of my suggestion yesterday about not even doing the parsing twice. If you did this _and_ didn't pass `argv` to this function at all, then you could have a much simpler pass 2 that just looks for `'\0'` and sets the argv pointers.)

arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
```diff
@@ -121,4 +122,93 @@ void _platform_debug_putc(int c) {
...
+
+    skip_spaces(p);
+  }
+
+  if (argv) {
+    argv[argc] = nullptr;
+    return argc;
+  }
+
+  return argc + 1; // Extra slot for the terminating nullptr in argv
```
statham-arm: Style suggestion: changing the semantics of the return value depending on which pass you're in is confusing! It even confused _me_ when I pasted this function into a quick standalone test program.

Perhaps it would be better to just `return argc` unconditionally, and move the `+1` to where the malloc happens.

statham-arm (approved):
LGTM. I've made two further suggestions, but I don't think they affect correctness, so I'll leave it up to you to decide whether you like them.

arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
voltur01: Indeed, done!

PR #706, opened 2026-01-30
[ATfE] Provide LLVM libc handlers for hardware setup

voltur01:
Add handlers in LLVM libc startup code that can be overridden by the user to customize hardware setup, e.g. provide a custom exceptions table or such.

### DIFF
--- a/arm-software/embedded/llvmlibc-support/crt0/memory.cpp
+++ b/arm-software/embedded/llvmlibc-support/crt0/memory.cpp
@@ -1,5 +1,5 @@
 //
-// Copyright (c) 2025, Arm Limited and affiliates.
+// Copyright (c) 2026, Arm Limited and affiliates.
 //
 // Part of the Arm Toolchain project, under the Apache License v2.0 with LLVM
 // Exceptions. See https://llvm.org/LICENSE.txt for license information.


arm-software/embedded/llvmlibc-support/crt0/bootcode.cpp
```diff
@@ -60,15 +63,14 @@ namespace {
 [[gnu::target("branch-protection=none")]]
 #endif
 void do_start() {
-  exceptions::setup();
-  memory::setup();
-  misc::setup();
+  _platform_setup_exceptions();
+  _platform_setup_memory();
+  _platform_setup_arch_extensions();
```
smithp35, line 37: I recommend extracting the `memcpy` and `memset` below the `// Perform the equivalent of scatterloading` into a separate function. 

That would allow someone to use a different linker script with different RW and .bss.

Perhaps:
// Perform linkerscript specific tasks such as copying RW data from Flash to RAM, and zero initializing .bss. The default implementation assumes one RW data region and one .bss region, described by the following linker defined symbols:
* __data_source ; the load address of the start of the RW in Flash.
* __data_start ; the destination address of the start of the RW in RAM.
* __data_size ; the size of the RW.
* __bss_start ; the address of the start of .bss in RAM.
* __bss_size ; the size of the .bss.

extern "C" [[gnu::weak]] void _platform_linkerscript_init();
```

When llvm-libc supports TLS the function would be a good place to setup the thread pointer for the "main" thread.

Comment (voltur01) [reply arm-software/embedded/llvmlibc-support/crt0/bootcode.cpp:37]: This is a good suggestion to address this here as well, thanks - let me add one.

Comment (voltur01) [reply arm-software/embedded/llvmlibc-support/crt0/bootcode.cpp:37]: Added a dedicated handler.

Comment (voltur01) [reply arm-software/embedded/llvmlibc-support/crt0/bootcode.cpp:37]: I split the code into multiple cpp files, a lot of changes, but mostly moving things around.

File: arm-software/embedded/llvmlibc-support/crt0/memory.cpp [hunk 2]
```diff
@@ -0,0 +1,19 @@
+//
+// Copyright (c) 2025, Arm Limited and affiliates.
```
smithp35: 2026

voltur01: Good spot - fixed!

smithp35 (approved):
LGTM thanks.

PR #715, opened 2026-02-03
[ATfL][docs] Bump ATfL version in the porting document and make it easier to carry across versions

pawosm-arm:
WARNING: This commit is for the arm-software branch which follows the
LLVM's main branch. The version typed into the PortingFromACfL.md
document matches with the current LLVM version as specified in the
cmake/Modules/LLVMVersion.cmake file. When cherry-picking to the
ATfL's releae branch, this version needs to be amended to match with
the current release candidate version.

### DIFF
--- a/arm-software/linux/docs/PortingFromACfL.md
+++ b/arm-software/linux/docs/PortingFromACfL.md
@@ -16,7 +16,7 @@ ATfL supports quadruple precision real and complex types. ACfL does not.
 |Compiler                      |Version|
 |------------------------------|-------|
 |Arm Compiler for Linux (ACfL) |24.10  |
-|Arm Toolchain for Linux (ATfL)|21.1   |
+|Arm Toolchain for Linux (ATfL)|23.0.0 |
 
 ## ArmPL integration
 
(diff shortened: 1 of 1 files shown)


arm-software/linux/docs/PortingFromACfL.md
```diff
@@ -89,16 +89,16 @@ gfortran and gives their equivalent in the Arm Fortran compiler:
...
+|ACfL                  |ATfL                  |Value                     |Description                                               |
+|----------------------|----------------------|--------------------------|----------------------------------------------------------|
+|`__FLANG`             |`__flang__`           |`1`                       |Selection of compiler dependent source at compile time    |
+|`__arch64`            |`__arch64`            |`1`                       |Selection of architecture dependent source at compile time|
+|`__aarch64__`         |`__arch64__`          |`1`                       |Selection of architecture dependent source at compile time|
+|`__ARM_ARCH`          |N/A                   |`8`                       |Selection of architecture dependent source at compile time|
+|`__ARM_ARCH__`        |N/A                   |`8`                       |Selection of architecture dependent source at compile time|
+|`__armflang_major__`  |`__flang_major__`     |major part                |Underlying LLVM version details                           |
```
kiranchandramohan: Are the value major part, minor part, version/patch level part defined anywhere?

kiranchandramohan:
There is also change in directives support, I believe support for ivdep, prefetch, vector vectorlength were added.

arm-software/linux/docs/PortingFromACfL.md
pawosm-arm: Yes they are, in this cmake file: https://github.com/llvm/llvm-project/blob/main/cmake/Modules/LLVMVersion.cmake

yet we can't turn this markdown file into a CMake template (from which CMake would generate a markdown file with version values filled in): our documentation pipeline is pulling these raw files directly from this repo, if we turn it into a CMake template, they will all choke on this...

kiranchandramohan: Could you explain what happens here? Does the documentation flow convert these into real numbers?

If so, should we do the same for Arm Toolchain for Linux (ATfL)|22.x ?

pawosm-arm: > Could you explain what happens here? Does the documentation flow convert these into real numbers?
> 

Cost cutting happens here. Nothing gets substituted, the text remains as is, I just wanted to avoid the situation when we need to amend (create a new PR) each time we do a release or a release candidate.

> If so, should we do the same for Arm Toolchain for Linux (ATfL)|22.x ?

The plan was, to cherry-pick it to the 22.x release branch as soon as this lands.

kiranchandramohan: It looks odd to have names whose values are in CMakeFiles in a documentation page.

pawosm-arm: > It looks odd to have names whose values are in CMakeFiles in a documentation page.

I don't understand :(

pawosm-arm: I guess this was too radical:

> Cost cutting happens here.

To make this patch easier to land, I'm giving up on this, and restoring the old way, with the places which need to be manually edited on each version change (and on a cherry-pick).

pawosm-arm: > I don't understand :(

I'm reading my response above (from last week) and now I can see I'd also misunderstood the earlier question.

> Are the value major part, minor part, version/patch level part defined anywhere?

I guess you refer to `major part` and so on. No, these are not defined, it is a literal text I've typed into the place where the version number would have to be updated on each version change. Somehow I had thought your question was about where we get the values of current version and thus I've answered wrongly.

PR #716, opened 2026-02-05
Update NewLib to version 4.6.0

MarkMurrayARM:
The new release of Newlib is out, so ship with it.

### DIFF
--- a/arm-software/embedded/patches/newlib/0001-Enable-newlib-build.patch
+++ b/arm-software/embedded/patches/newlib/0001-Enable-newlib-build.patch
@@ -1,6 +1,18 @@
-Update patch for newlib-4.6.0 tag
+From 0e6f71752d4f5c84be7562a567d80a5be38c8b75 Mon Sep 17 00:00:00 2001
+From: Mark Murray <mark.murray@arm.com>
+Date: Tue, 10 Feb 2026 14:35:01 +0000
+Subject: Update patch for newlib-4.6.0 tag
 
-Many of the original patches have been upstreamed.
+Many of the patches for the 4.5.0 release have been merged.
+---
+ libgloss/Makefile.in               | 2 +-
+ libgloss/aarch64/syscalls.c        | 6 ++++++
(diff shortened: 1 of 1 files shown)


arm-software/embedded/versions.json
```diff
@@ -12,9 +12,9 @@
       "tag": "main"
     },
     "newlib": {
-      "url": "https://github.com/RTEMS/sourceware-mirror-newlib-cygwin.git",
-      "tagType": "commithash",
-      "tag": "5e5e51f1dc56a99eb4648c28e00d73b6ea44a8b0"
+      "url": "https://github.com/bminor/newlib.git",
```
dcandler: This repo no longer exists, we can't switch back to it.

arm-software/embedded/versions.json
```diff
@@ -12,9 +12,9 @@
...
+      "tagType": "tag",
+      "tag": "newlib-4.6.0"
```
dcandler: We discussed in https://github.com/arm/arm-toolchain/pull/688 that it would be better to fix to a commit hash rather than a tag, since the new mirror is unofficial and a hash ensures the build gets the code we're expecting.

arm-software/embedded/patches/newlib/0001-Enable-newlib-build.patch
```diff
@@ -19,11 +19,14 @@ Change-Id: Iec24f4305ad35a3a6df1672bb14555bf81249130
  newlib/libc/sys/arm/trap.S         |   2 +-
  12 files changed, 75 insertions(+), 65 deletions(-)
 
+Now stripped down for newlib-4.6.0 tag. Many of the original patches
+have been upstreamed.
```
dcandler: Nit: The commit message usually goes before the diff stats, not after. Additionally the subject line still refers to version 4.5.0 which is probably going to be confusing.

I'd recommend formatting these patches through git (i.e. `git format-patch -1 -k`) since that's usually the easiest way to keep the formatting consistent.

PR #718, opened 2026-02-06
[ATfE] Extend FVP sample to run AArch64 FVP

voltur01:
The sample Makefile extended to be able to run AArch64 FVP out of box as many users do target A-profile.

### DIFF
--- a/arm-software/embedded/samples/src/baremetal-semihosting-fvp/Makefile
+++ b/arm-software/embedded/samples/src/baremetal-semihosting-fvp/Makefile
@@ -29,6 +29,13 @@ else
     FVP_CONFIG_NAME:=cortex-m85.cfg
 endif
 
+IN_TREE_FVP_CONFIG_NAME_AARCH64:=../../../fvp/config/v8a-aarch64.cfg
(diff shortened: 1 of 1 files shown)


arm-software/embedded/samples/src/baremetal-semihosting-fvp/Makefile
```diff
@@ -85,12 +95,37 @@ run: setup hello.elf
 		--fvp-config=cortex-m85 \
 		hello.elf
 
+# AArch64 build and run command
+
+AARCH64_FVP_MEMORY_MAP=\
```
vhscampos, line 100: Just to check: do we usually define these values in the command line like this, or do we use a linker script?

voltur01: When there were only micro:bit samples we provided a fixed linker script, however as we added more targets like AArch64, then two different libraries, etc maintaining linker scripts for the product of these possible target options is a hassle - the command line alternative is used in our testing and avoids creating the linker script file.

This sample refers to our testing setting in the JSON file, so it seemed simpler to use command line for the memory map as well.

VladiKrapp-Arm:
Looks good

PR #723, opened 2026-02-11
[ATfE] Document _platform_debug_putc()

voltur01:
Add more details about `void _platform_debug_putc(int c)` to the documentation page.

### DIFF
--- a/arm-software/embedded/docs/llvmlibc.md
+++ b/arm-software/embedded/docs/llvmlibc.md
@@ -146,9 +146,10 @@ With `crt0-semihost` the output is directed to `stderr` using semihosting.
 
 With `crt0` the output is discarded. You can implement
 `void _platform_debug_putc(int c)` in your application to redirect it,
-for example, to an UART. The implementation must be safe to be called with
+for example, to a UART. The implementation must be safe to be called with
 only the assumption that the stack pointer is setup: it must not use heap or
-other LLVM libc functions.
+other LLVM libc functions. It must not access any code or data that is
+initialized by `_platform_init_data_segments()`.
 
 ## Providing command line options
 


arm-software/embedded/docs/llvmlibc.md
```diff
@@ -137,6 +137,19 @@ Example implementations are provided for:
...
+
+LLVM libc startup code uses `void _platform_debug_putc(int c)` to
+emit debug diagnostic messages, for example, inside exception handlers.
+
+With `crt0-semihost` the output is directed to `stderr` using semihosting.
+
+With `crt0` the output is discarded. You can implement
+`void _platform_debug_putc(int c)` in your application to redirect it,
+for example, to an UART. The implementation must be safe to be called with
+only the assumption that the stack pointer is setup: it must not use heap or
```
smithp35, line 150: On the assumption that this may be called before we copy code and data from flash to RAM could be worth mentioning that. Possibly:

it must not use the heap or call other LLVM libc functions. It must not access any code or data that depends on `_platform_init_data_segments()`.

arm-software/embedded/docs/llvmlibc.md
```diff
@@ -137,6 +137,19 @@ Example implementations are provided for:
...
+### Debug output
+
+LLVM libc startup code uses `void _platform_debug_putc(int c)` to
+emit debug diagnostic messages, for example, inside exception handlers.
+
+With `crt0-semihost` the output is directed to `stderr` using semihosting.
+
+With `crt0` the output is discarded. You can implement
+`void _platform_debug_putc(int c)` in your application to redirect it,
+for example, to an UART. The implementation must be safe to be called with
```
MarkMurrayARM: "UART" is pronounced with a consonant "Y' sound if you both say the letters or the word as "you-art", so that "an" should be "a" instead.

voltur01: Thank you - fixed!

arm-software/embedded/docs/llvmlibc.md
voltur01: Really good point, thanks - added!

PR #732, opened 2026-02-20
ATfL: no more static libomp after #182022

pawosm-arm:
We were using the OpenMP standalone build mode in order to build the static libomp.a runtime library which cannot be build together with the shared libomp.so (there is no FLANG_RT_ENABLE_SHARED and FLANG_RT_ENABLE_STATIC equivalent for libomp, the LIBOMP_ENABLE_SHARED flag is binary: one can get either shared or static library, but never both in the same build).

The OpenMP standalone build mode has been removed by https://github.com/llvm/llvm-project/pull/182022

This complicates our situation. Considering almost nonexistent demand for the static OpenMP runtime library, it is not worth pursuing it.

### DIFF
--- a/arm-software/linux/build.sh
+++ b/arm-software/linux/build.sh
@@ -55,7 +55,6 @@ STAGES=(
     "bootstrap_compiler_build"
     "libcpp_build"
     "product_build"
-    "static_libomp_build"
 )
 ZLIB_STATIC_PATH=${ZLIB_STATIC_PATH:-"/usr/lib/`uname -m`-linux-gnu/libz.a"}
 COMMON_LINKER_FLAGS="-Wl,--build-id"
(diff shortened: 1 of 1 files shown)


pawosm-arm:
> What will happen if a user tries either of the following options?
> 
> ```
> clang -static-openmp -fopenmp par_hello.c
> ```
> 
> or
> 
> ```
> clang -static -fopenmp par_hello.c
> ```

As expected:
```
/usr/bin/ld: cannot find -lomp: No such file or directory
```
in both cases.

pawosm-arm:
> > No use case has been removed. All you have to do is replace
> 
> Ok, I'll try it out. Thanks for the hint @Meinersbur

It worked. Closing this one, as I opened a new PR with corrected fix: #735

PR #747, opened 2026-02-25
[ATfE] Spill LR for the SVC call in LLVM libc support code's semihost

vhscampos:
In certain circumstances, the LR may be clobbered by the semihosting operation that is requested via the SVC instruction.

This patch adds LR to the clobber list of the inline assembly. This makes the compiler spill it if it is alive at that point. The LR is usually alive in leaf functions.


voltur01:
> I don't think we necessarily have to align them, since the spec does allow the broader use of HLT.

I agree that we should align only where it is pragmatic/helpful: to simplify migration and avoid confusion/surprises for users.

PR #748, opened 2026-02-25
[ATfE] Increase more timeouts

dcandler:
In addition to the armebv7a_soft big-endian variants, the armebv7a_hard variants are also simulating slow enough to sometimes hit timeout. This patch broadens the timeout case to include both.


simpal01:
LGTM

simpal01 (approved):

PR #768, opened 2026-03-03
[ATfL][docs] Updated native package manager installation

pawosm-arm:
Updated to specify installation using the arm-toolchains-transition package which configures the updated the pathname and key. Adds SLES-16 support using a new arm-toolchains-repository convenience package.


kiranchandramohan (approved):
LG.

PR #772, opened 2026-03-05
[ATfE] Fix M-profile memory setup

voltur01:
Replace unconditional enabling of caches with runtime detection and enabling only on devices that may have it based on the base architecture version. Add data cache invalidation.

Make runtime check for MPU disabling.

Known good implementation of cache management is in CMSIS https://github.com/ARM-software/CMSIS_6/blob/main/CMSIS/Core/Include/m-profile/armv7m_cachel1.h However it uses pre-computed parameters for bit manipulation.

References to relevant system registers:
* ICIALLU: https://developer.arm.com/documentation/ddi0601/2025-12/AArch32-Instructions/ICIALLU--Instruction-Cache-Invalidate-All-to-PoU?lang=en
* CLIDR: https://developer.arm.com/documentation/ddi0601/2025-12/AArch32-Registers/CLIDR--Cache-Level-ID-Register?lang=en
* CSSELR: https://developer.arm.com/documentation/ddi0601/2025-12/AArch32-Registers/CSSELR--Cache-Size-Selection-Register?lang=en
* CCSIDR: https://developer.arm.com/documentation/ddi0601/2025-12/AArch32-Registers/CCSIDR--Current-Cache-Size-ID-Register?lang=en
* DCISW: https://developer.arm.com/documentation/ddi0601/2025-12/AArch32-Instructions/DCISW--Data-Cache-line-Invalidate-by-Set-Way
* MPU_TYPE: https://developer.arm.com/documentation/ddi0419/c/System-Level-Architecture/System-Address-Map/Protected-Memory-System-Architecture--PMSAv6/MPU-Type-Register--MPU-TYPE

Tested with the FVP and by comparing generated bit patterns against simulated CMSIS run.

### DIFF
--- a/arm-software/embedded/llvmlibc-support/crt0/memory_m.h
+++ b/arm-software/embedded/llvmlibc-support/crt0/memory_m.h
@@ -36,7 +36,7 @@ using namespace sysreg;
 
 #if BOOTCODE_M_ARCH_CACHE_SUPPORTED
 static inline void sync_barriers() {
-  static constexpr unsigned int FullSystemScope = 0xf;
+  constexpr unsigned int FullSystemScope = 0xf;
 
   __dsb(FullSystemScope);
   __isb(FullSystemScope);


arm-software/embedded/llvmlibc-support/crt0/memory_m.h
```diff
@@ -19,24 +19,99 @@ namespace memory {
 
 using namespace sysreg;
 
+#if defined(__ARM_ARCH_7M__) || defined(__ARM_ARCH_7EM__) ||                   \
```
vhscampos, line 22: Where are these macros defined?

arm-software/embedded/llvmlibc-support/crt0/memory_m.h
```diff
@@ -19,24 +19,99 @@ namespace memory {
...
+#if defined(__ARM_ARCH_7EM__) || defined(__ARM_ARCH_8M_MAIN__) ||              \
+    defined(__ARM_ARCH_8_1M_MAIN__)
+#define BOOTCODE_M_ARCH_CACHE_SUPPORTED 1
+#else
+#define BOOTCODE_M_ARCH_CACHE_SUPPORTED 0
+#endif
+
+#if BOOTCODE_M_ARCH_CACHE_SUPPORTED
+static inline void sync_barriers() {
+  static constexpr unsigned int FullSystemScope = 0xf;
```
vhscampos: I don't think `static constexpr` makes any difference here. Unless I'm missing something, I suggest you change it to `const unsigned int`.

arm-software/embedded/llvmlibc-support/crt0/memory_m.h
voltur01: I expect they are generated here https://github.com/llvm/llvm-project/blob/3286c4c4f6e31a6728ff115800bec597aff0e7c9/clang/lib/Basic/Targets/ARM.cpp#L691

Example:
```
$ clang --target=arm-none-eabi -march=armv7-m -mthumb -dM -E - < /dev/null | grep ARM_ARCH
#define __ARM_ARCH 7
#define __ARM_ARCH_7M__ 1
#define __ARM_ARCH_EXT_IDIV__ 1
#define __ARM_ARCH_ISA_THUMB 2
#define __ARM_ARCH_PROFILE 'M'
```

arm-software/embedded/llvmlibc-support/crt0/memory_m.h
voltur01: `static` would not make a difference, agreed, I will remove, `constexpr` I expect to ensure this is compile time only (which it probably would be in this trivial case anyway - just to express the intent).

voltur01: Removed the `static` here.

PR #775, opened 2026-03-05
[ATfE] Update LLVM libc docs and sample for iostream support

voltur01:
Update LLVM libc to indicate that iostream is available and remove workaround for missing iostream in the C++ sample.

### DIFF
--- a/arm-software/embedded/docs/llvmlibc.md
+++ b/arm-software/embedded/docs/llvmlibc.md
@@ -205,9 +205,9 @@ $ LIBC=llvmlibc make build
 
 ## Limitations of LLVM libc in Arm Toolchain for Embedded
 
-At present, this toolchain builds C++ libraries limited to what is supported with
-LLVM libc, for example, file IO and fstream is not available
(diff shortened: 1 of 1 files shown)


arm-software/embedded/docs/llvmlibc.md
```diff
@@ -206,7 +206,8 @@ $ LIBC=llvmlibc make build
 ## Limitations of LLVM libc in Arm Toolchain for Embedded
 
 At present, this toolchain builds C++ libraries limited to what is supported with
-LLVM libc, for example, iostream is not available.
+LLVM libc, for example, file IO and fstream is not available
```
vhscampos: I suggest a tad different wording.

"At present, this toolchain builds the C++ libraries limited to what is supported in conjunction with LLVM libc. For example, file I/O and fstream are not available – only console iostream can be used.

voltur01: This is nicer - thank you! Updated.

PR #794, opened 2026-03-17
[ATfE] Enable static TLS for LLVM libc

voltur01:
Initialize static TLS in the LLVM libc startup code so TLS-backed libc state is available during language conformance testing. This does not add multithreading support.

Follow picolibc’s TLS layout and linker-script symbols to simplify migration from picolibc to LLVM libc. Also add a weak _platform_init_tls() startup handler so targets can override the default TLS initialization.

### DIFF
--- a/arm-software/embedded/llvmlibc-support/crt0/tls.cpp
+++ b/arm-software/embedded/llvmlibc-support/crt0/tls.cpp
@@ -57,6 +57,9 @@ static uintptr_t __llvm_libc_thread_pointer = 0;
 #endif
 
 [[gnu::weak]] void _platform_init_tls() {
+  // Linker defined symbols for TLS must be aligned in the linker script to
+  // meet the ABI requirements, see for an example:
+  // https://github.com/arm/arm-toolchain/blob/arm-software/arm-software/embedded/llvmlibc-support/llvmlibc.ld.in
   if (linker_value(__tls_size) == 0 || linker_value(__tls_first) == 0)
     return;
 
--- a/arm-software/embedded/llvmlibc-support/crt0/tls.cpp
+++ b/arm-software/embedded/llvmlibc-support/crt0/tls.cpp
@@ -1,0 +1,66 @@
+//
+// Copyright (c) 2026, Arm Limited and affiliates.
+//
+// Part of the Arm Toolchain project, under the Apache License v2.0 with LLVM
+// Exceptions. See https://llvm.org/LICENSE.txt for license information.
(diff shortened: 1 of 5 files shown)


vrukesh (dismissed):
LGTM

arm-software/embedded/llvmlibc-support/crt0/tls.cpp
```diff
@@ -0,0 +1,66 @@
...
+}
+#else
+[[gnu::weak]] void _set_tls(void *tls) { (void)tls; }
+#endif
+
+[[gnu::weak]] void _platform_init_tls() {
+  if (linker_value(__tls_size) == 0 || linker_value(__tls_first) == 0)
+    return;
+
+  _set_tls(__tls_first);
```
simpal01, line 66: No validation that __tls_first meets required TLS alignment? If this code assumes TLS is correctly aligned by the linker script It would be safer to explicitly document or enforce alignment  in linker script?

arm-software/embedded/llvmlibc-support/crt0/tls.cpp
```diff
@@ -0,0 +1,66 @@
...
+  const uintptr_t tp =
+      reinterpret_cast<uintptr_t>(tls) - linker_value(__arm32_tls_tcb_offset);
+  __llvm_libc_thread_pointer = tp;
+#if __ARM_ARCH_PROFILE != 'M' && __ARM_ARCH >= 6
+  __asm__ volatile("mcr p15, 0, %0, c13, c0, 3" : : "r"(tp)); // TPIDRURO
+#endif
+}
+
+[[gnu::weak]] void *__aeabi_read_tp(void) {
+  return reinterpret_cast<void *>(__llvm_libc_thread_pointer);
```
simpal01, line 53: what if __aeabi_read_tp()  called before _platform_init_tls() ? Assert is better than returning null?

voltur01: Thank you for the comments!

Using assert() literally, probably, does not make a lot of sense since we do not build debug version of the support library.

Adding a permanent check may be too big overhead to have the branch in every access.

arm-software/embedded/llvmlibc-support/crt0/tls.cpp
voltur01: The expectation is that the linker script should take care - it may be either too late or even too surprising to try to fix the pointer here, so I added a comment to clarify and linked to our linker script as an example.

simpal01 (approved):
LGTM

PR #796, opened 2026-03-18
[ATfE] Add check-llvmlibc-lit target to run LLVM libc tests with lit

voltur01:
Add check-llvmlibc-lit target to run LLVM libc tests with lit to enable initial test run and analysis before it is added to the normal check targets.

Depends on https://github.com/llvm/llvm-project/pull/187296


arm-software/embedded/arm-runtimes/CMakeLists.txt
```diff
@@ -855,8 +858,18 @@ if(C_LIBRARY STREQUAL llvmlibc)
...
         )
+        ExternalProject_Add_Step(
+            llvmlibc
+            check-libc-lit
+            COMMAND "${CMAKE_COMMAND}" --build <BINARY_DIR> --target check-libc-lit -- -k 0 || true
+            DEPENDEES install
+            USES_TERMINAL TRUE
+            EXCLUDE_FROM_MAIN TRUE
+            ALWAYS TRUE
+        )
```
vhscampos, line 869: How does this work? This step is named `check-libc-lit`, and its command runs the target with the same name. It looks cyclic to me

arm-software/embedded/arm-runtimes/CMakeLists.txt
```diff
@@ -844,6 +846,7 @@ if(C_LIBRARY STREQUAL llvmlibc)
 
     if(ENABLE_LIBC_TESTS)
         add_custom_target(check-llvmlibc)
+        add_custom_target(check-llvmlibc-lit)
```
vhscampos, line 849: A target with the same name is created in the arm-multilib CMakeLists.txt. How do they interact with each other? Does this create a name conflict?

arm-software/embedded/arm-runtimes/CMakeLists.txt
```diff
@@ -855,8 +858,18 @@ if(C_LIBRARY STREQUAL llvmlibc)
             EXCLUDE_FROM_MAIN TRUE
             ALWAYS TRUE
         )
+        ExternalProject_Add_Step(
+            llvmlibc
+            check-libc-lit
+            COMMAND "${CMAKE_COMMAND}" --build <BINARY_DIR> --target check-libc-lit -- -k 0 || true
```
vhscampos: Why does it ignore errors with `-k 0` and also always succeeds?

arm-software/embedded/CMakeLists.txt
```diff
@@ -685,6 +685,9 @@ set(LLVM_DEFAULT_EXTERNAL_LIT "${LLVM_BINARY_DIR}/bin/llvm-lit")
 
 add_custom_target(check-llvm-toolchain-runtimes)
 set(LLVM_TOOLCHAIN_LIBC_CHECK_TARGETS "")
+if(LLVM_TOOLCHAIN_ENABLE_LLVMLIBC)
+    add_custom_target(check-llvmlibc-lit)
```
vhscampos, line 689: Same question about naming conflict

arm-software/embedded/arm-runtimes/CMakeLists.txt
voltur01: Sorry for the confusion, however it is almost intentional: LLVM libc CMake now defines a new target to run tests with `lit` (vs own internal framework similar to GTest before) - this change adds ATfE top-level CMake ninja target to be able to pass through and call this LLVM libc target so that we can involve LLVM libc `lit` testing in ATfE CI.

So in short this target invokes the target with the same name, but in the inner CMake/project.

We may want to avoid this in the future by assuming that `lit` testing is the default and we do not need these special names, however before we can enable LLVM libc testing this way, we need a (special) way to invoke it locally to analyze the results and prepare xfails.

arm-software/embedded/arm-runtimes/CMakeLists.txt
voltur01: Good question, this was there for GTest runs to avoid breaking builds - I think this should be removed here, indeed. Let me see.

voltur01: Removed.

voltur01:
See https://github.com/arm/arm-toolchain/pull/833

PR #801, opened 2026-03-25
[ATfE] Increase picolibc test timeouts.

simpal01:
Armebv7a library variants can run slow enough to trigger timeouts in CI. This adjusts
the multiplier higher while the slowdown is investigated.

Timeout was applied on the 22.x branch earlier
https://github.com/arm/arm-toolchain/pull/712.


vrukesh (approved):
LGTM.

PR #827, opened 2026-04-13
[ATfE] Replace __aarch64_sme_accessible with __arm_sme_state implementation

MacDue:
This patch adds (identical) implementations of `__arm_sme_state` to newlib and picolibc. This allows us to remove the (undocumented) `__aarch64_sme_accessible` function from both newlib and picolibc.

This corresponds to the following upstream PR: https://github.com/llvm/llvm-project/pull/191434


voltur01 (dismissed):
LGTM, thank you for the update!

voltur01 (dismissed):

arm-software/embedded/patches/picolibc/0003-Add-minimal-SME-ABI-routines-__arm_sme_state.patch
```diff
@@ -0,0 +1,74 @@
...
++
++// __arm_sme_state fills the result registers based on id_aa64pfr1_el1.
++ENTRY (__arm_sme_state)
++  .variant_pcs __arm_sme_state
++  mov x0, xzr
++  mov x1, xzr
++
++  /* Although the ISA reserves BITS 24~27 for SME, at this point of time,
++     only BITS 24 (SME) and 25 (SME + SME2 ZT0 register) have meaning. */
++  mrs x16, id_aa64pfr1_el1
```
sdesmalen-arm: Does this need to be executed and cached at start-up when it's running at a higher exception level? (I'd expect that user code is still only ever executed at EL0).

MacDue: @voltur01 Do you have any thoughts on this? To me it looks like picolibc/newlib are expected to boot in EL1 (and I don't see any logic that would drop the exception level).

voltur01: The default startup code provided by both picolibc and llvmlibc does not change EL, however it is intended mostly for testing. So the previous and current implementation should be OK for using these libraries out-of-box, however for production code where a custom startup code is provided and changes the EL, this may be an issue. 

If this function is expected to be called from user code and frequently, then caching might make sense.

MacDue: This could be called frequently from user code (e.g., for streaming compatible functions), so I've updated the patch to cache "has_sme". 

Note: I've only been able to validate picolibc. I think newlib should be working too but I was unable to get the linker scripts working to test it.

voltur01: Thanks! Did you add modifications of LLVM files on purpose? I expect any LLVM changes should go into upstream LLVM Project.

voltur01:
BTW, do you plan to add this to the LLVM libc as well https://github.com/llvm/llvm-project/tree/main/libc ?

arm-software/embedded/patches/picolibc/0003-Add-minimal-SME-ABI-routines-__arm_sme_state.patch
MacDue: I'll remove the LLVM changes :+1: (forgot I'd cherry-picked patch before pushing).

MacDue:
> BTW, do you plan to add this to the LLVM libc as well https://github.com/llvm/llvm-project/tree/main/libc ?

Not at the moment (I can't see any previous work to enable SME there). It should be easy enough to port over though.

arm-software/embedded/patches/newlib/0003-Add-minimal-SME-ABI-routines-__arm_sme_state.patch
```diff
@@ -0,0 +1,125 @@
...
++
++#include "asmdefs.h"
++
++// __arm_sme_state fills the result registers based on id_aa64pfr1_el1.
++ENTRY (__arm_sme_state)
++  .variant_pcs __arm_sme_state
++  mov x0, xzr
++  mov x1, xzr
++
++  /* Check if we have SME (global initialized in crt0.c) */
```
voltur01: This refers to `crt0.c` but I cannot not see any changes in `crt0.c` in this patch.

MacDue: I've placed the `__arm64_has_sme` setup in the "Initialize SME" patches. For newlib though, this should refer to `rdimon-aem-el3.S`.

arm-software/embedded/patches/picolibc/0003-Add-minimal-SME-ABI-routines-__arm_sme_state.patch
MacDue: Btw, are you able to suggest a way to test using newlib? For picolibc, I was able to use:

`./llvm/bin/clang --target=aarch64-none-elf -fno-exceptions -fno-rtti -nostartfiles -lcrt0-semihost -lsemihost -T picolibc.ld ./test.c -march=armv8a+sme -o check_has_sme --sysroot=<picolibc sysroot>`

However, for newlib I couldn't find an equivalent linker script.

sdesmalen-arm (dismissed):
The changes look sensible to me, but please verify the patch with `newlib` before merging the PR.

arm-software/embedded/patches/picolibc/0003-Add-minimal-SME-ABI-routines-__arm_sme_state.patch
voltur01: newlib has to be built as a separate overlay package, see https://github.com/arm/arm-toolchain/blob/arm-software/arm-software/embedded/scripts/build_newlib_overlay.sh , when you build it like that and copy over an existing folder with ATfE you can use the linker script provided https://github.com/arm/arm-toolchain/blob/arm-software/arm-software/embedded/docs/newlib.md

There are also a few newlib samples here https://github.com/arm/arm-toolchain/tree/arm-software/arm-software/embedded/newlib-samples they are installed with the overlay so you may just edit one of them for your needs.

voltur01:
Just to confirm, is there a specific reason/benefit of splitting the changes into two patch files vs having one self-contained one? The dependency may not be obvious for later maintainers.

MacDue:
> Just to confirm, is there a specific reason/benefit of splitting the changes into two patch files vs having one self-contained one? The dependency may not be obvious for later maintainers.

My intent was to separate the initialization of SME (in crt0/libgloss) from the implementation of the routines in libc. It's possible more routines could depend on `__arm64_has_sme` in the future, and those routines could just be additional patches on `libc` (rather than expanding the initialization patch). 

Note: I ordered the patches so the initialization patch comes before the libc patches.

voltur01:
> My intent was to separate the initialization of SME (in crt0/libgloss) from the implementation of the routines in libc. It's possible more routines could depend on `__arm64_has_sme` in the future, and those routines could just be additional patches on `libc` (rather than expanding the initialization patch).

I see, thank you for the clarification!

voltur01 (dismissed):

arm-software/embedded/patches/picolibc/0003-Add-minimal-SME-ABI-routines-__arm_sme_state.patch
MacDue: Okay, I've tested newlib now and it seems to be working too. For future reference, building a basic newlib example required:
```
clang --target=aarch64-none-elf -nostartfiles -lrdimon -lcrt0-rdimon  <path to>/rdimon-aem-el3.o  ./test.c -march=armv8a+sme  --sysroot=<newlib sysroot> -o ./newlib -Wl,--defsym=__bss_start__=__bss_start -Wl,--defsym=__end__=end -Wl,--defsym=__bss_end__=end
```
I also had to provide empty stubs for `_init` and `_fini`:
```
void _init(void) {};
void _fini(void) {};
```

MacDue:
I've landed the LLVM side of this change (https://github.com/llvm/llvm-project/commit/b37a60793347046bb85739da3e50e2408eb5b7a9), this PR can be committed once the auto-merge picks it up.

voltur01:
Thanks, I can see it in arm-toolchain: https://github.com/arm/arm-toolchain/commit/896cd563972ebac5cb492679d35335cd5c355829

I started the missing pre-commit check and will merge afterwards.

MacDue:
Looks like there's an issue with the ARM (32) builds. I'll look into it soon (I'd only built AArch64 locally).

MacDue:
Okay, I think the issue should be resolved now.

voltur01 (approved):

voltur01:
Thank you for the update. Pre-commit checks passed, so I re-approved and merged. I will check the CI tomorrow morning.

PR #832, opened 2026-04-16
[ATfL][build] Have the clang binary bolted

pawosm-arm:
This patch tries to reuse the available stage2 mechanism for bolting the clang binary. As the currently available mechanism has its limitations, we had to introduce a lot of glue in order to make it certain that the effective clang binary is functionally the same as the one we were shipping thus far. Also we wanted it to be linked against the prebuilt libc++ binary.

Some of the points to notice:

- Currently clang with the assertions turned on cannot be bolted; subject to ongoing investigation.

- To make it easier to inject our CMake settings into the BOLT.cmake file, they had to be rearranged into several BASH associative arrays.

- The stage2 build has no install target and no proper testing target, therefore the bolted clang binary replaces the one produced in the product_build stage, and then it participates in the check-all round of tests.

- The COMMON_CMAKE_FLAGS variable should not be passed to the build.sh script anymore. A check for this has been added.

### DIFF
--- a/arm-software/linux/build.sh
+++ b/arm-software/linux/build.sh
@@ -33,6 +33,12 @@ INTERACTIVE=false
 ## Configuration: Build ##
 ##########################
 
+if [[ -n "${COMMON_CMAKE_FLAGS}" ]]
+then
+    echo "Do not use the obsolete COMMON_CMAKE_FLAGS variable."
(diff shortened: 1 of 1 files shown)


pawosm-arm:
Too many things has changed around. Closing it.

PR #834, opened 2026-04-20
[ATfE] Add temporary xfails for LLVM libc testing

dcandler:
Mark failing LLVM libc tests as expected to fail. The xfail list should be updated with more information as the causes are investigated.

Some aarch64 test timeout rather than fail, so these have been set to be excluded.

Some variants still have LLVM libc disabled as the tests still fail to build, or all the tests do build but all fail. These can be enabled later.


vrukesh (approved):
LGTM

PR #835, opened 2026-04-21
[ATfL][docs] Add SIMD and INLINEALWAYS Compiler Directives

Stylie777:
These were added in Upstream LLVM as part of https://github.com/llvm/llvm-project/pull/192969 and https://github.com/llvm/llvm-project/pull/192674 for llvm-flang. They are now recorded as existing in ATfL and ACfL, as these were previously supported by classic-flang.


Stylie777:
Currently, the patches upstream are not merged, but hope they will be landed in the next few days.

pawosm-arm (approved):
Thanks for updating it. This should be merged only after the upstream change is merged.

PR #837, opened 2026-04-21
[ATfE] Rename testsuite in LLVM libc xml results

dcandler:
Post-process any xml results from LLVM libc to rename the testsuite from the default `libc` to `libc-<variant name>`. This prevents results from different variants overwriting one another in a CI environment.

A python script already exists to do similar for the compiler-rt results, so this patch repurposes that script to work on both.

### DIFF
--- a/arm-software/embedded/arm-runtimes/test-support/update-xml-suitename.py
+++ b/arm-software/embedded/arm-runtimes/test-support/update-xml-suitename.py
@@ -1,10 +1,12 @@
 #!/usr/bin/env python3
 
-# Helper script to modify the xml results from compiler-rt.
+# Helper script to rename the testsuite in xml results.
 
-# compiler-rt always puts all the test results into the "compiler-rt"
-# testsuite in the junit xml file. We have multiple variants of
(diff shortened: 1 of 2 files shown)


arm-software/embedded/arm-runtimes/test-support/update-xml-suitename.py
```diff
@@ -13,16 +15,16 @@
...
         required=True,
-        help="Path to compiler-rt build directory",
+        help="Path to project build's test directory",
     )
     parser.add_argument(
-        "--variant",
+        "--testsuite",
         required=True,
-        help="Name of the variant under test",
+        help="New name for the testsuite",
```
HugoSilvaSantos, line 27: How about:
help="Name for the testsuite",
?

dcandler: That seems more ambiguous to me: testsuite could refer to the existing name to be replaced, or the name to replace it with. The help string should make it clear that it's the latter.

HugoSilvaSantos (approved):

PR #840, opened 2026-04-24
[ATfE] Only run library tests on Linux

dcandler:
The qemu version used on MacOS runners is a different version to Linux, which leads to libc test failures on certain unaligned variants. Until the versions can be kept consistant and/or Linux updated, library tests will only be run on Linux.

### DIFF
--- a/arm-software/embedded/scripts/test_package_only.sh
+++ b/arm-software/embedded/scripts/test_package_only.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# Copyright (c) 2025, Arm Limited and affiliates.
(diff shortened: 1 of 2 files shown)


arm-software/embedded/scripts/test_package_only.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+
+# Copyright (c) 2025, Arm Limited and affiliates.
```
voltur01:
```suggestion
# Copyright (c) 2025-2026, Arm Limited and affiliates.
```

HugoSilvaSantos (dismissed):

voltur01 (approved):

PR #843, opened 2026-04-24
[ATfE] Enable modular printf in LLVM libc

voltur01:
This enables modular printf LLVM libc implementation that for now can leave floating point support out, if it is not used, see https://discourse.llvm.org/t/rfc-printf-code-size-optimization/83146/35

This gives a clear benefit in code size: a trivial printf a float number sample is 37K of code while similar printf an integer is only 13K.

Depends on https://github.com/llvm/llvm-project/pull/194003


vhscampos (approved):
LGTM. This is an excellent addition

PR #844, opened 2026-04-27
[ATfE] Add clangd to the package

### LINKED ISSUE #838: [Feature Request] Include clangd in release binaries (Crucial for C++20 Modules developer experience)
**Problem:**
Currently, clangd (e.g., clangd.exe on Windows) is excluded from the pre-built release binaries of the arm-toolchain. While this might have been a deliberate choice to reduce distribution size, it creates a significant roadblock for developers adopting modern C++ standards—specifically C++20 Modules.

**Why this matters for C++20 Modules:**
C++20 Modules require exact version and AST alignment between the compiler generating the Built Module Interfaces (BMIs) and the language server parsing them.

Because clangd is missing, developers are forced to use a system-level clangd (via IDEs like VSCode) while compiling with the ARM-specific clang. This version mismatch almost guarantees that the system clangd cannot properly parse the BMIs generated by the ARM toolchain. Even with --query-driver workarounds, the cross-compilation module resolution is extremely brittle and often fails.

**Proposed Solution:**
Please consider restoring/adding clangd to the packaged release binaries in future versions. Since the toolchain already ships clang, including the exact matching clangd would provide a robust, out-of-the-box modern C++ experience and save developers from endless IDE configuration hell.

**Context:**
- Feature: IDE integration / C++20 Modules
- Impact: Severely degrades the modern C++ developer experience.

voltur01:
Add clangd to the package to facilitate use of C++20 modules, closes https://github.com/arm/arm-toolchain/issues/838

### DIFF
--- a/arm-software/embedded/CMakeLists.txt
+++ b/arm-software/embedded/CMakeLists.txt
@@ -281,6 +281,7 @@ set(BUG_REPORT_URL "https://github.com/arm/arm-toolchain/issues" CACHE STRING ""
 set(LLVM_DISTRIBUTION_COMPONENTS
     clang-resource-headers
     clang
+    clangd
     clang-check
     clang-format
     clang-scan-deps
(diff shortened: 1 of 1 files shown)


vhscampos (approved):
The impact on build time notwithstanding, I don't see any downside.

PR #849, opened 2026-05-08
[ATfE] Implement __arm_sme_state in LLVM libc bootcode

voltur01:
This adds LLVM libc bootcode support for __arm_sme_state as per https://github.com/ARM-software/abi-aa/blob/main/aapcs64/aapcs64.rst#811__arm_sme_state to match this LLVM change https://github.com/llvm/llvm-project/pull/191434

### DIFF
--- a/arm-software/embedded/llvmlibc-support/crt0/misc_a.h
+++ b/arm-software/embedded/llvmlibc-support/crt0/misc_a.h
@@ -18,6 +18,35 @@ namespace misc {
 
 using namespace sysreg;
 
+#if defined(__ARM_ARCH_ISA_A64)
+[[clang::always_inline]] inline bool has_sme_or_sme2() {
+  constexpr auto FEAT_SME_1 = 1UL << 24;
+  constexpr auto FEAT_SME_2 = 1UL << 25;
+
+  return (static_cast<unsigned long>(ID_AA64PFR1) & (FEAT_SME_1 | FEAT_SME_2)) != 0;
+}
+
+[[clang::always_inline]] inline void init_sme_state() {
+  if (!has_sme_or_sme2())
+    return;
+
+  // TPIDR2_EL0 resets to an architecturally unknown value, so clear it.
+  asm volatile("msr s3_3_c13_c0_5, xzr\n\tisb" : : : "memory");
+
+  // Try to set the streaming vector length to the architectural maximum.
+  unsigned long smcr;
+  if (__arm_rsr("CurrentEL") == 3 << 2) {
+    asm volatile("mrs %0, s3_6_c1_c2_6" : "=r"(smcr));
+    smcr = (smcr & ~0xfUL) | 0xfUL;
+    asm volatile("msr s3_6_c1_c2_6, %0\n\tisb" : : "r"(smcr) : "memory");
+  } else if (__arm_rsr("CurrentEL") == 2 << 2) {
+    asm volatile("mrs %0, s3_4_c1_c2_6" : "=r"(smcr));
+    smcr = (smcr & ~0xfUL) | 0xfUL;
+    asm volatile("msr s3_4_c1_c2_6, %0\n\tisb" : : "r"(smcr) : "memory");
+  }
+}
+#endif
+
 extern "C" [[gnu::weak]] void _platform_setup_arch_extensions() {
 #ifdef __ARM_FEATURE_PAUTH
   // Set all of the pointer authentication keys to different values. In
@@ -54,6 +83,7 @@ extern "C" [[gnu::weak]] void _platform_setup_arch_extensions() {
   // relevant feature these bits are ignored, so safe to set unconditionally.
   CPTR.EZ = 1;
   CPTR.ESM = 1;
+  init_sme_state();
 #else
   // Enable VFP and SIMD
   __arm_wsr("fpexc", 1 << 30);
--- a/arm-software/embedded/llvmlibc-support/crt0/misc_a.h
+++ b/arm-software/embedded/llvmlibc-support/crt0/misc_a.h
@@ -20,10 +20,7 @@ using namespace sysreg;
 
 #if defined(__ARM_ARCH_ISA_A64)
 [[clang::always_inline]] inline bool has_sme_or_sme2() {
-  constexpr auto FEAT_SME_1 = 1UL << 24;
-  constexpr auto FEAT_SME_2 = 1UL << 25;
-
-  return (static_cast<unsigned long>(ID_AA64PFR1) & (FEAT_SME_1 | FEAT_SME_2)) != 0;
+  return ID_AA64PFR1.SME != 0 || ID_AA64PFR1.SME2 != 0;
 }
 
 [[clang::always_inline]] inline void init_sme_state() {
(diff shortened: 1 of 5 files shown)


voltur01:
@MacDue @sdesmalen-arm please review __arm_sme_state for LLVM libc, thanks!

arm-software/embedded/llvmlibc-support/crt0/sme_abi.cpp
```diff
@@ -0,0 +1,37 @@
...
+    ".type __arm_sme_state, %function\n"
+    ".variant_pcs __arm_sme_state\n"
+    "__arm_sme_state:\n"
+    "  mov x0, xzr\n"
+    "  mov x1, xzr\n"
+    "\n"
+    "  /* Check whether SME or SME2 is present. */\n"
+    "  mrs x16, id_aa64pfr1_el1\n"
+    "  tst w16, #0x3000000\n"
+    "  b.eq .L__arm_sme_state_no_sme\n"
```
MacDue: Do we want to TODO/FIXME to handle this similar to picolibc/newlib? i.e., avoiding re-checking `id_aa64pfr1_el1` in user code.

arm-software/embedded/llvmlibc-support/crt0/misc_a.h
```diff
@@ -18,6 +18,35 @@ namespace misc {
...
+  // Try to set the streaming vector length to the architectural maximum.
+  unsigned long smcr;
+  if (__arm_rsr("CurrentEL") == 3 << 2) {
+    asm volatile("mrs %0, s3_6_c1_c2_6" : "=r"(smcr));
+    smcr = (smcr & ~0xfUL) | 0xfUL;
+    asm volatile("msr s3_6_c1_c2_6, %0\n\tisb" : : "r"(smcr) : "memory");
+  } else if (__arm_rsr("CurrentEL") == 2 << 2) {
+    asm volatile("mrs %0, s3_4_c1_c2_6" : "=r"(smcr));
+    smcr = (smcr & ~0xfUL) | 0xfUL;
+    asm volatile("msr s3_4_c1_c2_6, %0\n\tisb" : : "r"(smcr) : "memory");
```
MacDue: Question: Should picolib/newlib also handle the EL2 case?

MacDue (commented):
This seems reasonable to me (though I've not worked on LLVM libc)

arm-software/embedded/llvmlibc-support/crt0/misc_a.h
voltur01: picolibc assumes it is running at EL1, LLVM libc boot code here can run at EL2 and EL3.

arm-software/embedded/llvmlibc-support/crt0/sme_abi.cpp
voltur01: Added a TODO.

arm-software/embedded/llvmlibc-support/crt0/misc_a.h
```diff
@@ -18,6 +18,35 @@ namespace misc {
...
+[[clang::always_inline]] inline void init_sme_state() {
+  if (!has_sme_or_sme2())
+    return;
+
+  // TPIDR2_EL0 resets to an architecturally unknown value, so clear it.
+  asm volatile("msr s3_3_c13_c0_5, xzr\n\tisb" : : : "memory");
+
+  // Try to set the streaming vector length to the architectural maximum.
+  unsigned long smcr;
+  if (__arm_rsr("CurrentEL") == 3 << 2) {
```
vhscampos: Please add parentheses to clarify the operation precedence here. It's confusing whether the shift occurs on the value 3 or on the comparison result

arm-software/embedded/llvmlibc-support/crt0/misc_a.h
```diff
@@ -18,6 +18,35 @@ namespace misc {
...
+    asm volatile("mrs %0, s3_6_c1_c2_6" : "=r"(smcr));
+    smcr = (smcr & ~0xfUL) | 0xfUL;
+    asm volatile("msr s3_6_c1_c2_6, %0\n\tisb" : : "r"(smcr) : "memory");
+  } else if (__arm_rsr("CurrentEL") == 2 << 2) {
```
vhscampos: Same

arm-software/embedded/llvmlibc-support/crt0/misc_a.h
```diff
@@ -18,6 +18,35 @@ namespace misc {
 
 using namespace sysreg;
 
+#if defined(__ARM_ARCH_ISA_A64)
+[[clang::always_inline]] inline bool has_sme_or_sme2() {
+  constexpr auto FEAT_SME_1 = 1UL << 24;
+  constexpr auto FEAT_SME_2 = 1UL << 25;
+
+  return (static_cast<unsigned long>(ID_AA64PFR1) & (FEAT_SME_1 | FEAT_SME_2)) != 0;
+}
```
vhscampos, line 24: Can you try and add a instance of the register class for ID_AA64PFR1 and fields for Feat SME 1 and 2? Like it works for SCTLR for example. This bootcode uses object oriented paradigm to access system registers and ideally we should keep the idea.

arm-software/embedded/llvmlibc-support/crt0/misc_a.h
```diff
@@ -18,6 +18,35 @@ namespace misc {
...
+
+[[clang::always_inline]] inline void init_sme_state() {
+  if (!has_sme_or_sme2())
+    return;
+
+  // TPIDR2_EL0 resets to an architecturally unknown value, so clear it.
+  asm volatile("msr s3_3_c13_c0_5, xzr\n\tisb" : : : "memory");
```
vhscampos: Ideally do the same for `s3_3_c13_c0_5` and for the remaining register accessing parts.

arm-software/embedded/llvmlibc-support/crt0/sme_abi.cpp
```diff
@@ -0,0 +1,39 @@
+//
+// Copyright (c) 2026, Arm Limited and affiliates.
+//
+// Part of the Arm Toolchain project, under the Apache License v2.0 with LLVM
```
vhscampos, line 4: I really wonder if this should be written in C or C++, even though it would probably have a bunch of inline assembly still. It would make it easier for an eventual upstreaming to LLVM libc: they have the stance of C and C++ by default, and accepting assembly only if really necessary.

This looks doable in C.

vhscampos (changes requested):

arm-software/embedded/llvmlibc-support/crt0/sme_abi.cpp
voltur01: Good point! It is very easy to get down to assembly when all the references are in assembly.

I tried to convert everything to C++.

The tests run with this edition as well. Note that because of the system registers wrapper, code like `if (ID_AA64PFR1.SME == 0 && ID_AA64PFR1.SME2 == 0)` results in two register reads, so may be considered a bit less efficient, but much more readable.

Please let me know what you think.

vhscampos: Since this runs only once at boot, I believe it is ok.

vhscampos (approved):
It looks really good now. Thanks for the updates

voltur01:
Thank you for the review and confirmation!

@MacDue any more comments from your side?

MacDue (approved):
LGTM

PR #850, opened 2026-05-08
[ATfE] Remove __cxa_demangle from libc++abi builds

voltur01:
This removes __cxa_demangle that pulls in a lot of snprintf and other IO code into the final application binary without actually using it since the bare-metal exception handler just calls abort(). So this saves a lot of code size.


statham-arm (approved):

smithp35 (commented):
LGTM.

Arm Compiler for Embedded had a similar change, for similar reasons.

PR #858, opened 2026-05-21
[ATfE] Make use of stack size from JSON in llvmlibc.ld

voltur01:
Apply the stack size value provided in the library variant JSON file as the default stack size in the LLVM libc linker script file instead of the hardcoded 4K.


simpal01 (approved):
LGTM

PR #859, opened 2026-05-21
[ATfE] Remove absolute build directory path from libraries

voltur01:
Some library files use macros like `__FILE__` which add the absolute build directory path to the binary. To make the builds more reproducible and easier to compare (especially for FuSa purposes) this patch reuses LLVM option LLVM_USE_RELATIVE_PATHS_IN_FILES to remove the absolute paths.

LLVM options itself does not apply to libraries, thus a local implementation for it is needed.

### DIFF
--- a/arm-software/embedded/llvmlibc-support/CMakeLists.txt
+++ b/arm-software/embedded/llvmlibc-support/CMakeLists.txt
@@ -26,6 +26,17 @@
 cmake_minimum_required(VERSION 3.20.0)
 project(llvmlibc-support LANGUAGES C CXX ASM)
 
+set(LLVM_SOURCE_PREFIX "" CACHE STRING "Use prefix for sources")
+option(LLVM_USE_RELATIVE_PATHS_IN_FILES "Use relative paths in sources and debug info" OFF)
+
+if(LLVM_USE_RELATIVE_PATHS_IN_FILES)
+    file(RELATIVE_PATH relative_root "${CMAKE_BINARY_DIR}" "${CMAKE_CURRENT_SOURCE_DIR}")
+    add_compile_options(
+        "$<$<COMPILE_LANGUAGE:C,CXX,ASM>:-ffile-prefix-map=${CMAKE_BINARY_DIR}=${relative_root}>"
+        "$<$<COMPILE_LANGUAGE:C,CXX,ASM>:-ffile-prefix-map=${CMAKE_CURRENT_SOURCE_DIR}/=${LLVM_SOURCE_PREFIX}>"
+    )
+endif()
+
 option(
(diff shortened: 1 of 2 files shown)


statham-arm (commented):
I think a reasonable starting question is, are we sure we want to treat `__FILE__` and debug info the same?

Definitely it's a good idea to normalise `__FILE__`. Those strings appear in the shipped product, which doesn't really want to mention irrelevant details of the pathnames on our build setup. And they're loaded into RAM, which we'd rather save. And they have no direct link back to the original filesystem, because in a release build, that's not even on the same machine.

But full pathnames in the debug info _are_ useful, because debug builds are used by developers, who run the thing on the same machine they built it, and want their debugger to be able to find the source files. Normalising those out makes the debugging experience worse. Not _impossible_ – you can work around the problem by giving gdb some `directory` commands – but I'm not sure why we _want_ to carefully remove this information just so that developers are forced to carefully put it back again!

When I looked at these options in a different context, I ended up deciding that `-fmacro-prefix-map` was the better option to use, because it removes irrelevant stuff from `__FILE__` but keeps debugging easy.

voltur01:
Good point, originally I tried to use LLVM_USE_RELATIVE_PATHS_IN_FILES and thus modeled the solution on it as well. Now I updated to change macro paths only. Please have a look.

arm-software/embedded/llvmlibc-support/CMakeLists.txt
```diff
@@ -26,6 +26,17 @@
 cmake_minimum_required(VERSION 3.20.0)
 project(llvmlibc-support LANGUAGES C CXX ASM)
 
+set(LLVM_SOURCE_PREFIX "" CACHE STRING "Use prefix for sources")
```
statham-arm: What's `LLVM_SOURCE_PREFIX` used for? You seem to have defined it here, used it below, but never set it in the parent CMakeLists. Is it for manual use? If so, perhaps a comment would be helpful to explain why someone other than the main build scripts might want to set it.

voltur01: This was meant for consistency with the upstream LLVM_USE_RELATIVE_PATHS_IN_FILES, however I do not have use cases for this option rather than theoretically someone building llvmlibc-support locally may want to set it.

Now that I changed the option anyway, maybe it is better to remove, than to comment?

statham-arm: I don't have a strong opinion – up to you if you think anyone's going to need it!

voltur01: I removed the option to keep it simpler.

statham-arm (approved):

PR #860, opened 2026-05-26
[ATfL] Retire the ability to pass COMMON_CMAKE_FLAGS from the outside

pawosm-arm:
With this patch, an attempt to pass `COMMON_CMAKE_FLAGS` from the outside (through the environment variables) will be greeted with an error message.

### DIFF
--- a/arm-software/linux/build.sh
+++ b/arm-software/linux/build.sh
@@ -33,6 +33,11 @@ INTERACTIVE=false
 ## Configuration: Build ##
 ##########################
 
+if [[ -n "${COMMON_CMAKE_FLAGS}" ]]; then
+    echo "Do not pass the obsolete/undocumented/misleading COMMON_CMAKE_FLAGS variable."
+    exit 1
+fi
+
 RELEASE_FLAGS=${RELEASE_FLAGS:-"false"}
 LLVM_VERSION_MAJOR=$(cat ${SOURCES_DIR}/cmake/Modules/LLVMVersion.cmake | grep -i set | grep LLVM_VERSION_MAJOR | grep -o '[0-9]\+')
 LLVM_VERSION_MINOR=$(cat ${SOURCES_DIR}/cmake/Modules/LLVMVersion.cmake | grep -i set | grep LLVM_VERSION_MINOR | grep -o '[0-9]\+')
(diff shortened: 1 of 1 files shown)


arm-software/linux/build.sh
```diff
@@ -33,6 +33,11 @@ INTERACTIVE=false
 ## Configuration: Build ##
 ##########################
 
+if [[ -n "${COMMON_CMAKE_FLAGS}" ]]; then
```
amilendra, line 36: The commit message says "gradually retire". So shouldn't we first emit a message of deprecation, continue without exiting and then completely remove it in a future release?

pawosm-arm: We never recommended to use it in the first place, and I don't remember what circumstances allowed this feature to appear here. The 'gradual retire' effectively means we're looking if something is going to fail (showing the error message) after we introduce this commit.

pawosm-arm: Looking at the log:

`This will allow us to tweak the CMake configuration.`

I can't remember which part of the CMake configuration we wanted to tweak using this hammer, I guess we were facing some problem, and this was one of the measures to address it. I have no idea if this has found any use in our CI, I hope it hasn't.

amilendra: Right, so in effect, we are going to revert #122. I see that #130 mentions overriding `COMMON_CMAKE_FLAGS `to disable the use of `ZSTD`. Won't that break? If it does not break CI we will keep this patch, and then it is not a "gradual retirement" and this patch will be final. So I think removing "Gradually" would make it easy to identify at a later date which commit removed support for overriding this flag.

And just out of interest, because you say our CI, you don't expect external users to have their custom workflows that override `COMMON_CMAKE_FLAGS`?

pawosm-arm: The part of the code modified by #130 is no longer in there. The struggle with zstd has been moving sideways, I doubt it was the main reason for introducing the changes in #122. The more I look into it, the more I regret I've introduced PR #122.

pawosm-arm: > And just out of interest, because you say our CI, you don't expect external users to have their custom workflows that override `COMMON_CMAKE_FLAGS`?

I don't think this is the right way for customizing workflows. It's far too easy to end up with something that claims to be ATfL but it isn't anymore.

amilendra (approved):
Thanks. LGTM.

arm-software/linux/build.sh
pawosm-arm: > So I think removing "Gradually" would make it easy to identify at a later date which commit removed support for overriding this flag.

I've reworded the commit message, it's not "Gradually" anymore.

PR #862, opened 2026-05-29
[ATfL][build] Copying the OpenMP Fortran modules and headers should be more adaptive

pawosm-arm:
For the sake of user convenience, we're copying the OpenMP Fortran modules and header files into an easily accessible `include` directory. The most recent upstream changes have spread those files across directories and changed the paths. As the change appears back and forth, we need to be more adaptive here.

This is to address the changes introduced by this upstream PR: https://github.com/llvm/llvm-project/pull/171515

### DIFF
--- a/arm-software/linux/build.sh
+++ b/arm-software/linux/build.sh
@@ -459,7 +459,22 @@ package() {
       cp "${LIBRARIES_DIR}/libamath.so" \
           "${ATFL_DIR}/lib/${ATFL_TARGET_TRIPLE}"
     fi
-    cp ${ATFL_DIR}/include/flang/omp* "${ATFL_DIR}/include"
+    if compgen -G "${ATFL_DIR}/include/flang/omp*" >/dev/null; then
+      # Handle the old directory layout
+      cp "${ATFL_DIR}"/include/flang/omp* "${ATFL_DIR}/include"
+    else
+      # Handle the new directory layout
+      cp "${ATFL_DIR}"/lib/clang/*/include/omp* "${ATFL_DIR}/include"
+      cp "${ATFL_DIR}"/lib/clang/*/finclude/flang/"${ATFL_TARGET_TRIPLE}"/omp* "${ATFL_DIR}/include"
+    fi
+    if ! compgen -G "${ATFL_DIR}/include/omp*.h" >/dev/null; then
+      echo "The OpenMP headers could not be copied to ${ATFL_DIR}/include"
+      exit 1
+    fi
+    if ! compgen -G "${ATFL_DIR}/include/omp*.mod" >/dev/null; then
+      echo "The OpenMP Fortran modules could not be copied to ${ATFL_DIR}/include"
+      exit 1
+    fi
     cp "${ATFL_DIR}/share/man/man1/clang.1" "${ATFL_DIR}/share/man/man1/armclang.1"
     sed -i "s/clang /armclang /g" "${ATFL_DIR}/share/man/man1/armclang.1"
     sed -i "s/Bclang/Barmclang/g" "${ATFL_DIR}/share/man/man1/armclang.1"


arm-software/linux/build.sh
```diff
@@ -459,7 +459,15 @@ package() {
...
-    cp ${ATFL_DIR}/include/flang/omp* "${ATFL_DIR}/include"
+    if compgen -G "${ATFL_DIR}/include/flang/omp*" >/dev/null; then
+      cp "${ATFL_DIR}"/include/flang/omp* "${ATFL_DIR}/include"
+    fi
+    if compgen -G "${ATFL_DIR}/lib/clang/*/include/omp*" >/dev/null; then
+      cp "${ATFL_DIR}"/lib/clang/*/include/omp* "${ATFL_DIR}/include"
+    fi
+    if compgen -G "${ATFL_DIR}/lib/clang/*/finclude/flang/${ATFL_TARGET_TRIPLE}/omp*" >/dev/null; then
+      cp "${ATFL_DIR}"/lib/clang/*/finclude/flang/"${ATFL_TARGET_TRIPLE}"/omp* "${ATFL_DIR}/include"
+    fi
```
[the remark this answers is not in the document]
pawosm-arm: IMO the solution I've just uploaded is more robust.

amilendra (commented):
Can you please link the recent upstream changes that made these changes necessary? That should help in deciding if the cases handled by this patch covers everything.

pawosm-arm:
> Can you please link the recent upstream changes that made these changes necessary? That should help in deciding if the cases handled by this patch covers everything.

@amilendra certainly, I've added the link to this PR's commit message.

arm-software/linux/build.sh
```diff
@@ -459,7 +459,23 @@ package() {
       cp "${LIBRARIES_DIR}/libamath.so" \
           "${ATFL_DIR}/lib/${ATFL_TARGET_TRIPLE}"
     fi
-    cp ${ATFL_DIR}/include/flang/omp* "${ATFL_DIR}/include"
+    if compgen -G "${ATFL_DIR}/include/flang/omp*" >/dev/null; then
```
amilendra, line 462: Would something like this be easier to extend?

```
  shopt -s nullglob

  omp_headers=(
    "${ATFL_DIR}/include/flang/omp"*
    "${ATFL_DIR}/lib/clang/"*/include/omp*
    "${ATFL_DIR}/lib/clang/"*/finclude/flang/"${ATFL_TARGET_TRIPLE}"/omp*
  )

  if ((${#omp_headers[@]})); then
    cp "${omp_headers[@]}" "${ATFL_DIR}/include"
  fi
```

arm-software/linux/build.sh
```diff
@@ -459,7 +459,23 @@ package() {
...
+      cp "${ATFL_DIR}"/include/flang/omp* "${ATFL_DIR}/include"
+    fi
+    if compgen -G "${ATFL_DIR}/lib/clang/*/include/omp*" >/dev/null; then
+      cp "${ATFL_DIR}"/lib/clang/*/include/omp* "${ATFL_DIR}/include"
+    fi
+    if compgen -G "${ATFL_DIR}/lib/clang/*/finclude/flang/${ATFL_TARGET_TRIPLE}/omp*" >/dev/null; then
+      cp "${ATFL_DIR}"/lib/clang/*/finclude/flang/"${ATFL_TARGET_TRIPLE}"/omp* "${ATFL_DIR}/include"
+    fi
+    if ! compgen -G "${ATFL_DIR}/include/omp*.h" >/dev/null; then
```
amilendra, line 470: This too seems to be using a repeatable pattern that can be refactored.

arm-software/linux/build.sh
pawosm-arm: You're overdoing it IMO, and putting all those paths into one `omp_headers` container leads a prospective reader to misleading conclusions. The first path is the one that goes away (I'm keeping it here, as I often use the same `build.sh` script during the bisects, if anything goes wrong upstream), the other two are the new ones (as I mentioned in the commit message, the headers and the modules now were spread apart). Normally, avoiding repetitions is good, but not if such avoidance obscures the intent of the code.

amilendra: Anyone else who is looking at the block at a later date would only see a test and copy of three paths and would very likely make the same refactoring attempt. Other than that I don't think anyone can come up with any other intent. If you want to stop anyone from refactoring this, add that in a comment close to this block.

Also, if the test for the legacy `"${ATFL_DIR}/include/flang/omp"*` path succeeds, do you still have to check for the two new paths?

pawosm-arm: Good point, it should be always either the old or the new layout. I've rearranged this.

amilendra: Thanks. Now the intention is clearer.

amilendra (approved):

PR #866, opened 2026-06-02
[ATfL][build.sh] Do not use run_command unnecessarily

pawosm-arm:
Using run_command in this place does not make sense and simply does not work (it was not even defined yet!).


vrukesh (approved):
LGTM

PR #867, opened 2026-06-02
[ATfE] Build packages in subdirectory

dcandler:
Instead of creating packages in the root of the build directory, put them in a subdirectory so they can be easily distinguished from the other build files.

The scripts also now remove the CPack working directory once the package is built, so that the package directory should contain nothing but the package/packages.


arm-software/embedded/scripts/build.ps1
```diff
@@ -18,5 +18,8 @@ $buildDir = (Join-Path $repoRoot build)
 mkdir $buildDir
 cd $buildDir
 
-cmake ..\arm-software\embedded -GNinja -DFETCHCONTENT_QUIET=OFF -DLLVM_TOOLCHAIN_ENABLE_PICOLIBC=ON -DLLVM_TOOLCHAIN_ENABLE_LLVMLIBC=ON -DENABLE_QEMU_TESTING=OFF -DENABLE_FVP_TESTING=OFF -DLLVM_PARALLEL_LINK_JOBS=2 -DSHORT_BUILD_PATHS=ON
+cmake ..\arm-software\embedded -GNinja -DFETCHCONTENT_QUIET=OFF -DCPACK_PACKAGE_DIRECTORY=atfe_packages -DLLVM_TOOLCHAIN_ENABLE_PICOLIBC=ON -DLLVM_TOOLCHAIN_ENABLE_LLVMLIBC=ON -DENABLE_QEMU_TESTING=OFF -DENABLE_FVP_TESTING=OFF -DLLVM_PARALLEL_LINK_JOBS=2 -DSHORT_BUILD_PATHS=ON
 ninja -j 48 package-llvm-toolchain
+
+# Remove CPack working directory.
+Remove-Item -Recurse -Force atfe_packages/_CPack_Packages
```
statham-arm, line 25: My Powershell isn't good enough to remember the answer offhand, and my Windows VM isn't in a fit state to check it myself, but are you sure it's OK to use a `/` as the path separator here? I know some Windows programs are fine with it and some object.

statham-arm (approved):
LGTM unless you decide my one nitpick has a point 🙂

arm-software/embedded/scripts/build.ps1
dcandler: I similarly had to open PowerShell while changing this just to check I had the syntax right (PowerShell accepts `rm` as an alias of `Remove-Item`, but doesn't understand the same flags). But yes, it does accept `/` as separators, although until you pointed this out I hadn't noticed the _rest_ of the script uses `\`.

This was originally a batch file, and unlike PowerShell `cmd.exe` doesn't understand `/` as separators, which is likely why.

PR #869, opened 2026-06-03
[ATfL]: Update build.sh for comments from shellcheck

vrukesh:
Update build.sh for comments from shellcheck.


pawosm-arm (approved):

pawosm-arm:
In a prospective followup PR, the whole part creating a legacy `compiler.yaml` file needs to be removed. We have released a separate `atfl` package for spack, so now we don't recommend using `compiler.yaml`.

PR #873, opened 2026-06-03
[ATfL][build.sh] Do not generate the obsolete compilers.yaml file

pawosm-arm:
After the atfl package has landed in the spack-packages repository, we do not need to generate the compilers.yaml file anymore.


vrukesh (approved):
LGTM

PR #874, opened 2026-06-04
Reject invalid BF encoding when target is next instruction

HamzaKodez:
When the BF instruction targets the next immediate label, the encoded branch offset becomes zero. Which caused LLVM to emmit invalid machine code.

Added validation in the fixup_bf_branch path to reject this case and emit an error instead.

Added MC regression test to cover new validation.

Assisted by ChatGPT. Human-verified, debugged, tested and validating by author.

Fixes: LLVMAENG-4768


voltur01:
Hi @HamzaKodez, this should be fixed in the upstream https://github.com/llvm/llvm-project itself - please create the PR there (and close this one).

HamzaKodez:
Closing as requested. Will submit upstream to llvm/llvm-project.

PR #877, opened 2026-06-09
[ATFE] Add build controls for layered runtime payloads

simpal01:
Add CMake cache options to control which runtime payloads are built and installed for layered variants. This allows llvm-libc variants to install only the required minor-variant payload, while also making compiler-rt, llvm-libc core archives, and llvm-libc support libraries independently selectable.

These options are currently defined in CMake so the layering flow can be enabled, but they are expected to move into the variant JSON handling in a follow-up patch..


arm-software/embedded/arm-runtimes/CMakeLists.txt
```diff
@@ -24,6 +24,12 @@ get_filename_component(llvmproject_src_dir_abs "${llvmproject_src_dir}" ABSOLUTE
 set(C_LIBRARY "picolibc" CACHE STRING "Which C library to use.")
 set_property(CACHE C_LIBRARY PROPERTY STRINGS picolibc newlib newlib-nano llvmlibc)
 
+set(ENABLE_LLVM_LIBC_MINOR_VARIANT_def OFF)
+set(LLVM_LIBC_SUBARCHIVE_NAME_def "")
+set(ENABLE_LLVMLIBC_CORE_LIBS_def ON)
```
voltur01, line 29: Is there a reason to use two spellings: LLVMLIBC and LLVM_LIBC above?

simpal01: Changed to match with other variables

arm-software/embedded/arm-runtimes/CMakeLists.txt
```diff
@@ -96,9 +102,14 @@ set(LIBRARY_BUILD_TYPE ${LIBRARY_BUILD_TYPE_def} CACHE STRING "Build type to use
 set_property(CACHE LIBRARY_BUILD_TYPE PROPERTY STRINGS ${SUPPORTED_LIBRARY_BUILD_TYPES})
 
 set(ENABLE_CXX_LIBS ${ENABLE_CXX_LIBS_def} CACHE BOOL "Build CXX libs")
+set(ENABLE_COMPILER_RT_LIBS ${ENABLE_COMPILER_RT_LIBS_def} CACHE BOOL "Build compiler-rt libraries.")
 set(ENABLE_LIBC_TESTS ${ENABLE_LIBC_TESTS_def} CACHE BOOL "Enable libc tests (picolibc, newlib, newlib-nano or llvm-libc).")
 set(ENABLE_COMPILER_RT_TESTS ${ENABLE_COMPILER_RT_TESTS_def} CACHE BOOL "Enable compiler-rt tests.")
 set(ENABLE_LIBCXX_TESTS ${ENABLE_LIBCXX_TESTS_def} CACHE BOOL "Enable libcxx tests.")
+set(ENABLE_LLVMLIBC_MINOR_VARIANT ${ENABLE_LLVMLIBC_MINOR_VARIANT_def} CACHE BOOL "Install only the llvm-libc minor variant payload for this variant.")
+set(LLVMLIBC_SUBARCHIVE_NAME ${LLVMLIBC_SUBARCHIVE_NAME_def} CACHE STRING "llvm-libc subarchive to install when ENABLE_LLVMLIBC_MINOR_VARIANT is enabled.")
+set(ENABLE_LLVMLIBC_CORE_LIBS ${ENABLE_LLVMLIBC_CORE_LIBS_def} CACHE BOOL "Build and install the core llvm-libc library archives.")
```
vhscampos: What are the "core llvmlibc" libraries?

arm-software/embedded/arm-runtimes/CMakeLists.txt
```diff
@@ -96,9 +102,14 @@ set(LIBRARY_BUILD_TYPE ${LIBRARY_BUILD_TYPE_def} CACHE STRING "Build type to use
 set_property(CACHE LIBRARY_BUILD_TYPE PROPERTY STRINGS ${SUPPORTED_LIBRARY_BUILD_TYPES})
 
 set(ENABLE_CXX_LIBS ${ENABLE_CXX_LIBS_def} CACHE BOOL "Build CXX libs")
+set(ENABLE_COMPILER_RT_LIBS ${ENABLE_COMPILER_RT_LIBS_def} CACHE BOOL "Build compiler-rt libraries.")
```
vhscampos: The descriptions below mention "build and install", where this one is limited to "build". Does this make sense?

arm-software/embedded/arm-runtimes/CMakeLists.txt
```diff
@@ -152,6 +163,28 @@ else()
     set(cpu_family arm)
 endif()
 
+# Some llvm-libc variants are packaged as minor layers. The subprojects
+# are still built as normal, but the installed payload is restricted to
+# the files that differ from the base layer.
+if(ENABLE_LLVMLIBC_MINOR_VARIANT)
+    if(NOT C_LIBRARY STREQUAL llvmlibc)
+        message(FATAL_ERROR "ENABLE_LLVMLIBC_MINOR_VARIANT is only supported with C_LIBRARY=llvmlibc.")
```
vhscampos, line 174: Isn't the CMake var to enable llvm libc `LLVM_TOOLCHAIN_ENABLE_LLVMLIBC`?

arm-software/embedded/arm-runtimes/CMakeLists.txt
```diff
@@ -345,93 +378,105 @@ set(
...
+            compiler_rt-build
+            clib-install
+            compiler-rt-xfail-lit-args
+        )
+        add_dependencies(check-compiler-rt compiler_rt-check-compiler-rt)
+    endif()
+else()
+    add_custom_target(compiler_rt-configure)
+    add_custom_target(compiler_rt-build)
+    add_custom_target(compiler_rt-install)
```
vhscampos: What's the point of these custom targets?

arm-software/embedded/arm-runtimes/CMakeLists.txt
```diff
@@ -345,93 +378,105 @@ set(
...
+        COMMAND
+            ${CMAKE_COMMAND}
+            -E copy_directory
+            <INSTALL_DIR>/lib/${normalized_target_triple}
+            "${TEMP_LIB_DIR}/lib"
+    )
+else()
+    add_custom_target(compiler_rt_profile-configure)
+    add_custom_target(compiler_rt_profile-build)
+    add_custom_target(compiler_rt_profile-install)
```
vhscampos: Ditto

arm-software/embedded/arm-runtimes/CMakeLists.txt
```diff
@@ -811,86 +856,138 @@ if(C_LIBRARY STREQUAL llvmlibc)
         # Let CMake know we're cross-compiling
         -DCMAKE_SYSTEM_NAME=Generic
         -DCMAKE_TRY_COMPILE_TARGET_TYPE=STATIC_LIBRARY
+        -DLLVM_LIBC_ENABLE_MINOR_VARIANT=${ENABLE_LLVMLIBC_MINOR_VARIANT}
```
vhscampos: I assume these CMake vars will be defined in another patch?

arm-software/embedded/arm-runtimes/CMakeLists.txt
```diff
@@ -811,86 +856,138 @@ if(C_LIBRARY STREQUAL llvmlibc)
...
+            -DLLVM_ENABLE_PER_TARGET_RUNTIME_DIR=ON
+            -DLLVM_ENABLE_RUNTIMES=libc
+            -DLLVM_LIBC_ALL_HEADERS=ON
+            -DLLVM_LIBC_FULL_BUILD=ON
+            -DLIBC_TEST_COMPILE_OPTIONS_DEFAULT=${llvmlibc_test_compile_options_default}
+            -DLIBC_TEST_LINK_OPTIONS_DEFAULT=${llvmlibc_test_link_options_default}
+            -DLIBC_TEST_CMD=${test_cmd}
+            -DLIBC_TEST_HERMETIC_ONLY=ON
+            -DLLVM_EXTERNAL_LIT=${external_lit_path}
+            -DLLVM_LIT_ARGS=${llvmlibc_lit_args}
```
vhscampos: Most of this is duplicated between the if and else branches. Is it possible to have a single `ExternalProject_Add` call, but decide between the required argument valuesusing the if-else?

arm-software/embedded/arm-runtimes/CMakeLists.txt
```diff
@@ -811,86 +856,138 @@ if(C_LIBRARY STREQUAL llvmlibc)
...
+            USES_TERMINAL_CONFIGURE TRUE
+            USES_TERMINAL_BUILD TRUE
+            USES_TERMINAL_INSTALL TRUE
+            USES_TERMINAL_TEST TRUE
+            LIST_SEPARATOR ,
+            CONFIGURE_HANDLED_BY_BUILD TRUE
+        )
+    else()
+        add_custom_target(llvmlibc-support-build)
+        add_custom_target(llvmlibc-support-install)
```
vhscampos: Why?

arm-software/embedded/arm-runtimes/CMakeLists.txt
simpal01: I think ENABLE_COMPILER_RT_LIBS also enables the install step for
compiler-rt, so I agree the description should say “Build and install” for consistency.

arm-software/embedded/arm-runtimes/CMakeLists.txt
simpal01: Here, “core llvm-libc libraries” means the libraries produced by the upstream libc runtime build itself, mainly the target specific llvm-libc archives installed.

That is separate from llvmlibc-support, which is ATfE’s support layer under arm-software/embedded/llvmlibc-support. The support layer builds things like libcrt0.a, libcrt0-none.a, libcrt0-semihost.a, and libsemihost.a.

I will change the help text to avoid the vague word “core”. 
```
set(ENABLE_LLVMLIBC_CORE_LIBS ${ENABLE_LLVMLIBC_CORE_LIBS_def} CACHE BOOL "Build and install library archives from the llvm-libc build.")
```

arm-software/embedded/arm-runtimes/CMakeLists.txt
simpal01: `LLVM_TOOLCHAIN_ENABLE_LLVMLIBC` is the top-level option for enabling the llvm-libc runtime. This check is in the arm-runtimes subproject, where the selected libc is passed in as **C_LIBRARY** . arm-multilib configures it with `-DC_LIBRARY=${C_LIBRARY}` So C_LIBRARY=llvmlibc is the local condition we need here.

arm-software/embedded/arm-runtimes/CMakeLists.txt
simpal01: Those empty add_custom_target() calls are placeholders for when ENABLE_COMPILER_RT_LIBS=OFF.

When compiler-rt is disabled, those targets would not exist. But the rest of the build graph still refers to them, for example:

  - picolibc has DEPENDS compiler_rt-install
  - cxxlibs has DEPENDS compiler_rt-install
  - etc

  So the empty targets keep those dependencies valid and make them no-ops when compiler-rt is disabled rather than forcing every dependency site to special-case compiler-rt.

arm-software/embedded/arm-runtimes/CMakeLists.txt
simpal01: These are also the same placeholder pattern

arm-software/embedded/arm-runtimes/CMakeLists.txt
simpal01: Yes. 
ENABLE_LLVMLIBC_MINOR_VARIANT and LLVMLIBC_SUBARCHIVE_NAME are cache variables defined in this patch, with ENABLE_LLVMLIBC_MINOR_VARIANT  defaulting to OFF and LLVMLIBC_SUBARCHIVE_NAME as empty.

The intent is to enable this later through the corresponding JSON files, which will come in a separate patch once we agree on the naming conventions used here. As it stands, this patch should not introduce functional changes because these new CMake controls are disabled by default.

arm-software/embedded/arm-runtimes/CMakeLists.txt
simpal01: These are also the same placeholder pattern as explained above

arm-software/embedded/arm-runtimes/CMakeLists.txt
dcandler: I think it would make more sense to only set the dependencies when they're actually needed, rather than have placeholder targets that are never intended to build anything.

arm-software/embedded/arm-runtimes/CMakeLists.txt
simpal01: Yeah, I agree. I’ve moved the layering-specific build/install commands into a small argument list, and now use a single shared ExternalProject_Add call for llvmlibc.

Just to clarify, there are two separate LLVM libc flows:

1.  The normal/full libc.a build, including the subarchive build.
2. The variant builds, which only build llvmlibc-support.

llvmlibc-support is ATfE’s support layer under arm-software/embedded/llvmlibc-support. It builds artifacts such as libcrt0.a, libcrt0-none.a, libcrt0-semihost.a, and libsemihost.a.

So for this minor variant, even though we do not build libc.a, we still need the libc headers because llvm-libc-support depends on them.

arm-software/embedded/arm-runtimes/CMakeLists.txt
vhscampos: But since this CMakeLists.txt here is internal, I wonder if the error message should mention the public facing CMake variable instead. I am not sure, just bringing this up for reflection

arm-software/embedded/arm-runtimes/CMakeLists.txt
vhscampos: I agree with @dcandler

simpal01: Done

arm-software/embedded/arm-runtimes/CMakeLists.txt
simpal01: Mentioning ENABLE_LLVMLIBC_MINOR_VARIANT is correct here. This check is mainly validating the internal JSON/ library entries, and that flag should only be used for the llvmlibc entry. LLVM_TOOLCHAIN_ENABLE_LLVMLIBC is the public option, but it is not the thing being validated by this error.

PR #879, opened 2026-06-10
[ATFE] Run llvmlibc tests against generated sysroot matching final multilib layout.

simpal01:
LLVM libc tests now run against a generated build-tree sysroot so tests can be compiled and linked using the same multilib layout that will be installed in the final package.

LLVM libc tests need to resolve headers and libraries through the multilib layout, including the optimised common-header layout when enabled. So the ATFE build system passes --sysroot when compiling and linking these tests. Before this change, that sysroot pointed at the runtimes tmp_install area, which is created as a per-variant temporary install directory. That layout is useful while building each variant, but it does not work in the layered layout where minor variants contain only the payload required for their subarchive and rely on content inherited from base variants. As a result, individual per-variant temporary install directories no longer represent a complete view of the installed sysroot and are not suitable as test environments. Building a dedicated test sysroot lets the tests exercise the generated multilib package structure instead of relying only on per-variant temporary install directories.

This patch:

- Adds LLVM_LIBC_TEST_SYSROOT.
- When C_LIBRARY=llvmlibc, creates a new target: llvmlibc-test-sysroot. 
      -  That target deletes/recreates the test sysroot, copies either:
         - multilib-optimised/ if header optimisation is enabled, or
         - multilib/ otherwise. 
- For LLVM libc tests, the test compile and link options now include: - --sysroot=<LLVM_LIBC_TEST_SYSROOT>. - the variant's multilib flags
- A dependency from LLVM libc check targets to the generated test sysroot

### DIFF
--- a/arm-software/embedded/arm-multilib/CMakeLists.txt
+++ b/arm-software/embedded/arm-multilib/CMakeLists.txt
@@ -25,6 +25,7 @@ set(ENABLE_VARIANTS "all" CACHE STRING "Semicolon separated list of variants to 
 set(C_LIBRARY "picolibc" CACHE STRING "Which C library to use.")
 set_property(CACHE C_LIBRARY PROPERTY STRINGS picolibc newlib newlib-nano llvmlibc)
 set(PROJECT_PREFIX "${CMAKE_BINARY_DIR}/lib-builds" CACHE STRING "Directory to build subprojects in.")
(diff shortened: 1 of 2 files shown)


vhscampos (dismissed):
LGTM

vhscampos (approved):

PR #880, opened 2026-06-10
[ATfL]: Enable xml test report output for test suite

vrukesh:
Also, ensure to continue tests even on test failure.
At end of build and test, add a check to indicate if there is test failure, and appropriately reflect the status in CI pipelines.

### DIFF
--- a/arm-software/linux/build.sh
+++ b/arm-software/linux/build.sh
@@ -236,7 +236,7 @@ run_command() {
     "$@"
 }
 
-run_lit_ninja() {
+run_test_command() {
     local xml_output="$1"
     local log_file="$2"
     shift 2
@@ -399,7 +399,7 @@ bootstrap_compiler_build() {
     run_command cmake --install . 2>&1 | tee -a "${LOGS_DIR}/bootstrap_compiler.txt"
     export PATH="${BUILD_DIR}/bootstrap_compiler/bin:${PATH}"
     bootstrap_compiler_default_config
-    run_lit_ninja "${LOGS_DIR}/bootstrap_check_all.xml" "${LOGS_DIR}/bootstrap_compiler.txt" check-all
+    run_test_command "${LOGS_DIR}/bootstrap_check_all.xml" "${LOGS_DIR}/bootstrap_compiler.txt" check-all
 }
 
 libcpp_build() {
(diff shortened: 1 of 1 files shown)


arm-software/linux/build.sh
```diff
@@ -236,6 +197,47 @@ run_command() {
...
+    shopt -s nullglob
+    xml_files=("${LOGS_DIR}"/*.xml)
+    shopt -u nullglob
+
+    if ((${#xml_files[@]} == 0)); then
+        echo "No lit XML result files found under ${LOGS_DIR}" >&2
+        return 1
+    fi
+
+    python3 - "${xml_files[@]}" <<'PY'
```
pawosm-arm: is this explicit use of `python3` intentional?
Maybe something like `/usr/bin/env python3` would be safer?

arm-software/linux/build.sh
```diff
@@ -678,12 +643,27 @@ then
...
+# This can be enabled with the --xunit-xml-output option.
+# Each check target appends its own --xunit-xml-output path under LOGS_DIR.
+export LIT_OPTS="${LIT_OPTS:+${LIT_OPTS} }--ignore-fail"
+
 main
 trap : 0
 if ${INTERACTIVE}; then
     bash
 fi
+
```
pawosm-arm: NIT: please remove extra newline

arm-software/linux/build.sh
pawosm-arm: looks better now

pawosm-arm (dismissed):

pawosm-arm:
What happened to this PR? Where is the xml part of it now...?

pawosm-arm (dismissed):

arm-software/linux/build.sh
```diff
@@ -390,7 +390,8 @@ bootstrap_compiler_build() {
     run_command cmake --install . 2>&1 | tee -a "${LOGS_DIR}/bootstrap_compiler.txt"
     export PATH="${BUILD_DIR}/bootstrap_compiler/bin:${PATH}"
     bootstrap_compiler_default_config
-    run_command ninja "${NINJA_ARGS[@]}" check-all 2>&1 | tee -a "${LOGS_DIR}/bootstrap_compiler.txt"
+    LIT_OPTS="${LIT_OPTS} --xunit-xml-output=${LOGS_DIR}/bootstrap_check_all.xml" \
```
amilendra: All these test commands use a common repetitive pattern
I think it is worth encapsulating them into a run_test_command bash function that takes the log file name, xml output file name and check command as parameters.

amilendra (changes requested):

arm-software/linux/build.sh
pawosm-arm: yeah, would make it look better.

pawosm-arm (dismissed):

arm-software/linux/build.sh
vrukesh: Latest version of PR implements this.

arm-software/linux/build.sh
```diff
@@ -236,6 +236,15 @@ run_command() {
     "$@"
 }
 
+run_lit_ninja() {
```
amilendra: Let's use a generic name like `run_test_command`. That will keep the intention of the function easier to understand.

amilendra: Could you also please write a similar wrapper to the `check*` commands in `arm-software/linux/scripts/test_atfl_sanitizer.sh` as well? As they don't use log files or xml output files, it would be a much simpler wrapper.

amilendra (changes requested):

arm-software/linux/build.sh
amilendra: Because build and install is done using `cmake --build` and ` cmake --install`, if you have time, worth checking it the test wrapper could call something like `cmake --test` rather than explicitly calling `ninja`.

pawosm-arm: > Because build and install is done using `cmake --build` and ` cmake --install`, if you have time, worth checking it the test wrapper could call something like `cmake --test` rather than explicitly calling `ninja`.

Not sure if this as straightforward to use as `--build` and `--install`. I remember in the past we decided not to do it such way, but I don't remember what the challenge was.

vrukesh: @amilendra 

As discussed, updated the test function name. 
Other 2 comments are out of scope this PR.

amilendra (approved):

pawosm-arm (approved):

PR #882, opened 2026-06-11
[ATfE] Add picolibc deprecation roadmap

voltur01:
Add picolibc deprecation/transition to LLVM libc roadmap to the CHANGELOG.

Add picolibc.cfg to allow early migration of build scripts to support future non-default picolibc configuration.


vrukesh (approved):
LGTM

PR #884, opened 2026-06-12
[ATfE] Provide fallback libatomic and enable atomics in compiler_rt

voltur01:
This enables atomic operations in compiler_rt for Armv7+ and provides a libatomic-fallback.a for older architectures.

This adds a new sub-folder for shared-support libraries and the first such library that is libatomic-fallback.a

The library provides trivial implementation of libatomic functions, enough for testing. For targets that do not natively support atomic operations, users can override them and provide atomicity in the project appropriate way.


arm-software/embedded/arm-runtimes/CMakeLists.txt
```diff
@@ -152,6 +152,18 @@ else()
     set(cpu_family arm)
 endif()
 
+# compiler-rt's upstream atomic builtins require lock-free pointer-size
+# compare/exchange. Pre-Armv7 targets do not provide that, so they use the
+# weak fallback libatomic from shared-support instead.
```
smithp35: If a user has to manually do -latomic for the fallback case (presumably compiler-rt just goes into the same builtins library), is there a case for using a distinct name, for example `-lfallbackatomic`

arm-software/embedded/arm-runtimes/CMakeLists.txt
```diff
@@ -152,6 +152,18 @@ else()
...
+set(target_supports_lock_free_atomics ON)
+if(TARGET_ARCH MATCHES "^(arm|armeb)v[4-6]")
```
smithp35: The arm1136 mpcore has load and store exclusive and can support atomics. That should have architecture v6k.

My guess is that we won't want to have a specific library variant for v6k, but could be worth making sure that we permit v6k to build compiler-rt in case someone builds a manual variant.

arm-software/embedded/docs/llvmlibc.md
```diff
@@ -56,6 +56,12 @@ following command line options, in addition to `--target`, `-march` or
   in LLVM's libc in terms of the Arm semihosting API (or else provide
   an alternative implementation of those functions yourself)
 
+* `-latomic` to include weak fallback definitions for compiler-emitted
+  `__atomic_*` libcalls. This is mainly useful for targets such as Armv4T,
+  Armv5TE, and Armv6-M where the compiler may emit calls to helper functions
```
smithp35, line 61: v4 and v5TE have the SWP instruction (https://developer.arm.com/documentation/ddi0406/c/Application-Level-Architecture/Instruction-Details/Alphabetical-list-of-instructions/SWP--SWPB) that could in theory be used to make an implementation of libatomic. However I don't think it is worth it due to the size of the use case, and the limitations of SWP (relies on unspecified properties of the system that may not universally hold).

arm-software/embedded/docs/shared-support.md
```diff
@@ -0,0 +1,73 @@
...
+multilib sysroot and can be used with `picolibc`, `newlib`, `newlib-nano`, and
+LLVM libc variants.
+
+## Atomic operation helpers
+
+LLVM can lower atomic operations that are not supported inline by the target to
+runtime helper functions such as `__atomic_load_4`,
+`__atomic_exchange_4`, and `__atomic_fetch_add_4`. This is common on older or
+small Arm targets where the maximum lock-free atomic size is smaller than the
+operation being compiled, for example Armv4T, Armv5TE, and Armv6-M.
```
smithp35, line 14: I think the one peole hit the most in practice is an 8-byte sized atomic on M-profile as at least on v7-M the max size is 4-bytes.

arm-software/embedded/arm-runtimes/CMakeLists.txt
smithp35: One possibe reason is that the fallback atomic library when built for a system that can support atomics does produce a viable (I hope) useful implementation.

However on most of the targets where that would be useful we would be using compiler-rt's implementation.

Do we know what the library order would be if compiler-rt has atomic.o and there's also a libatomic.a? If compiler-rt comes first then we'll never use the fallback atomic library, if the fallback atomic comes first, then won't we be making the user experience worse as presumably compiler-rt's routines are better?

arm-software/embedded/docs/shared-support.md
```diff
@@ -0,0 +1,73 @@
+# Shared Support Libraries
+
+Arm Toolchain for Embedded provides small support libraries that are shared by
+multiple C library variants. These libraries are installed in the selected
+multilib sysroot and can be used with `picolibc`, `newlib`, `newlib-nano`, and
+LLVM libc variants.
+
+## Atomic operation helpers
+
+LLVM can lower atomic operations that are not supported inline by the target to
```
smithp35: lower is quite a compiler specific term. Perhaps
LLVM can emit calls to runtime helper functions such as `__atomic_load_4` when the target does not have instructions that handle an atomic operation inline.

arm-software/embedded/docs/shared-support.md
```diff
@@ -0,0 +1,73 @@
...
+runtime helper functions such as `__atomic_load_4`,
+`__atomic_exchange_4`, and `__atomic_fetch_add_4`. This is common on older or
+small Arm targets where the maximum lock-free atomic size is smaller than the
+operation being compiled, for example Armv4T, Armv5TE, and Armv6-M.
+
+ATfE provides `libatomic.a` with weak definitions for the LLVM `__atomic_*`
```
smithp35: If we are also incorporating compiler-rt's implementation should we mention that here. I can't see why anyone would want to use the fallback library if compiler-rt is present. Especially as the implementations have different properties.

smithp35 (commented):
I think it could be useful to split out the inclusion of compiler-rt's atomic.c and this fallback path.

I'm a bit nervous about the inlclusion of atomic calls in the fallback library. Short of someone making the call directly, I'm unsure of when the compiler would ever use it. For example if <atomic.h> or <atomic> can check the lock free nature and do the operation inline, then it wouldn't call out to libatomic. If it does call out to libatomic, then won't compiler-rt handle it.

I think we should have some tests for hand-written library code. Particularly if is using atomic instructions. EDIT, to clarify; some of the code-paths will be tested indirectly via the libc++ tests, however if we're building the library for all multilib variants as well as compiler-rt atomic then I'm not sure all the code-paths will be executed by the existing tests.

arm-software/embedded/arm-runtimes/CMakeLists.txt
voltur01: Thank you for the review!

First, to clarify, this libatomic is only built for v4/5/6 now, so there are should not be variants in which both compiler_rt atomics are enabled and this libatomic is provided. If they were both provided it looks like user provided libraries go first on the linker command line, so they would indeed prevent compiler_rt versions from being used.

On the naming, it is a good point, however on one side calling it libatomic-fallback would be clear, however I understand that during migration from GCC one would have to update the build scripts to amend the name. An intermediate option could be to keep the library name as libatomic (for compatibility), but update the source folder/file name to be clear for developers that this is intended to be a fallback - what do you think?

arm-software/embedded/arm-runtimes/CMakeLists.txt
voltur01: Thank you for the pointer - I will update the check.

arm-software/embedded/docs/llvmlibc.md
voltur01: Indeed, also v4 and v5 are strictly speaking experimental, so spending too much effort there without a clear use case does not make sense.

arm-software/embedded/docs/shared-support.md
voltur01: Updating now.

arm-software/embedded/docs/shared-support.md
voltur01: For v7 there will be compiler_rt implementations provided.

__atomic_fetch_add_4 came from the current libc++ failure with v6-M.

arm-software/embedded/docs/shared-support.md
voltur01: Good point, extended.

voltur01:
Please review the updates and let me know what you think about the name for the library that I will address separately.

voltur01:
I decided that it is clearer to remove the real atomic code and then call the library `libatomic-fallback.a` (also do not allow building it for targets that do support required atomic operations) - please have a look.

arm-software/embedded/arm-runtimes/CMakeLists.txt
```diff
@@ -1148,6 +1185,85 @@ else() # if not ENABLE_CXX_LIBS
...
+        STAMP_DIR ${PROJECT_PREFIX}/shared-support/${VARIANT_BUILD_ID}/stamp
+        BINARY_DIR ${PROJECT_PREFIX}/shared-support/${VARIANT_BUILD_ID}/build
+        DOWNLOAD_DIR ${PROJECT_PREFIX}/shared-support/${VARIANT_BUILD_ID}/dl
+        TMP_DIR ${PROJECT_PREFIX}/shared-support/${VARIANT_BUILD_ID}/tmp
+        SOURCE_DIR ${TOOLCHAIN_SOURCE_DIR}/shared-support
+        INSTALL_DIR ${TEMP_LIB_DIR}
+        DEPENDS cxxlibs-install
+        CMAKE_ARGS
+        ${compiler_launcher_cmake_args}
+        -DCMAKE_AR=${LLVM_BINARY_DIR}/bin/llvm-ar${CMAKE_EXECUTABLE_SUFFIX}
```
smithp35, line 1210: Maybe not for this patch, but could be worth raising a follow up ticket to factor out the common parts into a separate variable that can be expanded here. I think `-DMAKE_AR` and friends are repeated 3 times already in separate invocations in the file.

arm-software/embedded/arm-runtimes/CMakeLists.txt
```diff
@@ -152,6 +152,21 @@ else()
...
+# compare/exchange. Armv6K and Armv6ZK have load/store exclusive support, but
+# earlier Arm architectures and Armv6-M use the weak libatomic-fallback from
+# shared-support instead.
+set(target_supports_lock_free_atomics ON)
+if(TARGET_ARCH MATCHES "^(arm|armeb)v[45]")
+    set(target_supports_lock_free_atomics OFF)
+elseif(TARGET_ARCH MATCHES "^(arm|armeb)v6" AND NOT TARGET_ARCH MATCHES "^(arm|armeb)v6z?k")
+    set(target_supports_lock_free_atomics OFF)
+endif()
+set(use_libatomic_fallback OFF)
```
smithp35: `Use_libatomic_fallback` is the inverse of `target_supports_lock_free_atomics` do we need two variables?

Not sure how CMake deals with OFF ON. I can see that we expand `{use_libatomic_fallback}` later on, I guess there's no expansion of `{NOT use_libatomic_fallback}`

arm-software/embedded/shared-support/atomic-fallback/atomic.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+// Weak libatomic ABI exports. Applications can provide strong definitions for
+// any of these symbols to use target-specific synchronization, such as masking
+// interrupts around a non-lock-free operation.
+#define WEAK extern "C" __attribute__((weak))
+#define ASM_LABEL(name) asm(name)
+
+// Size-specialized libatomic API functions. Clang emits calls to these for
+// fixed-width atomic operations, for example __atomic_fetch_add_4.
+#define DEFINE_ATOMIC_SIZE(N, TYPE)                                                                                  \
```
smithp35: In the libatomic docs the notation `_n` is used. I wonder if DEFINE_ATOMIC_N would work better.
https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html

arm-software/embedded/shared-support/atomic-fallback/atomic.cpp
```diff
@@ -0,0 +1,158 @@
...
+  WEAK TYPE __atomic_load_##N(TYPE *ptr, int) { return fallback_load(ptr); }                                         \
+  WEAK void __atomic_store_##N(TYPE *ptr, TYPE value, int) { fallback_store(ptr, value); }                           \
+  WEAK TYPE __atomic_exchange_##N(TYPE *ptr, TYPE value, int) { return fallback_exchange(ptr, value); }              \
+  WEAK bool __atomic_compare_exchange_##N(TYPE *ptr, TYPE *expected, TYPE desired, int, int) {                       \
+    return fallback_compare_exchange(ptr, expected, desired);                                                        \
+  }                                                                                                                  \
+  WEAK TYPE __atomic_fetch_add_##N(TYPE *ptr, TYPE value, int) { return fallback_fetch_modify(ptr, value, Add{}); }  \
```
smithp35: I think we could use a lambda directly rather than a functor class with an overloaded `operator()` as we don't need a template lambda.

for example:
```
WEAK TYPE __atomic_fetch_sub_##N(TYPE *ptr, TYPE value, int) { 
  return fallback_fetch_modify(ptr, value, [](TYPE x, TYPE y){ return x - y; });
}
```

arm-software/embedded/shared-support/atomic-fallback/atomic.cpp
```diff
@@ -0,0 +1,158 @@
...
+  WEAK TYPE __atomic_fetch_nand_##N(TYPE *ptr, TYPE value, int) { return fallback_fetch_modify(ptr, value, Nand{}); }
+
+DEFINE_ATOMIC_SIZE(1, std::uint8_t)
+DEFINE_ATOMIC_SIZE(2, std::uint16_t)
+DEFINE_ATOMIC_SIZE(4, std::uint32_t)
+DEFINE_ATOMIC_SIZE(8, std::uint64_t)
+
+// Generic libatomic API functions. The C++ source uses internal shared-support
+// names with asm labels to avoid redeclaring Clang's builtin function names.
+WEAK bool __shared_support_atomic_is_lock_free(std::size_t, const volatile void *) noexcept
```
smithp35, line 112: I think it would be easier to read if the aliases were interleaved with the definition, and move the definition of ASM_LABEL to just before use. As an aside I think `ASM_LABEL()` may be less readable than `asm()`. 
```
#define ASM_LABEL(name) asm(name)

WEAK bool __shared_support_atomic_is_lock_free(std::size_t, const volatile void *) noexcept ASM_LABEL("__atomic_is_lock_free");

WEAK bool __shared_support_atomic_is_lock_free(std::size_t, const volatile void *) noexcept { return false; }
```

arm-software/embedded/shared-support/atomic-fallback/atomic.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+struct Xor {
+  template <typename T> T operator()(T lhs, T rhs) const { return lhs ^ rhs; }
+};
+
+struct Nand {
+  template <typename T> T operator()(T lhs, T rhs) const { return ~(lhs & rhs); }
+};
+
+void copy_bytes(void *dst, const void *src, std::size_t size) {
```
smithp35, line 53: Isn't this just `memcpy`? I think it is fair game to use that directly as implementations are allowed to assume presence of memcpy even without a C-library.

arm-software/embedded/shared-support/atomic-fallback/atomic.cpp
```diff
@@ -0,0 +1,158 @@
...
+  auto *dst_bytes = static_cast<unsigned char *>(dst);
+  auto *src_bytes = static_cast<const unsigned char *>(src);
+  for (std::size_t i = 0; i != size; ++i)
+    dst_bytes[i] = src_bytes[i];
+}
+
+bool equal_bytes(const void *lhs, const void *rhs, std::size_t size) {
```
smithp35, line 60: Similarly, can we use `memcmp` instead?

smithp35 (commented):
I think the description needs to be updated for the library change.

arm-software/embedded/arm-runtimes/CMakeLists.txt
voltur01: Yes, there is a bit if tension between descriptive names and duplication - I tried to rely on `target_supports_lock_free_atomics` and use `if` instead of direct expansion (with the next update).

arm-software/embedded/arm-runtimes/CMakeLists.txt
voltur01: Good point, I created an issue and will have a look later.

arm-software/embedded/shared-support/atomic-fallback/atomic.cpp
voltur01: In theory yes, however I am pretty sure adding the dependency on libc in such a low level library will break LLVM libc hermetic testing (again - we run into this with the bootcode before).
I can try using `__builtin_memcpy` directly, however I understand it is not guaranteed to avoid using `memcpy()` so can potentially break anyway.

voltur01: I am adding a comment to explain this.

arm-software/embedded/shared-support/atomic-fallback/atomic.cpp
voltur01: Good suggestion! Updated.

arm-software/embedded/shared-support/atomic-fallback/atomic.cpp
voltur01: A great suggestion - done!

arm-software/embedded/shared-support/atomic-fallback/atomic.cpp
voltur01: Agreed, updated.

voltur01:
I uploaded the changes as discussed in the comments.

arm-software/embedded/shared-support/atomic-fallback/atomic.cpp
smithp35: There could be a danger that clang will idiom recognize this as memcpy and replace it with memcpy https://github.com/llvm/llvm-project/blob/main/llvm/lib/Transforms/Scalar/LoopIdiomRecognize.cpp

Empirically it isn't happening. It seems like I need to add `__restrict` to the `dst` and `src` pointers to trigger.

Not sure if there's anything we can do besides adding to the comment to say don't add `__restrict` to avoid LoopIdiomRecognition. 

There isn't currently any idiom recognition for memcmp.

smithp35 (dismissed):
One minor suggestion to the comment for the non-mempcy to make sure we don't run into the LoopIdiomRecognizer.

LGTM.

arm-software/embedded/shared-support/atomic-fallback/atomic.cpp
voltur01: Expanded.

smithp35 (approved):
Thanks for the update. LGTM.

PR #885, opened 2026-06-15
[ATfE] Switch to using llvm-lit internal shell

vrukesh:
Upstream LLVM is removing support for the lit external shell and has deprecated the `execute_external` argument in `ShTest`.

ATfE currently defaults to using the external shell through the standard LLVM lit configuration. The resulting deprecation warning causes our `lit.cfg` generation step to fail.

Our tests do not depend on any external-shell-specific functionality, so it is safe to switch to the internal shell. This change removes the use of the external shell and has been verified to work without affecting test behaviour.

Refer to:
- [RFC 90951](https://discourse.llvm.org/t/rfc-removal-of-the-lit-external-shell/90951)
- [RFC 80179](https://discourse.llvm.org/t/rfc-enabling-the-lit-internal-shell-by-default/80179)
- Deprecation in upstream PR: [203689](https://github.com/llvm/llvm-project/pull/203689)

### DIFF
--- a/arm-software/embedded/arm-runtimes/test-support/gen-picolibc-lit-tests.py
+++ b/arm-software/embedded/arm-runtimes/test-support/gen-picolibc-lit-tests.py
@@ -90,7 +90,7 @@ lit.llvm.initialize(lit_config, config)
 
 config.name = "%CONFIG_NAME%"
 config.suffixes = [".test"]
-config.test_format = lit.formats.ShTest(not lit.llvm.llvm_config.use_lit_shell)
+config.test_format = lit.formats.ShTest()
 config.test_source_root = os.path.join(os.path.dirname(__file__), "tests")
 """
     with open(os.path.join(args.output, "lit.cfg.py"), "w", encoding="utf-8") as f:
--- a/arm-software/embedded/packagetest/lit.cfg.py
+++ b/arm-software/embedded/packagetest/lit.cfg.py
@@ -11,7 +11,7 @@ from lit.llvm.subst import FindTool, ToolSubst
 # Configuration file for the 'lit' test runner.
 
 config.name = "package"
-config.test_format = lit.formats.ShTest(not llvm_config.use_lit_shell)
+config.test_format = lit.formats.ShTest()
 config.suffixes = [".c", ".cpp", ".test"]
 config.excludes = ["CMakeLists.txt", "README.md"]
 config.test_source_root = os.path.dirname(__file__)
--- a/arm-software/embedded/test/lit.cfg.py
+++ b/arm-software/embedded/test/lit.cfg.py
@@ -9,7 +9,7 @@ from lit.llvm import llvm_config
 # Configuration file for the 'lit' test runner.
 
 config.name = "Arm Toolchain for Embedded"
-config.test_format = lit.formats.ShTest(not llvm_config.use_lit_shell)
+config.test_format = lit.formats.ShTest()
 config.suffixes = [".c", ".cpp", ".test"]
 config.excludes = ["CMakeLists.txt"]
 config.test_source_root = os.path.dirname(__file__)


smithp35 (commented):
Can you rewrite the title to say what the changes are rather than changes to do with something else? For example "Remove deprecated external shell in lit.cfg"

For the non-expert in lit, it would be good to add a reason to the description for why removing it is OK? For example it isn't needed any more.

vrukesh:
@smithp35 

Kindly check if title and description are fine now.

smithp35:
> Since ShTest execute_external=True deprecation breaks Arm Embedded lit config generation, this PR removes all those patterns.

> Since ShTest execute_external=True deprecation breaks Arm Embedded lit config generation, this PR removes all those patterns.

We know it is deprecated and we will need to remove it. What I'm after is why is this the right change to make? It could be that some internal changes to lit have meant that this is no longer needed and we can safely remove it. Alternatively we could see some side-effects, but these are acceptable because ... Maybe we are losing some functionality and there's an alternative implementation?

If the answers are buried somewhere in the Discourse thread and PR, can you quote the relevant parts.

vrukesh:
@smithp35 

Kindly check if description is fine now.

smithp35:
That's better, if a little verbose. With the additional information I can determine that your change is going cause us to switch to the lit internal shell. The key bit of information I want to know as a reviewer is what is the impact of the change as we were explicitly using the external shell before? It could be nothing, all our tests can run on the internal shell, it could be that I don't know and want to find out.

Making a few assumptions:
* The deprecation warning is causing a failure because we have some warnings as errors, or are not expecting a warning. Otherwise we would have till LLVM 24 to remove it.
* The internal shell is sufficient for our tests so we no longer require the external shell.

I suggest something like:

[ATfE]: Switch to using llvm-lit internal shell

Upstream LLVM has removed support for the lit external shell and deprecated execute_external=True in ShTest. We are currently using the external shell in our lit.cfg and the deprecation warning is causing our lit.cfg generation step to fail. As we no longer require the functionality of the external shell, it is safe for us to switch to the internal shell.

See upstream RFC's for more details
* [RFC 80179](https://discourse.llvm.org/t/rfc-enabling-the-lit-internal-shell-by-default/80179)
* [RFC 90951](https://discourse.llvm.org/t/rfc-removal-of-the-lit-external-shell/90951)
Deprecation in upstream PR https://github.com/llvm/llvm-project/pull/203689

dcandler:
One nit: `execute_external=True` isn't being deprecated, the `execute_external` argument is being deprecated. The ATfE scripts don't even explicitly set `execute_external` to `True`, so I'm confused why that has gotten repeated so much.

Like many of the existing lit scripts, the ATfE scripts simply set `execute_external` to whatever the `use_lit_shell` LLVM lit configuration variables says it should be. `use_lit_shell` means use the internal lit shell, so `not llvm_config.use_lit_shell` means don't use the internal shell, i.e. use the external shell. Our scripts don't actually set `use_lit_shell`, so we haven't expressed any preference for internal or external, and therefore `use_lit_shell` stays at its default of `False` unless the `LIT_USE_INTERNAL_SHELL` environment variable is set. I don't think internal or external has ever mattered for us, and the code is there just to respect the configuration/environment variables if a user overrides them.

I would agree with Peter that the PR summary is getting quite verbose and isn't getting across the key information:
- Upstream LLVM is deprecating the lit external shell, and enabling `execute_external` now causes an error to warn about this.
- ATfE has until now defaulted to using the external shell.
- The tests used by ATfE do not rely on the external shell and can work on the internal one, so we can safely switch to the internal shell.

A real petty nit: our repo generally follows LLVM's policy for commit messages, so you don't need a colon in the title after the tags, i.e. `[ATfE] Switch to using llvm-lit internal shell`

vrukesh:
Kindly check if title and description are fine now.

smithp35:
First half LGTM. Thanks for verifying that the tests are unaffected by the change! 

I think you can drop the quotes now. They have been adequately summarised by the first part. Suggest removing:
> [RFC 80179](https://discourse.llvm.org/t/rfc-enabling-the-lit-internal-shell-by-default/80179) mentions:
LLVM’s main test running tool, lit, is used as a test driver for end-to-end tests. It is designed to parse and execute shell-like commands, which are specified in the comments of source files. Currently, lit’s default behavior is to use the system’s shell when executing RUN commands. However, lit also has an internal shell implementation, which is currently the default for Windows systems. This internal shell can be enabled regardless of platform by turning on the LIT_USE_INTERNAL_SHELL environment variable.

>
> [RFC 90951](https://discourse.llvm.org/t/rfc-removal-of-the-lit-external-shell/90951) mentions:
lit internal shell has been enabled by default in upstream LLVM. There are no upstream users of the external shell, and downstream users are starting to migrate off, I think it makes sense to remove support for the external shell at this time.

vrukesh:
@smithp35 

Done !

smithp35 (approved):
LGTM thanks.

PR #888, opened 2026-06-16
[ATfE] Enable ASSUME_ROUND_NEAREST_ONLY in LLVM libc

statham-arm:
This setting, newly introduced in upstream commit 582643f1ec62d0c, conditions out code to handle non-default rounding modes in the libm functions. We were not reliably able to handle directed rounding modes correctly across ATfE anyway, since it's not implemented by the software FP library. Removing support for it saves code size and makes a number of existing test failures go away.

It's slightly awkward to set this option, because it's not a simple definition on the cmake command line: instead, it's one of many bit flags that are set in LIBC_CONF_MATH_OPTIMIZATIONS, which is a C integer expression. Moreover, the default bare-metal libc configuration sets a lot of flags in that variable already, so I had to extract the existing setting from the JSON file in order to decide on the changed version to set on our command lines. Happily, cmake 3.19 introduced JSON-handling facilities that make this easy, and we set cmake_minimum_required to 3.20.


voltur01 (approved):
Looks good, thank you for taking care!

PR #891, opened 2026-06-17
[ATfE] Apply timeout multiplier to slow picolibc tests

dcandler:
Certain picolibc tests take significantly longer to run than others, due to the test running through a lot of loops combined with a lower execution speed of the model. Most tests complete in seconds, but some can take minutes.

Rather than set an excessively long timeout for all tests, this patch specifies a longer timeout for the tests observed to take the longest (variations on test-long and test-mem), on the variants observed to have slower performing models (arm v6/v7 big endian and v8.1m main PACBTI optimized for code size).

Running the tests through meson usually only allows a timeout multiplier to be set for all tests, however since ATfE has a lit wrapper around the tests to invoke meson individually for each test, each test can instead have its own multiplier. This has been split off into it's own option in the lit generator script: one option for a global multiplier and another specifically for the slow tests. The CMake target for running the tests directly through meson cannot support both, but this target is not the default and was only retained for diagnostics.

### DIFF
--- a/arm-software/embedded/arm-runtimes/CMakeLists.txt
+++ b/arm-software/embedded/arm-runtimes/CMakeLists.txt
@@ -537,8 +537,9 @@ if(C_LIBRARY STREQUAL picolibc)
         # Timeout multipliers. The actual timeout value used in testing is:
         # Timeout in seconds (defined in the picolibc's testing specs) x Timeout multiplier
         # In general, tests execute slower with FVPs compared to qemu, but specific tests are
-        # more impacted by the model performance than others due to significant looping. This
-        # timeout is only applied to those tests, via lit.
+        # more impacted by the model performance than others due to significant looping. Running
+        # the tests directly through meson only allows a timeout to be set for all tests, but
+        # using the lit wrapper allows the timeout to be customized per test.
         # Certain variants can also be slower due to the particular model, or specific
         # features enabled, for example:
         # * Armv6m big-endian
@@ -550,7 +551,7 @@ if(C_LIBRARY STREQUAL picolibc)
             else()
                 set(timeout_multiplier 2)
             endif()
-            set(picolibc_timeout_arg --slow-timeout ${timeout_multiplier})
+            set(picolibc_timeout_arg --slow-tests-timeout-multiplier ${timeout_multiplier})
         endif()
 
         # Step target to run the picolibc test via meson.
--- a/arm-software/embedded/arm-runtimes/CMakeLists.txt
+++ b/arm-software/embedded/arm-runtimes/CMakeLists.txt
@@ -536,26 +536,28 @@ if(C_LIBRARY STREQUAL picolibc)
 
         # Timeout multipliers. The actual timeout value used in testing is:
         # Timeout in seconds (defined in the picolibc's testing specs) x Timeout multiplier
-        # In general, tests execute slower with FVPs compared to qemu. However certain
-        # variants can also be slower due to features enabled, for example: 
+        # In general, tests execute slower with FVPs compared to qemu, but specific tests are
+        # more impacted by the model performance than others due to significant looping. This
+        # timeout is only applied to those tests, via lit.
+        # Certain variants can also be slower due to the particular model, or specific
+        # features enabled, for example:
         # * Armv6m big-endian
         # * Armv7m big-endian
         # * Armv8.1-M Main PACBTI optimized for code size
         if(TEST_EXECUTOR STREQUAL fvp)
-            if(VARIANT MATCHES "armebv7a")
-                set(timeout_multiplier 3)
+            if(VARIANT MATCHES "^(armebv|^armv8\.1m\.main.*pacret.*size)")
+                set(timeout_multiplier 5)
             else()
                 set(timeout_multiplier 2)
             endif()
-        else()
-            set(timeout_multiplier 1)
+            set(picolibc_timeout_arg --slow-timeout ${timeout_multiplier})
         endif()
 
         # Step target to run the picolibc test via meson.
         ExternalProject_Add_Step(
             picolibc
             check-meson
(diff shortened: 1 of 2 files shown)


arm-software/embedded/arm-runtimes/test-support/gen-picolibc-lit-tests.py
```diff
@@ -45,7 +45,14 @@ def main():
         help="Name to give the lit suite.",
     )
     arg_parser.add_argument(
-        "--timeout-multiplier", type=float, help="Timeout multiplier (float)."
+        "--timeout-multiplier",
+        type=float,
+        help="Global timeout multiplier, applied to all tests (float).",
+    )
+    arg_parser.add_argument(
+        "--slow-timeout",
```
voltur01: Does it make sense to keep the name consistent with the other one as `--slow-timeout-multilplier`?

arm-software/embedded/arm-runtimes/CMakeLists.txt
```diff
@@ -536,26 +536,28 @@ if(C_LIBRARY STREQUAL picolibc)
 
         # Timeout multipliers. The actual timeout value used in testing is:
         # Timeout in seconds (defined in the picolibc's testing specs) x Timeout multiplier
-        # In general, tests execute slower with FVPs compared to qemu. However certain
-        # variants can also be slower due to features enabled, for example: 
+        # In general, tests execute slower with FVPs compared to qemu, but specific tests are
+        # more impacted by the model performance than others due to significant looping. This
+        # timeout is only applied to those tests, via lit.
```
voltur01: I understand that in the end the timeout will be still enforced by meson via `-t` option, but this option will be set differently for slow tests individually, correct?

It is a bit confusing together with the overall PR description that seems to state the meson is not involved anymore - may benefit from a clarification.

arm-software/embedded/arm-runtimes/test-support/gen-picolibc-lit-tests.py
```diff
@@ -45,7 +45,14 @@ def main():
         help="Name to give the lit suite.",
     )
     arg_parser.add_argument(
-        "--timeout-multiplier", type=float, help="Timeout multiplier (float)."
+        "--timeout-multiplier",
```
vhscampos: Rename this option to `--global-timeout-multiplier`.

arm-software/embedded/arm-runtimes/test-support/gen-picolibc-lit-tests.py
vhscampos: I suggest the option is named `--slow-tests-timeout-multiplier`. Simply "slow timeout" is not a clear term.

dcandler: Funnily enough `--global-timeout-multiplier` and `--slow-test-timeout-multiplier` were my initial choices, but I worried the names were getting a bit too long. However given this script is only getting called by other scripts rather than users, clarity is probably better here.

arm-software/embedded/arm-runtimes/CMakeLists.txt
dcandler: I've updated both the comments and the PR summary.

arm-software/embedded/arm-runtimes/CMakeLists.txt
```diff
@@ -573,7 +576,7 @@ if(C_LIBRARY STREQUAL picolibc)
                 --build <BINARY_DIR>
                 --output ${picolibc_lit_dir}
                 --name picolibc-${variant_xml_name}
-                --timeout-multiplier ${timeout_multiplier}
+                ${picolibc_timeout_arg}
```
vhscampos, line 579: Doesn't it need an option name? IIUC the timeout arg is not positional

arm-software/embedded/arm-runtimes/CMakeLists.txt
```diff
@@ -536,26 +536,29 @@ if(C_LIBRARY STREQUAL picolibc)
...
+        # Certain variants can also be slower due to the particular model, or specific
+        # features enabled, for example:
         # * Armv6m big-endian
         # * Armv7m big-endian
         # * Armv8.1-M Main PACBTI optimized for code size
         if(TEST_EXECUTOR STREQUAL fvp)
-            if(VARIANT MATCHES "armebv7a")
-                set(timeout_multiplier 3)
+            if(VARIANT MATCHES "^(armebv|^armv8\.1m\.main.*pacret.*size)")
+                set(timeout_multiplier 5)
```
vhscampos, line 550: `timeout_multiplier` doesn't seem to be used?

arm-software/embedded/arm-runtimes/CMakeLists.txt
voltur01: LGTM, thank you for the update!

arm-software/embedded/arm-runtimes/CMakeLists.txt
dcandler: `picolibc_timeout_arg` contains both the option name and the value, so that on variants that don't need any special multiplier, `picolibc_timeout_arg` is blank and adds nothing to the command.

arm-software/embedded/arm-runtimes/CMakeLists.txt
dcandler: `timeout_multiplier` becomes part of `picolibc_timeout_arg`, line 554.

vhscampos: Line 554 is in the else branch, so it's either or. Am I missing something?

dcandler: It was previously in an `else` branch: line 550 was removed though.

vhscampos: You are right. Sorry about the noise

vhscampos (approved):

PR #892, opened 2026-06-18
[ATfL] Enable OpenMP tests

vrukesh:
Since the required dependencies are not found, check-openmp is not currently executed.

This PR enables required FileCheck and static library dependencies for OpenMP test execution.

Also, add the runtime library paths required for OpenMP tests and re-enable check-openmp result validation.

### DIFF
--- a/arm-software/linux/build.sh
+++ b/arm-software/linux/build.sh
@@ -489,6 +489,8 @@ product_build() {
 }
 
 static_libomp_build() {
+    rm -rf "${BUILD_DIR}/stage/static_libomp_build"
+    rm -rf "${BUILD_DIR}/stage/static_libomp_install"
     mkdir -p "${BUILD_DIR}/stage/static_libomp_build"
     cd "${BUILD_DIR}/stage/static_libomp_build"
     bootstrap_compiler_default_config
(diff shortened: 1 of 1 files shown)


arm-software/linux/build.sh
```diff
@@ -514,18 +516,29 @@ static_libomp_build() {
...
+        -DCMAKE_DISABLE_FIND_PACKAGE_Clang=TRUE \
+        -DClang_DIR=Clang_DIR-NOTFOUND \
+        -DLLVM_DIR="${BUILD_DIR}/stage/product_build/lib/cmake/llvm" \
+        -DLLVM_TOOLS_BINARY_DIR="${BUILD_DIR}/stage/product_build/bin" \
+        -DLLVM_EXTERNAL_LIT="${BUILD_DIR}/stage/product_build/bin/llvm-lit" \
+        -DOPENMP_NOT_EXECUTABLE:FILEPATH="${BUILD_DIR}/stage/product_build/bin/not" \
         2>&1 | tee "${LOGS_DIR}/static_libomp.txt"
     run_command cmake --build . "${CMAKE_BUILD_ARGS[@]}" 2>&1 | tee -a "${LOGS_DIR}/static_libomp.txt"
-    rm -rf "${ATFL_DIR}.keep" "${ATFL_DIR}.libs"
-    mv "${ATFL_DIR}" "${ATFL_DIR}.keep"
```
pawosm-arm, line 522: any reason this `.keep`/`.libs` swap had to be removed? How do you plan to control whether the `cmake --install .` step below is not overwriting anything in `${ATFL_DIR}` with worse (static) versions?

pawosm-arm (changes requested):

arm-software/linux/build.sh
pawosm-arm: What we want to obtain here is a copy of the .installed. lib*.a files landing in `${ATFL_DIR}/lib/${ATFL_TARGET_TRIPLE}` (the `mkdir -p` doesn't hurt, although it would be seriously bad if this directory didn't exist thus far). Nothing else in `${ATFL_DIR}` should change, and without this swap that's being removed here, one cannot ensure that. And no, using different install prefix is not the right solution: all the build steps of the product binaries should consistently use the same installation prefix.

pawosm-arm:
The libomp codebase is being tested in the previous stage, it's only the static build of it that is not being tested. Instead of power-struggle with AI bots, I'd rather abandon this one and preserve our energy resources to come up with a proper fix in the upstream LLVM which would allow us to get rid of this entire stage.
Namely, `flang-rt` runtime library has it right, there are two CMake flags: `FLANG_RT_ENABLE_SHARED` and `FLANG_RT_ENABLE_STATIC` which are not mutually exclusive and we're setting both to `ON`, and there's no need to create a separate stage for building flang-rt static. If the `libomp` runtime library could also have them flags (`LIBOMP_ENABLE_SHARED` AND `LIBOMP_ENABLE_STATIC` NOT mutually exclusive), we could set both to `ON` and get rid of this entire stage. The equivalent single-flag setting that `libomp` currently has IS either shared-or-static but not both, and it's bad.

vrukesh:
Abandoning PR as suggested above.

PR #893, opened 2026-06-19
[ATfE] Fix missing __atomic_fetch_add_4 in packagetest

voltur01:
This adds use of newly introduced libatomic-fallback.a in the packagetest to provide the missing builtin.


smithp35 (approved):
LGTM.

PR #896, opened 2026-06-19
[ATfL][build.sh] Continue despite test failures

amilendra:
It is becoming increasingly common for LLVM test failures to bring down our whole build process resulting in no packages to run against our other test suites.
Therefore consider test failures as soft failures, thus allowing the build to continue till the package creation stage.

A future patch will signal the caller Jenkins job to mark test failures in Amber.

### DIFF
--- a/arm-software/linux/build.sh
+++ b/arm-software/linux/build.sh
@@ -241,8 +241,10 @@ run_test_command() {
     local log_file="$2"
     shift 2
 
-    LIT_OPTS="${LIT_OPTS} --xunit-xml-output=${xml_output}" \
-        run_command ninja "${NINJA_ARGS[@]}" "$@" 2>&1 | tee -a "${log_file}"
+    if ! LIT_OPTS="${LIT_OPTS} --xunit-xml-output=${xml_output}" \
+        run_command ninja "${NINJA_ARGS[@]}" "$@" 2>&1 | tee -a "${log_file}"; then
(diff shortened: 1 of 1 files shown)


pawosm-arm:
Hey, the build.sh script is also being started in command line (outside CI environment, as described here: https://github.com/arm/arm-toolchain/blob/arm-software/arm-software/linux/build.sh-HOWTO.md). If the tests are failing during the process and it continues until completion, the users who use this script in the command line may not be so eager to look through the log just to make sure there was no failure. Is there any console output from build.sh which could make the user aware that there was a problem during test cases execution (also pointing at the stage during which this has happened)?

amilendra:
> Hey, the build.sh script is also being started in command line (outside CI environment, as described here: https://github.com/arm/arm-toolchain/blob/arm-software/arm-software/linux/build.sh-HOWTO.md). If the tests are failing during the process and it continues until completion, the users who use this script in the command line may not be so eager to look through the log just to make sure there was no failure. Is there any console output from build.sh which could make the user aware that there was a problem during test cases execution (also pointing at the stage during which this has happened)?

Yes, there is some warning messages that are posted, but those are buried with the rest of the build logs, so unless the user is keeping his eyes glued in the screen it is very easy to miss. I'll make it more visible when I add support for this script to return a non-zero return code that Jenkins (or any other caller process) can use to identify test failure. I think as a first step this is going in the right direction to allow us at least have builds even if some tests fail.

pawosm-arm (approved):

PR #900, opened 2026-06-23
[ATFE] Remove resolved libatomic XFAILs for older Arm architectures.

simpal01:
Remove libcxx XFAIL entries that are now expected to pass after adding the fallback libatomic implementation for older Arm architectures here https://github.com/arm/arm-toolchain/pull/884/


voltur01 (approved):
LGTM, thank you!

PR #904, opened 2026-06-26
Update to Fast Models 11.31 for most FVPs.

statham-arm:
The newer downloadable archives unpack to different filesystem layouts, and the one containing the Crypto.so plugin no longer has an installer script inside it, so get_fvps.sh is substantially changed (not just rewriting URLs), and run_fvp.py has to change correspondingly for the new internal pathnames.

The update brings in a useful feature: when FM 11.31 emulates an AArch32 system supporting both Arm and Thumb instruction set states, it chooses the initial state to match the Thumb bit of the entry point of the ELF image you tell it to load and run. This unblocks testing of LLVM libc in some of our library variants, where the entry point happened to be the opposite state from FM 11.28's default.

The exception is the Corstone-310 FVP, which doesn't have an update available. But we use that for M-profile testing, which doesn't have that problem in the first place, since the M profile only supports Thumb, and no confusion is possible.


statham-arm:
This PR is currently a not-fully-tested draft. I've checked all the download URLs, and all the pathnames within them, and I've run some test binaries with the updated x86-64 hosted models, but I haven't done a full test on both x86-64 and AArch64 hosted Linux to make sure _all_ the models work sensibly. I'm hoping our CI servers are better able to do that part than I am!

vhscampos (approved):

PR #905, opened 2026-06-29
[ATfL][build.sh] Return 2 if XML result checking fails.

amilendra:
In the event any of the stage commands (build/config/install/package) fail, 'set -o errexit/pipefail' will make build.sh immediately exit 1.

Because of the `--ignore-fail` LIT_OPTS flag, test commands will return zero on failure, but the XML result checking code will make build.sh exit 2.

If everything succeeds build.sh exits 0.

### DIFF
--- a/arm-software/linux/build.sh
+++ b/arm-software/linux/build.sh
@@ -657,7 +657,7 @@ main() {
     done
     echo_bold "Executed build stages."
     if ! check_lit_xml_results; then
-	test_status=2
+        test_status=2
     fi
     echo_bold "Packaging...."
     package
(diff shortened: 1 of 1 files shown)


arm-software/linux/build.sh
```diff
@@ -745,7 +754,9 @@ make_and_clean_directory "${LOGS_DIR:?}"
 export LIT_OPTS="${LIT_OPTS:+${LIT_OPTS} }--ignore-fail"
 
 main
+main_status="${BUILD_SH_STATUS}"
```
pawosm-arm: why this extra variable is needed?

arm-software/linux/build.sh
```diff
@@ -653,10 +656,16 @@ main() {
         echo_bold "Completed stage: ${stage}."
     done
     echo_bold "Executed build stages."
-    check_lit_xml_results
+    if ! check_lit_xml_results; then
+	test_status=2
```
pawosm-arm: nit: indent

arm-software/linux/build.sh
amilendra: Thanks. Not needed. Removed.

arm-software/linux/build.sh
amilendra: Fixed.

pawosm-arm (approved):

PR #906, opened 2026-06-29
[ATfL][build.sh] Have the clang and flang binaries bolted

pawosm-arm:
This feature is turned OFF by default. Some preparations on the CI side may be required.

### DIFF
--- a/arm-software/linux/build.sh
+++ b/arm-software/linux/build.sh
@@ -51,6 +51,7 @@ fi
 OS_NAME=${OS_NAME:-"linux"}
 TAR_NAME=${TAR_NAME:-"atfl-${ATFL_VERSION}-${OS_NAME}-$(uname -m).tar.gz"}
 ATFL_ASSERTIONS=${ATFL_ASSERTIONS:-"ON"}
+ATFL_BOLTED=${ATFL_BOLTED:-"OFF"}
 ATFL_TARGET_TRIPLE=${ATFL_TARGET_TRIPLE:-"$(uname -m)-unknown-linux-gnu"}
 ARM_TOOLCHAIN_ID=$(cmake -DLLVM_TOOLCHAIN_PROJECT_CODE=L -P "${SOURCES_DIR}"/arm-software/shared/cmake/generate_toolchain_id.cmake)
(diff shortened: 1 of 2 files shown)


arm-software/linux/build.sh
```diff
@@ -63,6 +64,7 @@ STAGES=(
     "static_libomp_build"
 )
 ZLIB_STATIC_PATH=${ZLIB_STATIC_PATH:-"/usr/lib/$(uname -m)-linux-gnu/libz.a"}
+RELOCS_LINKER_FLAGS="-Wl,--emit-relocs,-znow"
```
amilendra, line 67: Shouldn't you change this flag depending on if `ATFL_BOLTED` is `ON` or `OFF`?

arm-software/linux/build.sh
```diff
@@ -416,7 +420,7 @@ libcpp_build() {
       print_forced_cmake_flags_cache "LIBUNWIND_NOSHARED_CMAKE_FLAGS"
     } > flags.cmake
```
amilendra, line 421: Is there a reason this flags.cmake file is not created in a `cmake_caches` directory?

arm-software/linux/build.sh
pawosm-arm: No, I want the end user to be able to use the libc++ library (where this variable is also being used) for their own optimizations, for this we need to always include those relocs. Also, if we plan to reorganize the way we build ATfL (reusing the previous builds) it may be much needed to always have those relocs in the C++ library.

arm-software/linux/build.sh
pawosm-arm: (I just watned to keep those `.cmake` files together in one place.) EDIT: now I can see what you mean. Well, I need to unify this.

pawosm-arm: I've unified the creation and use of `flangs.cmake`, no point in changing it in this commit.

pawosm-arm:
The way `check-` targets are being invoked has changed, as we'd agreed.

amilendra (approved):
LGTM.

PR #909, opened 2026-07-02
[ATfE] Remove some variants from the xfail of picolibc_sys_seek

vhscampos:
These tests started passing on these variants because we updated the FVPs to the latest version at the time of writing (11.31).


simpal01:
Can you mention the reason in the PR description ? In this case, the issue was resolved by updating to the latest FVP version.

simpal01 (approved):

Issue #910, opened 2026-07-06
[Downstream Change] PAC-RET hardening against the PACMAN attack


vhscampos:
**What is the change**
This change introduces the `-mharden-pac-ret=load-return-address` option to enable code generation of the mitigation against the PACMAN attack via load of return address. The PACMAN attack tries to extract secrets via side-channel exploits that leverage Pointer Authentication (PAC).

The change adds a new command-line option, `-mharden-pac-ret`, and a corresponding function attribute with the same values.

**Why this change cannot be done upstream**
Upstream is still reviewing the patches, but we need the changes in our main branch so that the feature is present in our releases until upstream makes a decision about the patches.

List of upstream PRs:
  - https://github.com/llvm/llvm-project/pull/176171
  - https://github.com/llvm/llvm-project/pull/176187

PR #911, opened 2026-07-06
[ATfE] PAC-RET hardening against the PACMAN attack

vhscampos:
Downstream issue:#910

The changes are still in upstream review, but we need the changes now for the Arm Toolchain for Embedded 23 release.

They include a new command-line option and a new function attribute. Their naming is still not decided in the upstream review, therefore it can change in future versions of the toolchain.

List of upstream PRs:
 - https://github.com/llvm/llvm-project/pull/176171
 - https://github.com/llvm/llvm-project/pull/176187


vhscampos:
The same feature was a downstream patch for ATfE 22. This version here has a few bug fixes in the clang side, but the core functionality is the same still.

clang/lib/CodeGen/TargetInfo.cpp
```diff
@@ -216,16 +216,35 @@ void TargetCodeGenInfo::setBranchProtectionFnAttributes(
   // Called on already created and initialized function where attributes already
   // set from command line attributes but some might need to be removed as the
   // actual BPI is different.
+// Begin downstream change #910
+
+// End downstream change #910
```
statham-arm: Possibly a carefully commented downstream change adding a single blank line is not adding enough value to be worth even a small risk of a future merge conflict 🙂

clang/lib/Driver/ToolChains/Clang.cpp
```diff
@@ -1399,84 +1399,189 @@ static void CollectARMPACBTIOptions(const ToolChain &TC, const ArgList &Args,
...
+      if (!isAArch64 && PBP.Key == "b_key")
+        D.Diag(diag::warn_unsupported_branch_protection)
+            << "b-key" << A->getAsString(Args);
+      Scope = PBP.Scope;
+      Key = PBP.Key;
+      BranchProtectionPAuthLR = PBP.BranchProtectionPAuthLR;
+      IndirectBranches = PBP.BranchTargetEnforcement;
+      GuardedControlStack = PBP.GuardedControlStack;
+// End downstream change #910
+// Begin downstream change #910
```
statham-arm: "End downstream change" followed immediately by "begin the same downstream change again" seems pointless!

llvm/lib/Target/AArch64/AArch64PointerAuth.cpp
```diff
@@ -517,6 +540,86 @@ bool AArch64PointerAuthImpl::run(MachineFunction &MF) {
     Modified = true;
   }
 
+// Begin downstream change #910
+  Modified |= emitSignReturnAddressHardening(MF);
+
+  return Modified;
+}
```
statham-arm, line 547: Wouldn't it be better to separately comment the downstream change inside `run()` and the one that adds an entire extra function, rather than commenting it as a single change that closes run() and adds all but the last two lines of the new function? That way, if there is a conflict with any future change, the comments will mark the semantics of the change and make it easier to figure out what to do.

It's confusing when automated diff tools do this, but they don't know any better. Humans do know better, so we should do better!

statham-arm (dismissed):
Approved on the basis that this looks like a faithful replication of the upstream PRs, and I had no further comments about those. But I have a few nits about the downstream-change markers (which makes sense, since those are the only new part!)

vhscampos:
@statham-arm Can you please take a final look and approve it again? I pushed the changes after your review, but then GitHub "dismissed your stale review".

statham-arm (approved):

PR #923, opened 2026-07-14
[ATFE] Improve multilib common header optimisation

simpal01:
The script currently seeds the common header set by comparing only the first two include directories
returned by os.listdir(). That makes the result sensitive to directory order and can under-detect common headers when one of those early directories is a sparse overlay like minor_libs.

This PR improves multilib header optimisation for Arm embedded toolchain builds by updating common header generation to select shared headers by relative path and content hash across primary multilib variants, while treating reduced minor variants separately so they do not shrink the common header set.

The common header generation script now:
  - groups headers by relative path and SHA-256 content hash
  - selects common headers from primary `stdlibs` variants only
  - treats minor/reduced variants separately so missing headers there do not reduce the shared header set
  - keeps variant-specific copies when header contents differ from the selected common version
  - validates that each variant has a `Group` entry in `multilib.yaml

The CMake integration now drives common-header generation through a stamp-file-backed custom command, making the generated output dependency-tracked.

### DIFF
--- a/arm-software/embedded/arm-multilib/CMakeLists.txt
+++ b/arm-software/embedded/arm-multilib/CMakeLists.txt
@@ -446,12 +446,22 @@ endif()
 
 # Extract common header files which spans across multilib variant directories.
 if(ENABLE_MULTILIB_HEADER_OPTIMISATION)
-    add_custom_target(generate-multilib-common-headers ALL
+    set(multilib_common_headers_stamp
+        "${CMAKE_CURRENT_BINARY_DIR}/multilib-optimised/.common-headers.stamp"
+    )
+
+    add_custom_command(
+        OUTPUT "${multilib_common_headers_stamp}"
         COMMAND ${Python3_EXECUTABLE} "${CMAKE_CURRENT_SOURCE_DIR}/common-headers-generate.py"
                 "${CMAKE_CURRENT_BINARY_DIR}/multilib"
                 "${CMAKE_CURRENT_BINARY_DIR}/multilib-optimised"
+        COMMAND ${CMAKE_COMMAND} -E touch "${multilib_common_headers_stamp}"
         COMMENT "Generating common headers"
         DEPENDS ${all_runtime_targets} multilib-yaml
+    )
+
+    add_custom_target(generate-multilib-common-headers ALL
+        DEPENDS "${multilib_common_headers_stamp}"
     )
 
     install(
--- a/arm-software/embedded/arm-multilib/common-headers-generate.py
+++ b/arm-software/embedded/arm-multilib/common-headers-generate.py
@@ -5,8 +5,9 @@ Identifies and extracts header files that are common across multiple multilib va
 
 This script scans all variant folders within a multilib target directory
 (e.g., `arm-none-eabi/variant1/include`, `arm-none-eabi/variant2/include`, etc.) and compares all
-`.h` files. If the same file (by name and content) appears in multiple variants, it will be moved
-to a shared include directory at:
+headers by relative path and content. Minor variants with reduced header sets are skipped during
+common header discovery, so they do not reduce the common header set. If a header path has matching
+content in a majority of non-minor variants, that content group is moved to a shared include directory at:
 
     <CMAKE_BINARY_DIR>/multilib-optimised/<target>/include/
 
@@ -21,20 +22,28 @@ Arguments:
     eg: build/multilib-builds/multilib/picolibc-build/multilib-optimised
 
 This is useful to reduce duplication in the toolchain by centralising common headers
-that are shared across architecture variants.
+that are shared across architecture variants. Variants with different content retain a local copy,
+which is found first because the variant include directory is searched before the target include
+directory. Minor variants are checked against the selected common headers in a second pass and only
+keep local copies for headers that are missing from, or differ from, the common include directory.
 """
 
 import argparse
+import hashlib
 import os
-import filecmp
 import shutil
 
 # Define the multilib target dirs which want to process
 MULTILIB_TARGET_DIRS = ["arm-none-eabi", "aarch64-none-elf"]
+PRIMARY_VARIANT_GROUP = "stdlibs"
 
 
-def files_are_identical(f1, f2):
-    return filecmp.cmp(f1, f2, shallow=False)
+def file_content_hash(path):
+    content_hash = hashlib.sha256()
+    with open(path, "rb") as file:
+        for chunk in iter(lambda: file.read(1024 * 1024), b""):
+            content_hash.update(chunk)
+    return content_hash.hexdigest()
 
 
 def collect_variant_include_paths(input_target_dir):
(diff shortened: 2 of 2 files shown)


arm-software/embedded/arm-multilib/CMakeLists.txt
```diff
@@ -446,14 +446,24 @@ endif()
...
+    )
+
+    add_custom_command(
+        OUTPUT "${multilib_common_headers_stamp}"
         COMMAND ${Python3_EXECUTABLE} "${CMAKE_CURRENT_SOURCE_DIR}/common-headers-generate.py"
                 "${CMAKE_CURRENT_BINARY_DIR}/multilib"
                 "${CMAKE_CURRENT_BINARY_DIR}/multilib-optimised"
+        COMMAND ${CMAKE_COMMAND} -E touch "${multilib_common_headers_stamp}"
         COMMENT "Generating common headers"
         DEPENDS ${all_runtime_targets} multilib-yaml
```
voltur01, line 460: What problem is this stamp solving and does it need to depend on the Python script itself? also the YAML file with library variants classification.

arm-software/embedded/arm-multilib/common-headers-generate.py
```diff
@@ -21,20 +22,28 @@
...
 MULTILIB_TARGET_DIRS = ["arm-none-eabi", "aarch64-none-elf"]
+PRIMARY_VARIANT_GROUP = "stdlibs"
 
 
-def files_are_identical(f1, f2):
-    return filecmp.cmp(f1, f2, shallow=False)
+def file_content_hash(path):
+    content_hash = hashlib.sha256()
+    with open(path, "rb") as file:
+        for chunk in iter(lambda: file.read(1024 * 1024), b""):
```
voltur01, line 44: Minor: You may want to introduce a named constant like HASH_CHUNK_SIZE or such.

arm-software/embedded/arm-multilib/common-headers-generate.py
```diff
@@ -44,12 +53,202 @@ def collect_variant_include_paths(input_target_dir):
...
+                content_hash = file_content_hash(variant_header_path)
+                # Create or update the nested entry for each header, then record
+                # which variant provides this exact header content.
+                headers.setdefault(header_name, {}).setdefault(content_hash, []).append(
+                    (variant, variant_header_path)
+                )
+    return headers
+
+
+def parse_yaml_scalar(value):
```
voltur01, line 101: Minor: Maybe strip instead of parse?

arm-software/embedded/arm-multilib/common-headers-generate.py
```diff
@@ -44,12 +53,202 @@ def collect_variant_include_paths(input_target_dir):
...
+    return value.strip().strip("\"'")
+
+
+def collect_variant_groups(multilib_yaml):
+    # Navigate multilib.yaml, only paying attention to the Variants: section
+    # and skipping other sections. Inside Variants:, remember each Dir until
```
voltur01, line 107: Nit: Colon is confusing here, at first it reads as an expanasion/expansion follows. For me, either "Variants:" with quotes or Variants without colon would be more clear.

arm-software/embedded/arm-multilib/common-headers-generate.py
```diff
@@ -44,12 +53,202 @@ def collect_variant_include_paths(input_target_dir):
...
+    # its matching Group is found, then record "target/variant" -> "group_name".
+    variant_groups = {}
+    current_dir = None
+    in_variants = False
+
+    with open(multilib_yaml, encoding="utf-8") as file:
```
voltur01, line 113: Does not LLVM libc build script already require a Python YAML parser module? Maybe we can also use it here, if it is easier/cleaner?

arm-software/embedded/arm-multilib/common-headers-generate.py
```diff
@@ -44,12 +53,202 @@ def collect_variant_include_paths(input_target_dir):
...
+
+
+def parse_yaml_scalar(value):
+    return value.strip().strip("\"'")
+
+
+def collect_variant_groups(multilib_yaml):
+    # Navigate multilib.yaml, only paying attention to the Variants: section
+    # and skipping other sections. Inside Variants:, remember each Dir until
+    # its matching Group is found, then record "target/variant" -> "group_name".
```
voltur01, line 108: Would it be useful to add a snippet of the YAML file here as a comment to see the structure being parsed?

arm-software/embedded/arm-multilib/common-headers-generate.py
```diff
@@ -44,12 +53,202 @@ def collect_variant_include_paths(input_target_dir):
...
+    variant_groups = {}
+    current_dir = None
+    in_variants = False
+
+    with open(multilib_yaml, encoding="utf-8") as file:
+        for line in file:
```
voltur01, line 114: This loop seems to be a state machine and should be correct as far as I can tell, however may be a bit tricky to reason about. For me it maybe easier to have a helper like `skip_until(tag)` so that the logic could be similar to:

before the loop skip_until('Variants')
loop:
    skip_until('Dir') -> remember the path
    skip_until('Group') -> add the record to the map

BTW, there may be a safety mechanism added (additional parameter to the helper) when looking for the Group to make sure there is no another Dir on the way to the next Group.

arm-software/embedded/arm-multilib/common-headers-generate.py
```diff
@@ -44,12 +53,202 @@ def collect_variant_include_paths(input_target_dir):
...
+            if line[0] not in (" ", "-") and stripped.endswith(":"):
+                in_variants = stripped == "Variants:"
+                current_dir = None
+                continue
+
+            if not in_variants:
+                continue
+
+            if stripped.startswith("- Dir:"):
+                current_dir = parse_yaml_scalar(stripped.split(":", 1)[1])
```
voltur01, line 128: This can be another helper like `get_attribute_value(line)`

arm-software/embedded/arm-multilib/common-headers-generate.py
```diff
@@ -44,12 +53,202 @@ def collect_variant_include_paths(input_target_dir):
...
+            # Any primary header content hash that was not selected as common
+            # must remain local to the variants that use that header content.
+            for variant, variant_header_path in entries:
+                variant_include_dir = os.path.join(
+                    output_target_dir, variant, "include"
+                )
+                copy_header(variant_header_path, variant_include_dir, header_name)
+
+    minor_headers = group_headers_by_name_and_content_hash(minor_includes)
+    for header_name in sorted(minor_headers):
```
voltur01, line 237: This whole block looks very similar to the non-common headers above - is not it possible to make a helper to call there and then here?

arm-software/embedded/arm-multilib/common-headers-generate.py
```diff
@@ -44,12 +53,202 @@ def collect_variant_include_paths(input_target_dir):
...
+    for header_name in sorted(headers):
+        content_hash_groups = headers[header_name]
+        common_group = select_common_header_group(
+            content_hash_groups, primary_variant_count
+        )
+        common_content_hash = None
+
+        if common_group:
+            common_content_hash, common_entries = common_group
+            copy_header(common_entries[0][1], output_include_dir, header_name)
```
voltur01, line 222: It would be useful to either add add a type annotation or a comment illustrating the structure of these maps to have a better idea what this refers to.
Or maybe a named helper function again with the structure explanation inside it.

voltur01 (commented):
Thank you for updating the script! As far as I can see, it should do the job, however I added a few ideas how to make the code easier to understand at least for me.

arm-software/embedded/arm-multilib/CMakeLists.txt
simpal01: The common-header stamp logic has been removed. I don’t remember exactly why it was added in the first place, so I’m removing those changes now.

arm-software/embedded/arm-multilib/common-headers-generate.py

arm-software/embedded/arm-multilib/common-headers-generate.py
simpal01: This code is no more existing now.

arm-software/embedded/arm-multilib/common-headers-generate.py
simpal01: Done

arm-software/embedded/arm-multilib/common-headers-generate.py
simpal01: Done

arm-software/embedded/arm-multilib/common-headers-generate.py
simpal01: This code is no more existing. It is simplified now

arm-software/embedded/arm-multilib/common-headers-generate.py
simpal01: This code is no more existing. It is simplified now

arm-software/embedded/arm-multilib/common-headers-generate.py
simpal01: The existing helper is called separately for primary and minor variants because they are processed in
 two distinct passes. I'll keep the current structure and add a comment explaining why.

arm-software/embedded/arm-multilib/common-headers-generate.py
simpal01: Done

voltur01 (approved):
Thank you for addressing the comments! This version looks good to me.

PR #926, opened 2026-07-14
[ATfL] Update Github Action flows for RHEL10 and Ubuntu 24.04

vrukesh:
Update Github Action workflows to build and test using Docker images for RHEL10 and Ubuntu 24.04.

Add supported files needed for building Docker images.

For both operating systems:
* Build and test using Docker images.
* Save XML test reports as artifacts.


pawosm-arm:
nit: the github `files changed` signals missing EOL in some of the files (with the red 'no entry' sign)
EDIT: solved.

pawosm-arm:
NB: I assume this really needs to be so complicated...

arm-software/linux/scripts/Docker_Files/rhel10-almalinux-10-1.repo
```diff
@@ -0,0 +1,335 @@
...
+[appstream-source]
+name=AlmaLinux 10 - AppStream - Source
+mirrorlist=https://mirrors.almalinux.org/mirrorlist/10/appstream-source
+# baseurl=https://repo.almalinux.org/vault/9/AppStream/Source/
+enabled=0
+gpgcheck=1
+gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-AlmaLinux-10
+metadata_expire=86400
+enabled_metadata=0
+priority=10
```
pawosm-arm: EOL missing

arm-software/linux/scripts/Docker_Files/install_entrypoint.sh
```diff
@@ -0,0 +1,14 @@
...
+# So rely on 'id'.
+CURRENT_USER_NAME="$(id -un)"
+
+# Source the script
+if [ -f "/home/${CURRENT_USER_NAME}/spack/share/spack/setup-env.sh" ]; then
+    source /home/${CURRENT_USER_NAME}/spack/share/spack/setup-env.sh
+fi
+
+# Execute the main command
+exec "$@"
```
pawosm-arm: EOL missing

pawosm-arm (approved):

pawosm-arm:
So much changes, so much responsibility. Do we have any plan for keeping these things up to date?

arm-software/linux/scripts/Docker_Files/install_entrypoint.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+set -e  # Exit immediately if a command exits with a non-zero status
+
+# Docker environment does not set the USER environment variable.
+# So rely on 'id'.
+CURRENT_USER_NAME="$(id -un)"
+
+# Source the script
```
pawosm-arm: We don't really use Spack here, these lines are not needed

pawosm-arm (dismissed):

arm-software/linux/scripts/Docker_Files/Dockerfile_rhel10
```diff
@@ -0,0 +1,121 @@
...
+
+RUN dnf install -y dnf-plugins-core && \
+    dnf config-manager --set-enabled crb && \
+    dnf update -y && \
+    dnf install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-10.noarch.rpm && \
+    dnf install -y --allowerasing --nobest \
+    binutils-devel \
+    cpio \
+    diffutils \
+    graphviz \
```
amilendra: What do we need graphviz for?

amilendra: I see a big difference between the dependencies in RHEL-10 and Ubuntu-24.04. Are the extra dependencies really necessary?

arm-software/linux/scripts/Docker_Files/Dockerfile_ubuntu2404
```diff
@@ -0,0 +1,92 @@
+FROM ubuntu:24.04
+
+ENV DEBIAN_FRONTEND=noninteractive
+RUN apt-get update && apt-get install -y ca-certificates curl
```
amilendra: Why do you need apt-install here?
Also I see ca-certificates being installed at line 29 as well.
Do apt-install in one place?

amilendra: Or if it is really needed here, please remove the duplicate `ca-certificates` in line 29

arm-software/linux/scripts/Docker_Files/Dockerfile_rhel10
pawosm-arm: I assume these were copied over from our guidance page (listing what is needed to build ATfL). The list on that page grew incrementally as we were discovering what is missing in the rpm-based systems which don't have Ubuntu-alike metapackages which could install them all for us.

NB. this list includes some extra tools to handle rpm files (cpio and alike), and since we decided to create a tarball here (instead of an rpm package), they indeed may not be needed (still, installing then won't hurt).

arm-software/linux/scripts/Docker_Files/Dockerfile_ubuntu2404
```diff
@@ -0,0 +1,92 @@
...
+
+# Install cmake from precompiled-binaries: https://cmake.org/download/
+RUN version=4.3.4 \
+    && mkdir /tmp/temp \
+    && cd /tmp/temp \
+    && wget https://github.com/Kitware/CMake/releases/download/v$version/cmake-$version-linux-aarch64.tar.gz \
+    && wget https://github.com/Kitware/CMake/releases/download/v$version/cmake-$version-SHA-256.txt \
+    && sed -i '/.* cmake-'"$version"'-linux-aarch64.tar.gz/!d' cmake-$version-SHA-256.txt \
+    && sha256sum -c cmake-$version-SHA-256.txt \
+    && curl -OL https://github.com/Kitware/CMake/releases/download/v$version/cmake-$version-SHA-256.txt \
```
amilendra: `curl` is installed just for this line. Rest of the downloads use `wget`
If you replace this line with `wget https://github.com/Kitware/CMake/releases/download/v$version/cmake-$version-SHA-256.txt -O cmake-$version-SHA-256.txt`, then you can remove the dependency on `curl`.

arm-software/linux/scripts/Docker_Files/Dockerfile_rhel10
pawosm-arm: > What do we need graphviz for?

LLVM uses the .dot format for representing graphs (in documentation generated during the build, and in analysis tools), it uses graphviz to handle the .dot format files.

arm-software/linux/scripts/Docker_Files/install_entrypoint.sh
amilendra: Actually I don't see even the install_entrypoint.sh file being used at all. It can be removed completely.

.github/workflows/atfl_nightly_build_and_test.yml
```diff
@@ -47,34 +59,64 @@ jobs:
...
-      - name: Build (Unix)
+      - name: Build Docker image
+        shell: bash
+        run: |
+          set -euo pipefail
+
+          docker build \
+            --file "arm-software/linux/scripts/Docker_Files/${{ matrix.dockerfile }}" \
+            --tag "${{ matrix.image_name }}" \
+            --build-arg USER_ID="$(id -u)" \
```
amilendra, line 76: The dockerfiles allow you to override `USER_ID`, `GROUP_ID` and `USER`.
Here you only override `USER_ID`, `GROUP_ID` while keeping the default `USER (cicd)`.
Is that intentional?
What happens when you override the `USER` as well? i.e. `--build-arg USER="$(id -un)"`

arm-software/linux/scripts/Docker_Files/Dockerfile_ubuntu2404
amilendra: Actually `curl` is downloading the same file previously downloaded using `wget` because it got edited by `sed`.
If you look at how RHEL-10 is doing the sha256sum check, you don't have to do that.

arm-software/linux/scripts/Docker_Files/Dockerfile_rhel10
```diff
@@ -0,0 +1,121 @@
...
+WORKDIR ${WORKSPACE}/python
+
+RUN python3.12 -m pip install --no-cache-dir poetry \
+    && poetry config virtualenvs.in-project true \
+    && poetry install --no-root --no-interaction --no-ansi
+
+ENV PATH="${WORKSPACE}/python/.venv/bin:$PATH"
+ENV PYTHON_HOME=${WORKSPACE}/python/.venv
+
+# On some of our machines, the host configuration created issues with the permissions of the filesystem
```
amilendra: Is this an issue specific to the downstream RHEL-10 hosts? Or does this apply to the github RHEL-10 hosts as well?

.github/workflows/atfl_nightly_build_and_test.yml
```diff
@@ -34,11 +40,17 @@ jobs:
       fail-fast: false  # Prevents one job failure from cancelling all
       matrix:
         include:
-          # Complete Toolchain Build & Test
+          # Ubuntu Build & Test
           - build_script: build.sh
-            install_script: install_dependencies_ubuntu.sh
```
amilendra, line 39: Should `install_dependencies_ubuntu.sh` be removed because it will be unused after this patch is merged?

.github/workflows/atfl_nightly_build_and_test.yml
```diff
@@ -34,11 +40,17 @@ jobs:
...
             target_os: Ubuntu-24.04-arm64
-            runner: ah-ubuntu_24_04-c7g_8x-100
+            dockerfile: Dockerfile_ubuntu2404
+            image_name: ubuntu2404_docker
+
+          # RHEL10 Build & Test
+          - build_script: build.sh
```
amilendra: Is there a point in having the `build_script` entry?
If RHEL-10 and Ubuntu-24 had different build scripts then I could understand the need for it, but both RHEL-10 and Ubuntu-24 have the same script `build.sh`.
So you can remove the `build_script` entry and use `./src/arm-software/linux/build.sh` instead of `./src/arm-software/linux/${{ matrix.build_script }}`.

arm-software/linux/scripts/Docker_Files/Dockerfile_rhel10
vrukesh: Resolving this conversation, based on the comments.

arm-software/linux/scripts/Docker_Files/Dockerfile_ubuntu2404
vrukesh: Updated in latest commit. Please check.

arm-software/linux/scripts/Docker_Files/Dockerfile_ubuntu2404
vrukesh: Updated in latest commit. Please check.

.github/workflows/atfl_nightly_build_and_test.yml
vrukesh: Updated in latest commit. Please check.

.github/workflows/atfl_nightly_build_and_test.yml
vrukesh: Updated in latest commit. Please check.

.github/workflows/atfl_nightly_build_and_test.yml
vrukesh: Updated in latest commit. Please check.

arm-software/linux/scripts/Docker_Files/install_entrypoint.sh
vrukesh: Updated in latest commit. Kindly check.

arm-software/linux/scripts/Docker_Files/Dockerfile_rhel10
vrukesh: Updated in latest commit. Kindly check.

.github/workflows/atfl_nightly_build_and_test.yml
```diff
@@ -47,34 +57,65 @@ jobs:
...
+
+          docker run --rm \
+            --volume "$GITHUB_WORKSPACE:/workspace/src" \
+            --workdir /workspace \
+            "${{ matrix.image_name }}" \
+            bash -lc '
+              set -euo pipefail
+              git config --global --add safe.directory /workspace/src
+
+              if [[ "${{ matrix.target_os }}" == RHEL10-* ]]; then
```
amilendra: This conditional is going to be a headache to maintain if/when we add more hosts.
A better approach would be to add the zlib_path to the matrix, and pass that as an environment variable to the docker run command. That in my opinion would make this easy to maintain and extend.

e.g.

```
      matrix:
        include:
          # Ubuntu Build & Test
          - target_os: Ubuntu-24.04-arm64
            dockerfile: Dockerfile_ubuntu2404
            image_name: ubuntu2404_docker
            zlib_path: /usr/lib/aarch64-linux-gnu/libz.a

          # RHEL10 Build & Test
          - target_os: RHEL10-arm64
            dockerfile: Dockerfile_rhel10
            image_name: rhel10_docker
            zlib_path: /usr/lib64/libz.a
```

```
          docker run --rm \
            --volume "$GITHUB_WORKSPACE:/workspace/src" \
            --workdir /workspace \
            -e ZLIB_STATIC_PATH="${{ matrix.zlib_path }}"
            "${{ matrix.image_name }}" \
            bash -lc '
              set -euo pipefail
              git config --global --add safe.directory /workspace/src
              ./src/arm-software/linux/build.sh
            ' 2>&1 | tee build-${{ matrix.target_os }}.log
```
If `set -euo pipefail` and `git config --global --add safe.directory /workspace/src` can be removed, we can further simplify the docker run as follows
```
          docker run --rm \
            --volume "$GITHUB_WORKSPACE:/workspace/src" \
            --workdir /workspace \
            -e ZLIB_STATIC_PATH="${{ matrix.zlib_path }}"
            "${{ matrix.image_name }}" \
            ./src/arm-software/linux/build.sh
            ' 2>&1 | tee build-${{ matrix.target_os }}.log
```

.github/workflows/atfl_nightly_build_and_test.yml
```diff
@@ -47,34 +57,65 @@ jobs:
...
+        shell: bash
+        run: |
+          set -euo pipefail
+
+          docker run --rm \
+            --volume "$GITHUB_WORKSPACE:/workspace/src" \
+            --workdir /workspace \
+            "${{ matrix.image_name }}" \
+            bash -lc '
+              set -euo pipefail
```
amilendra: Are these two (set pipefail and git config) really needed? You are doing a `set pipefail` before the `docker run` too. What happens when you remove these two lines?

arm-software/linux/scripts/Docker_Files/Dockerfile_rhel10
```diff
@@ -0,0 +1,109 @@
...
+    && alternatives \
+        --install /usr/local/bin/cmake cmake /opt/cmake-$version-linux-aarch64/bin/cmake 10 \
+        --slave /usr/local/bin/ctest ctest /opt/cmake-$version-linux-aarch64/bin/ctest \
+        --slave /usr/local/bin/cpack cpack /opt/cmake-$version-linux-aarch64/bin/cpack \
+        --slave /usr/local/bin/ccmake ccmake /opt/cmake-$version-linux-aarch64/bin/ccmake \
+        --family cmake \
+    && cmake --version \
+    && ctest --version \
+    && cpack --version
+
```
amilendra, line 76: Please add ``` \
    && ccmake --version```. It is there in the Ubuntu docker file.

amilendra (changes requested):

.github/workflows/atfl_nightly_build_and_test.yml
vrukesh: Updated in latest version. Kindly check.

arm-software/linux/scripts/Docker_Files/Dockerfile_rhel10
vrukesh: Updated in latest version. Please check.

.github/workflows/atfl_nightly_build_and_test.yml
vrukesh: Updated in latest version. Please check.

arm-software/linux/scripts/Docker_Files/Dockerfile_rhel10
```diff
@@ -0,0 +1,107 @@
...
+RUN sed -i '/^\[.*\]/a priority=1' /etc/yum.repos.d/ubi.repo
+
+RUN dnf install -y dnf-plugins-core && \
+    dnf config-manager --set-enabled crb && \
+    dnf install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-10.noarch.rpm && \
+    dnf install -y --allowerasing --nobest \
+    binutils-devel \
+    diffutils \
+    file \
```
amilendra: Building locally works fine without `file`

arm-software/linux/scripts/Docker_Files/Dockerfile_rhel10
```diff
@@ -0,0 +1,107 @@
...
+    gnupg2 \
+    glibc-common \
+    libzstd-devel \
+    libzstd-static \
+    gcc \
+    gcc-c++ \
+    git \
+    make \
```
amilendra: Also works without `make`.

arm-software/linux/scripts/Docker_Files/Dockerfile_rhel10
vrukesh: Fixed in latest version. Build is passing !

arm-software/linux/scripts/Docker_Files/Dockerfile_rhel10
vrukesh: Fixed in latest version. Build is passing !

amilendra (approved):

vrukesh:
> NB: I assume this really needs to be so complicated...

Earlier packaging was planned too. Now, lot of packages are removed. So, flows should be simpler.

vrukesh:
> nit: the github `files changed` signals missing EOL in some of the files (with the red 'no entry' sign) EDIT: solved.

Earlier packaging was planned too. Now, lot of packages are removed. So, flows should be simpler.

arm-software/linux/scripts/Docker_Files/install_entrypoint.sh
amilendra: Changes requested have been carried out.

PR #928, opened 2026-07-15
Rename release workflows and correct commands

dcandler:
The `[` character is sorted after alphabetical ones, so use `(` instead to try to keep the workflows at the top of the list.

Additionally, add missing double quotes to the run commands and increase the fetch depth to fetch all history in order to find the branch points.

### DIFF
--- a/.github/workflows/create_arm_release_branch.yml
+++ b/.github/workflows/create_arm_release_branch.yml
@@ -7,7 +7,7 @@
 # The branch should be created at the commit where the upstream branch
 # point was merged on to the arm-software branch.
 
-name: "[arm-toolchain] Create arm-toolchain Release Branch"
+name: "(arm-toolchain) Create arm-toolchain Release Branch"
 on:
   workflow_dispatch:
     inputs:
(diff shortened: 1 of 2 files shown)


pawosm-arm (dismissed):

.github/workflows/create_arm_release_branch.yml
```diff
@@ -30,6 +30,7 @@ jobs:
         uses: actions/checkout@v6
         with:
           token: ${{ steps.generate-token.outputs.token }}
+          fetch-depth: 0
```
pawosm-arm, line 33: No symmetric change in `llvm_release_branch` needed?

dcandler: No, creating the LLVM branch doesn't need the `arm-software` commit history, the LLVM main and release branches are fetched separately.

Whereas the downstream arm release branch does need the `arm-software` commit history to order find the LLVM branch point and corresponding automerge commit on the downstream branch.

pawosm-arm (approved):

PR #934, opened 2026-07-16

pawosm-arm:

As the commit introduced by #535 has resulted in a faulty C++ compiler
which could not see system-provided standard C++ library, an alternative
solution is now being cherry-picked here which appears to work with
Amazon Linux 2023.

This is an one off change for the ATfL23 release.

Downstream issue: #533

Co-authored-by: Benjamin Maxwell <benjamin.maxwell@arm.com>


pawosm-arm:
Mid-air collision in the github world...

pawosm-arm:
github has lost it completely, I'll try rebase and force push.

pawosm-arm:
Ah, so they had the branches wrong way!

PR #935, opened 2026-07-16

pawosm-arm:

As the commit introduced by #535 has resulted in a faulty C++ compiler which could not see system-provided standard C++ library, an alternative solution is now being cherry-picked here which appears to work with Amazon Linux 2023.

This is an one off change for the ATfL23 release.

Downstream issue: #533


pawosm-arm:
@MacDue I still cannot add you as a reviewer, but feel free to approve it (or anyone else).

vrukesh (approved):
LGTM

PR #937, opened 2026-07-17
[ATfL][release/23.x][docs] Changelog update

pawosm-arm:


schander-arm (approved):

arm-software/linux/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to this project will be documented in this file.
 
+## [23.1.0]
+
+- The `armclang` and `armflang` executable binaries are now optimized with BOLT.
```
amilendra, line 7: Does this mean that the clang and flang binaries in ATfL 23.1.0 will be built with `ATFL_BOLTED=ON`?

pawosm-arm: > Does this mean that the clang and flang binaries in ATfL 23.1.0 will be built with `ATFL_BOLTED=ON`?

Yes, that's the plan, and we need to make sure the CI is suitable for it before the end of next week. All the release builds that go into debs and rpms, must be done with `ATFL_BOLTED=ON`

amilendra: Thanks for the clarification

PR #941, opened 2026-07-20
[ATfL][build] Add the ability to bypass the bootstrap building with a provided bootstrap compiler

pawosm-arm:
This is to avoid the issues caused by the aging GCC versions by removing the GCC compiler from the picture.

### DIFF
--- a/arm-software/linux/build.sh
+++ b/arm-software/linux/build.sh
@@ -22,6 +22,7 @@ BUILD_DIR=${BUILD_DIR:-"${BASE_DIR}/build"}
 ATFL_DIR=${ATFL_DIR:-"${BUILD_DIR}/atfl"}
 LOGS_DIR=${LOGS_DIR:-"${BASE_DIR}/logs"}
 OUTPUT_DIR=${OUTPUT_DIR:-"${BASE_DIR}/output"}
+BOOTSTRAP_COMPILER_DIR=${BOOTSTRAP_COMPILER_DIR:-"${BUILD_DIR}/bootstrap_compiler"}
 
 #########################
 ## Configuration: Mode ##
@@ -87,8 +88,8 @@ declare -A COMMON_CMAKE_FLAGS=(
 )
 
 declare -A USE_BOOTSTRAP_CMAKE_FLAGS=(
-    ["CMAKE_C_COMPILER:FILEPATH"]="\"${BUILD_DIR}/bootstrap_compiler/bin/clang\""
-    ["CMAKE_CXX_COMPILER:FILEPATH"]="\"${BUILD_DIR}/bootstrap_compiler/bin/clang++\""
+    ["CMAKE_C_COMPILER:FILEPATH"]="\"${BOOTSTRAP_COMPILER_DIR}/bin/clang\""
+    ["CMAKE_CXX_COMPILER:FILEPATH"]="\"${BOOTSTRAP_COMPILER_DIR}/bin/clang++\""
     ["CMAKE_INSTALL_PREFIX:PATH"]="\"${ATFL_DIR}\""
     ["LLVM_ENABLE_LLD:BOOL"]=ON
 )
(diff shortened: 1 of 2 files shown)


arm-software/linux/build.sh
```diff
@@ -356,13 +359,19 @@ print_forced_cmake_flags_cache() {
...
+    echo "-fuse-ld=lld" >"${BOOTSTRAP_COMPILER_DIR}/bin/clang++.cfg"
 }
 
 bootstrap_compiler_build() {
-    mkdir -p "${BUILD_DIR}/stage/bootstrap_compiler"
-    cd "${BUILD_DIR}/stage/bootstrap_compiler"
+    if [[ -e "${BOOTSTRAP_COMPILER_DIR}/bin/clang++" ]]; then
+      echo "Using the existing bootstrap compiler."
+      export PATH="${BOOTSTRAP_COMPILER_DIR}/bin:${PATH}"
+      bootstrap_compiler_default_config
```
amilendra, line 370: I don't think it is good practice to modify the configuration file of an externally provided toolchain.  Also `USE_BOOTSTRAP_CMAKE_FLAGS` already uses the CMake setting `-DLLVM_ENABLE_LLD=ON` to ensure that the lld linker is used by the subsequent stages. So I think `bootstrap_compiler_default_config` is unnecessary and we can completely remove `bootstrap_compiler_default_config` from this script. That also will remove the requirement of the bootstrap compiler location having to be writable.

arm-software/linux/build.sh
```diff
@@ -403,7 +412,7 @@ bootstrap_compiler_build() {
         2>&1 | tee "${LOGS_DIR}/bootstrap_compiler.txt"
     run_command cmake --build . "${CMAKE_BUILD_ARGS[@]}" 2>&1 | tee -a "${LOGS_DIR}/bootstrap_compiler.txt"
     run_command cmake --install . 2>&1 | tee -a "${LOGS_DIR}/bootstrap_compiler.txt"
-    export PATH="${BUILD_DIR}/bootstrap_compiler/bin:${PATH}"
+    export PATH="${BOOTSTRAP_COMPILER_DIR}/bin:${PATH}"
     bootstrap_compiler_default_config
     run_test_command "${LOGS_DIR}/bootstrap_check_all.xml" "${LOGS_DIR}/bootstrap_compiler.txt" check-all
```
amilendra, line 418: On early return when an external bootstrap compiler is used, `"${LOGS_DIR}/bootstrap_check_all.xml"` will not be created. However `check_lit_xml_results` unconditionally expects `"${LOGS_DIR}/bootstrap_check_all.xml"` to be there and fails if it does not exist.

arm-software/linux/build.sh
pawosm-arm: Sadly, this git log does not go that far back so there's no log of the justification behind forcing to use LLD the most brutal way ever. We may first try to introduce an PR which lifts that and see if anything fails (could be the reason behind it was to workaround some upstream issue haunting the building process at that time) and if it is not failing, we would effectively get rid of it.

amilendra: Yeah that makes sense. Let's test that cleanup separately ASAP.

pawosm-arm: BTW, this bootstrap compiler directory still needs to be writable, there's no way to pass `LD_LIBRARY_PATH` to lit tests executed by `check-all` anymore (there was an long an annoying discussion why there is no going back to this), so forcing the use of Rpath in the compiler's config file is the only way to do it:

```
    echo "-Wl,-rpath=${ATFL_DIR}/lib" >> "${BUILD_DIR}"/bootstrap_compiler/bin/clang++.cfg
    run_test_command "${LOGS_DIR}/product_check_all.xml" "${LOGS_DIR}/product.txt" check-all
```

pawosm-arm: > Yeah that makes sense. Let's test that cleanup separately ASAP.

https://github.com/arm/arm-toolchain/pull/1042

arm-software/linux/build.sh
pawosm-arm: I've added a line for XML file creation, and also rebased this PR so it could be merged cleanly.

pawosm-arm: Well, not rebased to the top, as the topmost change needs to be verified first.

arm-software/linux/build.sh
pawosm-arm: Apparently, `-DLLVM_ENABLE_LLD=ON` applies only to build time, and the default linker (GNU ld) is being used when testing. And this reveals two problems in RHEL8: one is that GNU ld needs explicit `-ldl` for many of the libcxx test files, and second one is more worrying, the code linked with GNU ld fails with exceptions. And the second issue cannot be solved by adding any more flags, it just needs to be linked with lld to make it pass on RHEL8.

I'm not sure how much we want to invest in solving it. If not much, then we'd better revert it.

amilendra: Let's leave it as it is for now. If upstream approves the libc++ changes we can do the cleanup.

arm-software/linux/build.sh
amilendra: The topmost change got reverted so I think now this can be rebased and merged.

amilendra (approved):

PR #944, opened 2026-07-21
[ATFE] Remove downstream patch now that it has been upstreamed.

simpal01:
The changes previously carried as a downstream patch have now been merged upstream in picolibc:

- Upstream PR: https://github.com/picolibc/picolibc/pull/1289

Since the functionality is now part of upstream picolibc, the downstream patch is no longer needed and has started conflicting with the upstream code.

This change removes the obsolete downstream patch and relies on the upstream implementation instead.


dcandler:
With one patch removed, the remaining patches go `0001`, `0003`, then `0004`, so it would help avoid confusion if they were renumbered/rebased.

simpal01:
> With one patch removed, the remaining patches go `0001`, `0003`, then `0004`, so it would help avoid confusion if they were renumbered/rebased.

Good catch David. Renumbered.

dcandler (approved):

HugoSilvaSantos (approved):

PR #947, opened 2026-07-22
Allow ATfE get_fvps.sh to use curl as well as wget.

statham-arm:
We auto-detect via `which`, and use whichever tool we can find on PATH, with an appropriate command line in each case.

My general impression is that curl seems more 'in fashion', so this change should enable more users to run this script without having to install an extra package specially.

I've also removed the use of `wget --content-disposition` (which saves the file under a name specified by the remote web server), so that instead we specify the output file name explicitly in each download to be the same name that the rest of the script will expect. curl does have an equivalent option (`-J`), but it seems more robust to avoid depending on the web server matching this script at all.

I added the `--retry 5` option in curl because in one of my test downloads, the web server cut off the connection after a couple of Mb. In that situation wget defaults to trying again to download from where the first attempt left off, but curl doesn't unless you tell it to.


vhscampos (approved):
Nice! LGTM

PR #948, opened 2026-07-22
[ATfE] Fix picolibc FVP startup selection for Arm based variants.

simpal01:
Adjusts picolibc startup object selection so Arm FVP targets use the shared semihosting startup object, while AArch64 FVP targets continue to use crt0-semihost-fvp.

### DIFF
--- a/arm-software/embedded/arm-runtimes/CMakeLists.txt
+++ b/arm-software/embedded/arm-runtimes/CMakeLists.txt
@@ -173,14 +173,10 @@ add_custom_target(check-all)
 if(ENABLE_LIBC_TESTS OR ENABLE_COMPILER_RT_TESTS OR ENABLE_LIBCXX_TESTS)
     # Flags required to link tests.
     if(C_LIBRARY STREQUAL picolibc)
-        if(TEST_EXECUTOR STREQUAL qemu)
-            set(picocrt "crt0-semihost")
-        # For Arm, picolibc only builds one semihosting startup object because
-        # the same startup code works for both QEMU and FVP. AArch64 still uses
-        # the FVP-specific startup object.
-        elseif(TEST_EXECUTOR STREQUAL fvp AND TARGET_ARCH MATCHES "^aarch64")
+        # AArch64 uses the FVP-specific startup object.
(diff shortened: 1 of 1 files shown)


arm-software/embedded/arm-runtimes/CMakeLists.txt
```diff
@@ -175,8 +175,13 @@ if(ENABLE_LIBC_TESTS OR ENABLE_COMPILER_RT_TESTS OR ENABLE_LIBCXX_TESTS)
     if(C_LIBRARY STREQUAL picolibc)
         if(TEST_EXECUTOR STREQUAL qemu)
             set(picocrt "crt0-semihost")
-        elseif(TEST_EXECUTOR STREQUAL fvp)
+        # For Arm, picolibc only builds one semihosting startup object because
+        # the same startup code works for both QEMU and FVP. AArch64 still uses
+        # the FVP-specific startup object.
+        elseif(TEST_EXECUTOR STREQUAL fvp AND TARGET_ARCH MATCHES "^aarch64")
             set(picocrt "crt0-semihost-fvp")
```
voltur01, line 178: Is this a good opportunity to simplify: there is one special case for AArch64 FVP, the rest is default?

Otherwise it maybe cleaner to have two levels of conditions: (1) is this QEMU or FVP?, (2) inside FVP is this AArch64 or not?

dcandler: As far as I'm aware, AArch64 using FVPs is the only special case requiring something different, so I believe
```
if(TEST_EXECUTOR STREQUAL fvp AND TARGET_ARCH MATCHES "^aarch64")
    set(picocrt "crt0-semihost-fvp")
else()
    set(picocrt "crt0-semihost")
endif()
```
would be sufficient (assuming `TEST_EXECUTOR` can only be `fvp` or `qemu`).

MarkMurrayARM: Concise. I like it.

simpal01: Done

voltur01 (approved):
Thank you for the update!

PR #949, opened 2026-07-22
[ATfE] Set version number in VERSION.txt

dcandler:
The script generating the VERSION.txt used the incorrect variable name, so the version number was missing from the file.

The version can be found in the ArmToolchainForEmbedded_VERSION variable, which is autogenerated by the project command. This was correctly being passed down to the script generating the VERSION.txt file, however the file template referenced a different variable name that didn't exist.

### DIFF
--- a/arm-software/embedded/cmake/VERSION.txt.in
+++ b/arm-software/embedded/cmake/VERSION.txt.in
@@ -1,4 +1,4 @@
-Arm Toolchain for Embedded ${armtoolchain_VERSION}
+Arm Toolchain for Embedded ${ArmToolchainForEmbedded_VERSION}
 
 Sources:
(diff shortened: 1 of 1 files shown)


statham-arm (approved):
I had to search around a bit to find out _why_ this is the right variable name; maybe mention that in the commit message? As far as I can see the answer is because `ArmToolchainForEmbedded` is the name of the project for the purposes of the `project()` command in the top-level CMakeLists.txt, and cmake sets [project name]`_VERSION` in response to the `VERSION` option in `project()`.

It would be nice if cmake `configure_file()` supported some analogue of the `${var?}` syntax in sh, which bombs out with an error if the variable name doesn't exist, instead of silently continuing and substituting the empty string. But I had a quick look in the docs and didn't find one.

voltur01 (approved):

PR #951, opened 2026-07-23
[ATfE] Add xfails for a crash in llvm-libc Float128 tests.

statham-arm:
Some explanation of the root cause is in a comment I left here: https://github.com/llvm/llvm-project/pull/200565#pullrequestreview-4764426705 The immediate cause of the crash is an out-of-bounds array reference, in an array very close to the top of the stack. I think the reason it fails on this precise set of library variants is not because they're the only ones with the _bug_, but they're the only ones for which QEMU causes a data abort for accesses just off the top of the stack.


arm-software/embedded/arm-runtimes/test-support/xfails.py
```diff
@@ -929,6 +929,35 @@ def check_r52_warning():
...
+                "armv8m.main_hard_fp_exn_rtti_unaligned_size",
+                "armv8m.main_hard_fp_unaligned_size",
+                "armv8m.main_hard_fp_unaligned_size",
+                "armv8m.main_hard_fp_unaligned_size",
+                "armv8m.main_soft_nofp_exn_rtti_unaligned_size",
+                "armv8m.main_soft_nofp_exn_rtti_unaligned_size",
+                "armv8m.main_soft_nofp_exn_rtti_unaligned_size",
+                "armv8m.main_soft_nofp_unaligned_size",
+                "armv8m.main_soft_nofp_unaligned_size",
+                "armv8m.main_soft_nofp_unaligned_size",
```
dcandler, line 945: Are there actually only 6 variants affected? The list repeats each name 3 times.

statham-arm: Ha! Good catch. I pasted the list of failures out of Jenkins, and sorted it into order only at the very last moment, apparently without looking at it very hard afterwards. And Jenkins ran each test on three platforms.

dcandler (approved):

PR #953, opened 2026-07-24
[ATfE] Enable minimal libc++ testing with LLVM libc

voltur01:
This adds one AArch64 and two Armv7-M (softfp and hardfp) configurations to test libc++ on top on LLVM libc.

This requires some small changes/fixes along the way:
- JSON files enable tests and also provide enough memory for C++ tests.
- CMake and lit option to tell which C library is used for C++ tests.
- CMake disable rt library to avoid use of clock_gettime() that bare-metal LLVM libc does not provide.
- Test runners to only pass the name of the test binary - not full paths that exceeds LLVM libc bootcode limit and is not used anyway.
- Enable stderr to be used separately from stdout in LLVM semihosting and FVP parameters by using OPENMODE_A to open the special semihosting file.

The resulting failures added to the xfail list: mostly missing functionality or mismatch with hosted behavior. A few stdin tests excluded since they hang with semihosting thus trigger the long timeout.


arm-software/embedded/arm-runtimes/CMakeLists.txt
```diff
@@ -1058,6 +1058,7 @@ if(ENABLE_CXX_LIBS)
             -DLIBCXX_ENABLE_FILESYSTEM=OFF
             -DLIBCXX_ENABLE_LOCALIZATION=ON
             -DLIBCXX_ENABLE_UNICODE=OFF
+            -DLIBCXX_HAS_RT_LIB=OFF
```
vhscampos, line 1061: Can you explain in the PR description why the rt lib needs to be disabled?

arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
```diff
@@ -116,7 +116,7 @@ bool __llvm_libc_timespec_get_utc(struct timespec *ts) {
 void _platform_init(void) {
   stdio_open(&__llvm_libc_stdin_cookie, OPENMODE_R);
   stdio_open(&__llvm_libc_stdout_cookie, OPENMODE_W);
-  stdio_open(&__llvm_libc_stderr_cookie, OPENMODE_W);
+  stdio_open(&__llvm_libc_stderr_cookie, OPENMODE_A);
```
vhscampos, line 119: Why this change?

arm-software/embedded/arm-runtimes/CMakeLists.txt
voltur01: Yes, expanded: it effectively controls use of clock_gettime() that is a POSIX function provided by bare-metal LLVM libc now (it seems to be provided for some other targets though).

arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
voltur01: Expanded the description as well: this is how the stderr is opened (vs stdout).

vhscampos (dismissed):

vhscampos (approved):
LGTM

PR #957, opened 2026-07-28
[ATfL][docs] Building documentation now requires furo

pawosm-arm:


arm-software/linux/build.sh-HOWTO.md
```diff
@@ -26,6 +26,7 @@ The following packages need to be installed (using `apt`):
 - `python3-pip`
 - `python3-pygments`
 - `python3-yaml`
+- `python3-furo` (since Ubuntu 24)
```
amilendra: Does this mean that this is not needed for `Ubuntu 22` and before?

arm-software/linux/build.sh-HOWTO.md
```diff
@@ -77,6 +78,7 @@ The following packages need to be installed (using `apt`):
 - `python3-pygments`
 - `python3-pyyaml`
 - `python3-myst-parser` (since RHEL10)
+- `python3-furo` (since RHEL10)
```
amilendra: Same question for RHEL8 and before

arm-software/linux/build.sh-HOWTO.md
pawosm-arm: like in other `since` cases, these packages are not available to those older distributions, and the users would have to find other means (like pip), or stop using ancient distros (like RHEL8, RHEL9, Ubuntu22, SLES15).

pawosm-arm: Bear in mind, this document is addressing professionals, they are expected to know what to take out of this list.

arm-software/linux/build.sh-HOWTO.md
pawosm-arm: Same answer as above.

arm-software/linux/build.sh-HOWTO.md
amilendra: I don't think the `since` is needed here at all and it could be misleading. `python3-furo` is needed since https://github.com/llvm/llvm-project/pull/211993 rather than Ubuntu-24. Just remove  `(since Ubuntu 24)` and let the user figure out how to get it if they are using a host without `python3-furo` support.

pawosm-arm: Actually, the package is called `furo` on deb systems, it's called `python3-furo` on rpm systems.

pawosm-arm: > Just remove `(since Ubuntu 24)` and let the user figure out how to get it if they are using a host without `python3-furo` support.

Done.

amilendra (approved):

PR #962, opened 2026-07-30
[ATfE] Fix aarch64 target triple passed to newlib build

vhscampos:
newlib does not support the "aarch64-unknown-none-elf" target triple name that is now used by ATfE. It does not understand the "unknown" segment.

This patch transforms the target triple passed to newlib in this specific case.

The final toolchain installation has paths that still follow the ATfE target triple naming, hence no change in user visible behaviour.


simpal01:
Small nit pick in the PR description it should be "aarch64-unknown-none-elf"  instead of "aarch-unknown-none-elf" .

simpal01 (approved):

PR #963, opened 2026-07-30
[ATfL][build.sh] We cannot optimize armflang due to race conditions in PGO and failing assertions in BOLT

pawosm-arm:


amilendra (approved):

pawosm-arm:
I need to verify something before merging.
EDIT: verified, there is no hope for completing flang optimization.

Issue #446, opened 2025-07-29
[Downstream Change] Extend the Multilib system to support an IncludeDirs field via downstream patching.


simpal01:
**What is the change**
This patch extends the Multilib Yaml format to allow specifying an IncludeDirs for the header path/s per multilib variant. The goal is to enable fine-grained control over header search paths for each multilib variant. This feature is especially useful in setups where header paths deviate from the default bare-metal assumptions. For example, when headers are shared across target triples, it becomes necessary to organise them under target-triple-specific directories to ensure correct resolution.

**Why this change cannot be done upstream**
This is already in upstream review https://github.com/llvm/llvm-project/pull/146651. Since the review is pending, a downstream patch is made to enable IncludeDirs support in multilib.yaml

PR #968, opened 2026-07-31
Revert "[ATFE] Remove ATFE downstream multilib IncludeDirs patch"

simpal01:
Reverts arm/arm-toolchain#924

Downstream issue:#446.

This patch causes our package tests fail in CI. From an initial look, when clang -cc1 is invoked directly with -triple thumbv6m-unknown-none-eabi, the include lookup appears to be based on that exact triple. That means it may try to resolve paths under lib/clang-runtimes/thumbv6m-unknown-none-eabi, rather than the packaged runtime directory lib/clang-runtimes/arm-unknown-none-eabi.


pawosm-arm (approved):
Nothing that concerns the ATfL toolchain.

vhscampos (approved):

PR #978, opened 2026-08-04
[ATfE] PAC-RET hardening against the PACMAN attack

vhscampos:
Downstream issue:#977

The changes are still in upstream review, but we need the changes now for the Arm Toolchain for Embedded 23 release.

They include a new command-line option and a new function attribute. Their naming is still not decided in the upstream review, therefore it can change in future versions of the toolchain.

List of upstream PRs:
 - https://github.com/llvm/llvm-project/pull/176171
 - https://github.com/llvm/llvm-project/pull/176187


vhscampos:
@dcandler FYI

dcandler (approved):
LGTM from the ATfE side: it's the same change as is currently on trunk, only with the the downstream issue number updated.

vhscampos:
@arm/arm-toolchain-linux Can you please also take a look? I plan to merge it by tomorrow if possible.

pawosm-arm (approved):
Since behind a flag, it's not affecting ATfL's customers.

PR #979, opened 2026-08-04
[ATfL][performance] Set use-dereferenceable-at-point-semantics to false by default

pawosm-arm:
This is to revert the performance regressions observed after the
introduction of https://github.com/llvm/llvm-project/pull/204795


pawosm-arm:
Wrong patch...
EDIT: now it is the right one.

david-arm (approved):
LGTM!

PR #980, opened 2026-08-04
[ATfL][performance] Make real-sum-reassociation the default in flang

pawosm-arm:
This change is intended to the ATfL23 release only.


tblah (approved):
More information and fortran standards argument is here: https://github.com/llvm/llvm-project/pull/207371

This fixes almost all of the ACfL->ATfL regression in snbone.

LGTM

PR #982, opened 2026-08-05
[ATfL][build.sh] Even non-bolted builds should introduce non-bolted clang/flang binaries

pawosm-arm:


pawosm-arm:
This is to make the packaging validation tests pass on RHEL8.

Also, this needs to be cherry-picked to the release-23 branch.

amilendra (approved):

PR #966, opened 2026-07-30
[Downstream Change] Revert ccef34 and related commits

john-brawn-arm:
This is being done to fix a regression in an important embedded benchmark for ATfE 23.

This reverts the following commits:
ca970f3bb3ce41365dfc85049db44489c88cb0a5
a47be1995a65a411527062e7ca7e92fff8d6ad1b.
024a691e7784e7ebe3783ddaf557ee07a127d934
ccef34d45ae1ec28c129612a53f24b0e160912aa

Downstream issue: #965

PR #984, opened 2026-08-05
[Downstream Change][flang][OpenMP] Resolve private array source from block-arg owner in alias analysis

pawosm-arm:
This is a cherry-pick of PR #208227 from the LLVM's main branch.

Part of #208086. `fir::AliasAnalysis::getSource` fails to recognize an OpenMP private array as an `Allocate` source when its `hlfir.declare` is nested inside an omp.loop_nest, and conservatively returns `MayAlias`. This changes `getSource` to resolve the clause-carrying OpenMP op from the private block argument's owner region, so the private is correctly classified and does not spuriously alias other objects. Assisted-by: Copilot

Downstream issue: #988

### DIFF
--- a/flang/lib/Optimizer/Analysis/AliasAnalysis.cpp
+++ b/flang/lib/Optimizer/Analysis/AliasAnalysis.cpp
@@ -1394,10 +1394,18 @@ AliasAnalysis::getSourceImpl(mlir::Value v, bool getLastInstantiationPoint,
           // but their handling is more complex. Maybe we can find better
           // abstractions to handle them in a general fashion.
           bool isPrivateItem = false;
+          // The private/map block argument is owned by the clause-carrying op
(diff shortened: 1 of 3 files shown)


pawosm-arm:
Looking at this PR and #966 we can clearly see a conflict of (performance) interest between ATfE and ATfL. Considering we plan to introduce more cherry-picks like that, I think we need to call an emergency meeting on how can we immediately decouple ATfL and ATfE repositories to avoid problems in this and any future releases.

tblah (approved):
Low risk backport. LGTM.

This is fortran-only so it shouldn't have any impact on ATfE

pawosm-arm:
> This is fortran-only so it shouldn't have any impact on ATfE

This one yes, but we have a few more in the pipeline, and some of them go beyond that.

PR #985, opened 2026-08-05
[Downstream Change][flang][OpenMP] Avoid boxing constant-size trivial private arrays

pawosm-arm:
This is a cherry-pick of PR #208315 from the LLVM's main branch.

Part of #208086. `privatizeSymbol` (used for both OpenMP privatization and `do concurrent` locals) currently boxes every array as a workaround for `fir.array` not being directly `alloca`-able in the OpenMP→LLVMIR translation. This PR carves out the common case: a constant-shape array of trivial intrinsic elements (e.g. `real(8) :: xx(3)`) is now privatized unboxed, as a plain `fir.array`.

Downstream issue: #988


tblah (approved):
LGTM. Another low risk performance improvement from upstream flang.

This is fortran-only

pawosm-arm:
I had to rebase it.

Issue #995, opened 2026-08-10
[Downstream Change][BOLT] Enable lite mode by default on AArch64 in ATfL 23


pawosm-arm:
**What is the change**
The bolting process of flang tends to be unstable sometimes. A good advice is to use the lite mode by default, which should improve the stability. The performance cost is marginal.

**Why this change cannot be done upstream**
It's too late at this point of the release cycle.

PR #997, opened 2026-08-10
[Downstream Change][RFC][BOLT] Enable lite mode by default on AArch64 (#205545)

pawosm-arm:
Cherry-pick from commit b6b9603c7675fff5ec9befacf54391a516fc723c

This is the default on the main branch and will remain so for the next release.

Downstream issue: #995


ilinpv (approved):
Looks good to me!

dcandler (approved):

amilendra (approved):
Provided that this is going to be in LLVM-24, I don't have objections of having this in LLVM-23 as well.

Issue #1000, opened 2026-08-11
[Downstream Change] Cherry-pick significant performance improvement


amilendra:
**What is the change**
See llvm/llvm-project#214206 for the main change, which fixes the cost of vector shifts with non-uniform constant amounts. This PR gives a significant performance improvement in an important Linux benchmark.

**Why this change cannot be done upstream**
PR #214206 has already landed upstream, but it happened after the LLVM-23 branch point.

PR #1001, opened 2026-08-11
[Downstream change][AArch64] Fix cost of vector shifts with non-uniform constant amounts

pawosm-arm:
This is a cherry-pick from the upstream LLVM. See https://github.com/llvm/llvm-project/pull/214206

Non-uniform constants use variable shifts and require materializing the shift vector. Account for a shift and materialization per legalized vector.

Downstream issue: #1000


pawosm-arm:
I've opened this, just in case the cherry-pick to the upstream release branch met any hard obstacles.

bipmis (approved):
LGTM. The patch recovers gain in 2 spec2017 benchmarks and will be an important cherry pick for ATfL23.

amilendra:
@voltur01 and @stuij 
Could you please approve this patch?
We believe this patch will not affect the ATfE embedded benchmarks. However we are happy to revert in the event it causes any regressions.

dcandler (approved):

PR #1004, opened 2026-08-17
[automerge.py] Print git's error message when it fails.

statham-arm:
The default dump of `subprocess.CalledProcessError` only prints the exit status. But when git exits with status 1, it generally printed a message explaining the problem in more detail. That message currently isn't getting as far as our log files.

The process's output is _available_ in the `CalledProcessError`; it just isn't printed by default. Now we print it before letting the exception propagate.


statham-arm:
Oops, I suddenly realise that starting the subject line "Automerge:" makes it look as if this _is_ a failed automerge, rather than a fix to the automerge script!

dcandler (approved):

pawosm-arm (approved):

PR #1006, opened 2026-08-18
[llvmlibc-support] Fix new hard-float PCS error from clang

statham-arm:
Commit b834b9eb8349dab has added a new diagnostic in clang, faulting any function definition which formally has the hard-float PCS but is compiled in a mode where the FP registers are unavailable. Calls _to_ such a function without the FP registers available are also faulted. This applies even if the function's type uses no floating-point registers anyway (e.g. `void f(void)`), and also even if the function is marked `[[clang::always_inline]]`.

The new diagnostic therefore affects our early setup function `_platform_setup_exceptions()`, which is marked
`[[gnu::target("no-fpregs")]]` because it runs before the FPU is enabled. It also affects the write to the VBAR register inside that function, because that's really the `operator=` method of a C++ class wrapping the system register. I've marked all those system register wrappers as soft-float, rather than trying to find just the one that caused the immediate problem.

### DIFF
--- a/arm-software/embedded/llvmlibc-support/crt0/exceptions.cpp
+++ b/arm-software/embedded/llvmlibc-support/crt0/exceptions.cpp
@@ -15,5 +15,5 @@
 #else
 // ARMv4T
 // TODO: fill in stub functions once we can start testing LLVM-libc
-extern "C" [[gnu::weak]] void _platform_setup_exceptions() {}
+extern "C" [[gnu::weak]] PLATFORM_SETUP_PCS void _platform_setup_exceptions() {}
 #endif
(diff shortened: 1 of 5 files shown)


vhscampos (approved):
Looks reasonable. Tagging @smithp35

smithp35 (approved):
LGTM too.

statham-arm:
The CI failure is unrelated. I'm not sure what causes it (the logs are confusing), but it's a build failure in the libunwind test suite in a picolibc-based library variant, so I can't believe it's caused by this change entirely confined to the llvmlibc-support sources.

PR #1008, opened 2026-08-19
[CodeGen][AArch64] Use vector parts for internal non-power-of-two vectors (#213982)

amilendra:
Cherry-pick from upstream llvm-project https://github.com/llvm/llvm-project/commit/d141332bd2ae3c07509f221feeb191b8bb27e50d

Decompose fixed-length, non-power-of-two vectors into legal vector parts, rather than scalarizing these. Apply this only to internal register  values.

Continue using the existing scalar breakdown for arguments and returns, preserving the existing calling convention behaviour.

Downstream issue: #1009


bipmis (approved):
LGTM. Now merged in upstream LLVM and gives a ~2% perf win in spec2017 namd.

dcandler (approved):

PR #1011, opened 2026-08-20
[ATfE] Prototype: LLVM libc file IO support

voltur01:
This is a prototype/RFC for adding file IO support to bare-metal LLVM libc. LLVM libc changes are intended to land in upstream LLVM - it is easier to show how the proposed interface is used with ATfE semihosting implementation though.

LLVM libc changes: The embedding API libc/src/__support/OSUtil/baremetal/io.h is extended with these functions:
- __llvm_libc_stdio_open
- __llvm_libc_stdio_close
- __llvm_libc_stdio_seek
- __llvm_libc_stdio_set_buffer
- __llvm_libc_stdio_flush which map naturally to, e.g. the API expected by picolibc.

These are used to implement the following additional C library functions that are enough to implement file IO required by libc++ fstream:
- fopen
- fclose
- fseek
- fseeko
- ftell
- ftello
- setbuf
- setvbuf
- fflush

The embedding API uses opaque pointer as a cookie, the semihosting implementation then keeps a small struct with either just the semihosting handle for standard streams or also current position to support seeking for file streams.

Semihosting implementation is unbuffered, however other implementations can provide it by default or through the buffering management functions (which are also required by C++ fstream).

Other implementations can use the struct of function pointers pattern for the cookie to easily implement support for multiple types of file systems, e.g. UART, flash or FAT.

The RFC is mainly about the embedding API and libc side implementation if it is reasonable for upstream LLVM libc.

Note: There is a TODO comment whether we need embedding API for ungetc, what we can do is declare a weak hook:

    __llvm_libc_stdio_ungetc(void *cookie, unsigned char ch)

and check in the current ungetc implementation if it is provided: use it if yes or keep current 1 char buffer otherwise.

### DIFF
--- a/arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
+++ b/arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
@@ -35,17 +35,17 @@ struct __llvm_libc_stdio_cookie __llvm_libc_stderr_cookie;
 
 namespace {
 // File cookie helpers
-bool is_std_stream_cookie(void *cookie) {
+bool is_std_stream(void *cookie) {
   return cookie == &__llvm_libc_stdin_cookie ||
          cookie == &__llvm_libc_stdout_cookie ||
          cookie == &__llvm_libc_stderr_cookie;
 }
 
-size_t handle_from_cookie(void *cookie) {
+size_t get_handle(void *cookie) {
   return static_cast<__llvm_libc_stdio_cookie *>(cookie)->handle;
 }
 
-__llvm_libc_semihost_file_cookie *allocate_file_cookie() {
+__llvm_libc_semihost_file_cookie *allocate_file() {
   for (size_t i = 0; i < __llvm_libc_semihost_file_cookie_storage.size; ++i) {
     auto *cookie = &__llvm_libc_semihost_file_cookie_storage.cookies[i];
     if (!cookie->in_use) {
@@ -57,11 +57,11 @@ __llvm_libc_semihost_file_cookie *allocate_file_cookie() {
   return nullptr;
 }
 
-void release_file_cookie(void *cookie) {
+void release_file(void *cookie) {
   static_cast<__llvm_libc_semihost_file_cookie *>(cookie)->in_use = false;
 }
 
-off_t position_from_cookie(void *cookie) {
+off_t get_file_position(void *cookie) {
   return static_cast<__llvm_libc_semihost_file_cookie *>(cookie)->position;
 }
 
@@ -69,17 +69,16 @@ void set_file_position(void *cookie, off_t position) {
   static_cast<__llvm_libc_semihost_file_cookie *>(cookie)->position = position;
 }
 
-void advance_file_position(void *cookie, ssize_t amount) {
-  if (amount > 0)
-    set_file_position(cookie, position_from_cookie(cookie) +
-                                  static_cast<off_t>(amount));
+void advance_file_position(void *cookie, size_t amount) {
+  set_file_position(cookie,
+                    get_file_position(cookie) + static_cast<off_t>(amount));
 }
 } // namespace
 
 namespace {
 // Helper functions implemented here to avoid dependency on libc which is not
 // available in LLVM libc hermetic testing.
-static int _isspace(char ch) {
+int _isspace(char ch) {
   return ch == ' ' || ch == '\t' || ch == '\r' || ch == '\n' || ch == '\v' ||
          ch == '\f';
 }
@@ -143,11 +142,11 @@ int semihost_seek(size_t handle, off_t position) {
   return semihosting_call(SYS_SEEK, args) == 0 ? 0 : semihost_errno_negative();
 }
 
-long semihost_open(const char *path, size_t mode) {
+long semihost_open(const char *path, size_t length, size_t mode) {
   size_t args[] = {
       reinterpret_cast<size_t>(path),
       mode,
-      _strlen(path),
+      length,
   };
   long handle = semihosting_call(SYS_OPEN, args);
   return handle < 0 ? semihost_errno_negative() : handle;
@@ -160,7 +159,8 @@ int semihost_close(size_t handle) {
 
 void stdio_cookie_open(struct __llvm_libc_stdio_cookie *cookie, size_t mode) {
   const char std_stream_name[] = ":tt";
-  cookie->handle = static_cast<size_t>(semihost_open(std_stream_name, mode));
+  cookie->handle = static_cast<size_t>(
+      semihost_open(std_stream_name, sizeof(std_stream_name) - 1, mode));
 }
 
 } // namespace
@@ -197,23 +197,27 @@ void __llvm_libc_exit(int status) {
 ssize_t __llvm_libc_stdio_read(void *cookie, char *buf, size_t size) {
   if (cookie == &__llvm_libc_stdin_cookie)
     return semihost_read_console(buf, size);
-  if (is_std_stream_cookie(cookie))
+  if (is_std_stream(cookie))
     return -EBADF;
 
-  ssize_t result = semihost_read(handle_from_cookie(cookie), buf, size);
-  advance_file_position(cookie, result);
+  ssize_t result = semihost_read(get_handle(cookie), buf, size);
+  if (result < 0)
+    return result;
+  advance_file_position(cookie, static_cast<size_t>(result));
   return result;
 }
 
 ssize_t __llvm_libc_stdio_write(void *cookie, const char *buf, size_t size) {
-  ssize_t result = semihost_write(handle_from_cookie(cookie), buf, size);
-  if (!is_std_stream_cookie(cookie))
-    advance_file_position(cookie, result);
+  ssize_t result = semihost_write(get_handle(cookie), buf, size);
+  if (result < 0)
+    return result;
+  if (!is_std_stream(cookie))
+    advance_file_position(cookie, static_cast<size_t>(result));
   return result;
 }
 
 int __llvm_libc_stdio_open(const char *path, const char *mode, void **cookie) {
-  auto *file_cookie = allocate_file_cookie();
+  auto *file_cookie = allocate_file();
   if (!file_cookie)
     return -EMFILE;
 
@@ -229,17 +233,17 @@ int __llvm_libc_stdio_open(const char *path, const char *mode, void **cookie) {
       open_mode |= OPENMODE_B;
   }
 
-  long handle = semihost_open(path, open_mode);
+  long handle = semihost_open(path, _strlen(path), open_mode);
   if (handle < 0) {
-    release_file_cookie(file_cookie);
+    release_file(file_cookie);
     return static_cast<int>(handle);
   }
   file_cookie->stdio_cookie.handle = static_cast<size_t>(handle);
   if (mode[0] == 'a') {
-    off_t length = semihost_file_length(handle_from_cookie(file_cookie));
+    off_t length = semihost_file_length(get_handle(file_cookie));
     if (length < 0) {
-      semihost_close(handle_from_cookie(file_cookie));
-      release_file_cookie(file_cookie);
+      semihost_close(get_handle(file_cookie));
+      release_file(file_cookie);
       return static_cast<int>(length);
     }
     set_file_position(file_cookie, length);
@@ -249,16 +253,16 @@ int __llvm_libc_stdio_open(const char *path, const char *mode, void **cookie) {
 }
 
 off_t __llvm_libc_stdio_seek(void *cookie, off_t offset, int whence) {
-  if (is_std_stream_cookie(cookie))
+  if (is_std_stream(cookie))
     return -ESPIPE;
 
   off_t base;
   if (whence == SEEK_SET)
     base = 0;
   else if (whence == SEEK_CUR)
-    base = position_from_cookie(cookie);
+    base = get_file_position(cookie);
   else if (whence == SEEK_END) {
-    base = semihost_file_length(handle_from_cookie(cookie));
+    base = semihost_file_length(get_handle(cookie));
     if (base < 0)
       return base;
   } else
@@ -267,7 +271,7 @@ off_t __llvm_libc_stdio_seek(void *cookie, off_t offset, int whence) {
   off_t position = base + offset;
   if (position < 0)
     return -EINVAL;
-  int error = semihost_seek(handle_from_cookie(cookie), position);
+  int error = semihost_seek(get_handle(cookie), position);
   if (error)
     return error;
   set_file_position(cookie, position);
@@ -285,9 +289,9 @@ int __llvm_libc_stdio_flush(void *) {
 }
 
 int __llvm_libc_stdio_close(void *cookie) {
-  int result = semihost_close(handle_from_cookie(cookie));
-  if (!is_std_stream_cookie(cookie))
-    release_file_cookie(cookie);
+  int result = semihost_close(get_handle(cookie));
+  if (!is_std_stream(cookie))
+    release_file(cookie);
   return result;
 }
 
--- a/arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
+++ b/arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
@@ -1,29 +1,168 @@
 //
-// Copyright (c) 2022-2025, Arm Limited and affiliates.
+// Copyright (c) 2022-2026, Arm Limited and affiliates.
 //
-// Part of the Arm Toolchain project, under the Apache License v2.0 with LLVM Exceptions. 
-// See https://llvm.org/LICENSE.txt for license information.
+// Part of the Arm Toolchain project, under the Apache License v2.0 with LLVM
+// Exceptions. See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
 
 #include "semihost.h"
 #include "platform.h"
 
+#include <errno.h>
 #include <stddef.h>
+#include <stdio.h>
 #include <stdlib.h>
 #include <time.h>
 
 namespace {
+// File cookies
+inline constexpr size_t DEFAULT_FILE_COOKIE_COUNT = 4;
+__llvm_libc_semihost_file_cookie
+    default_file_cookies[DEFAULT_FILE_COOKIE_COUNT];
+} // namespace
 
-void stdio_open(struct __llvm_libc_stdio_cookie *cookie, size_t mode) {
-  const char std_stream_name[] = ":tt";
-  size_t args[] = {
-      reinterpret_cast<size_t>(std_stream_name),
-      mode,
-      sizeof(std_stream_name) - 1UL,
-  };
-  cookie->handle = semihosting_call(SYS_OPEN, args);
+extern "C" {
+// Override this weak definition with a pool backed by an application-owned
+// array to change the maximum number of simultaneously open files.
+__attribute__((weak)) __llvm_libc_semihost_file_cookie_pool
+    __llvm_libc_semihost_file_cookie_storage = {default_file_cookies,
+                                                DEFAULT_FILE_COOKIE_COUNT};
+struct __llvm_libc_stdio_cookie __llvm_libc_stdin_cookie;
+struct __llvm_libc_stdio_cookie __llvm_libc_stdout_cookie;
+struct __llvm_libc_stdio_cookie __llvm_libc_stderr_cookie;
 }
+
+namespace {
+// File cookie helpers
+bool is_std_stream_cookie(void *cookie) {
+  return cookie == &__llvm_libc_stdin_cookie ||
+         cookie == &__llvm_libc_stdout_cookie ||
+         cookie == &__llvm_libc_stderr_cookie;
+}
+
+size_t handle_from_cookie(void *cookie) {
+  return static_cast<__llvm_libc_stdio_cookie *>(cookie)->handle;
+}
+
+__llvm_libc_semihost_file_cookie *allocate_file_cookie() {
+  for (size_t i = 0; i < __llvm_libc_semihost_file_cookie_storage.size; ++i) {
+    auto *cookie = &__llvm_libc_semihost_file_cookie_storage.cookies[i];
+    if (!cookie->in_use) {
+      cookie->in_use = true;
+      cookie->position = 0;
+      return cookie;
+    }
+  }
+  return nullptr;
+}
+
+void release_file_cookie(void *cookie) {
+  static_cast<__llvm_libc_semihost_file_cookie *>(cookie)->in_use = false;
+}
+
+off_t position_from_cookie(void *cookie) {
+  return static_cast<__llvm_libc_semihost_file_cookie *>(cookie)->position;
+}
+
+void set_file_position(void *cookie, off_t position) {
+  static_cast<__llvm_libc_semihost_file_cookie *>(cookie)->position = position;
+}
+
+void advance_file_position(void *cookie, ssize_t amount) {
+  if (amount > 0)
+    set_file_position(cookie, position_from_cookie(cookie) +
+                                  static_cast<off_t>(amount));
+}
+} // namespace
+
+namespace {
+// Helper functions implemented here to avoid dependency on libc which is not
+// available in LLVM libc hermetic testing.
+static int _isspace(char ch) {
+  return ch == ' ' || ch == '\t' || ch == '\r' || ch == '\n' || ch == '\v' ||
+         ch == '\f';
+}
+
+__attribute__((no_builtin("strlen"))) size_t _strlen(const char *str) {
+  const char *end = str;
+  while (*end)
+    ++end;
+  return static_cast<size_t>(end - str);
+}
+
+// Semihosting helpers
+int semihost_errno_negative() {
+  int error = semihosting_call(SYS_ERRNO, nullptr);
+  return error > 0 ? -error : -EIO;
+}
+
+off_t semihost_file_length(size_t handle) {
+  size_t args[] = {handle};
+  off_t length = semihosting_call(SYS_FLEN, args);
+  return length < 0 ? semihost_errno_negative() : length;
+}
+
+ssize_t semihost_read(size_t handle, char *buf, size_t size) {
+  size_t args[] = {
+      handle,
+      reinterpret_cast<size_t>(buf),
+      size,
+  };
+  ssize_t not_read = semihosting_call(SYS_READ, args);
+  return not_read < 0 ? semihost_errno_negative()
+                      : static_cast<ssize_t>(size) - not_read;
+}
+
+ssize_t semihost_read_console(char *buf, size_t size) {
+  for (size_t i = 0; i < size; ++i) {
+    long ch = semihosting_call(SYS_READC, nullptr);
+    buf[i] = static_cast<char>(ch & 0xff);
+    if (buf[i] == '\r')
+      buf[i] = '\n';
+  }
+  return static_cast<ssize_t>(size);
+}
+
+ssize_t semihost_write(size_t handle, const char *buf, size_t size) {
+  size_t args[] = {
+      handle,
+      reinterpret_cast<size_t>(buf),
+      size,
+  };
+  ssize_t not_written = semihosting_call(SYS_WRITE, args);
+  return not_written < 0 ? semihost_errno_negative()
+                         : static_cast<ssize_t>(size) - not_written;
+}
+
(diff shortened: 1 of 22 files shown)


arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
```diff
@@ -1,29 +1,168 @@
...
+
+void set_file_position(void *cookie, off_t position) {
+  static_cast<__llvm_libc_semihost_file_cookie *>(cookie)->position = position;
+}
+
+void advance_file_position(void *cookie, ssize_t amount) {
+  if (amount > 0)
+    set_file_position(cookie, position_from_cookie(cookie) +
+                                  static_cast<off_t>(amount));
+}
```
smithp35, line 75: Have we got the opportunity to error/abort with negative offsets not-supported.

arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
```diff
@@ -1,29 +1,168 @@
...
+
+void release_file_cookie(void *cookie) {
+  static_cast<__llvm_libc_semihost_file_cookie *>(cookie)->in_use = false;
+}
+
+off_t position_from_cookie(void *cookie) {
+  return static_cast<__llvm_libc_semihost_file_cookie *>(cookie)->position;
+}
+
+void set_file_position(void *cookie, off_t position) {
```
smithp35, line 68: The naming convention is a bit inconsistent. I think it would be better to remove the `_cookie` suffix from the previous functions, so `release_file_cookie` becomes `release_file`, may need to rename `position_from_cookie` to `get_position`.

Alternatively make sure that `_cookie` is always added.

arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
```diff
@@ -1,29 +1,168 @@
...
+  if (amount > 0)
+    set_file_position(cookie, position_from_cookie(cookie) +
+                                  static_cast<off_t>(amount));
+}
+} // namespace
+
+namespace {
+// Helper functions implemented here to avoid dependency on libc which is not
+// available in LLVM libc hermetic testing.
+static int _isspace(char ch) {
```
smithp35: This is in an anonymous namespace already. Does it need `static`?

arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
```diff
@@ -1,29 +1,168 @@
...
+  };
+  return semihosting_call(SYS_SEEK, args) == 0 ? 0 : semihost_errno_negative();
+}
+
+long semihost_open(const char *path, size_t mode) {
+  size_t args[] = {
+      reinterpret_cast<size_t>(path),
       mode,
-      sizeof(std_stream_name) - 1UL,
+      _strlen(path),
```
smithp35: Could make length a parameter to make the caller pass it in. Would save a call to _strlen when the length can be statically determined, for example `:tt`.

arm-software/embedded/llvmlibc-support/semihost/semihost.h
```diff
@@ -112,6 +114,22 @@ inline constexpr uint32_t OPENMODE_B = 1;
...
 } // namespace
 
-struct __llvm_libc_stdio_cookie { int handle; };
+struct __llvm_libc_stdio_cookie {
+  size_t handle;
+};
+
+struct __llvm_libc_semihost_file_cookie {
+  __llvm_libc_stdio_cookie stdio_cookie;
+  bool in_use;
```
smithp35, line 123: struct will be smaller if the fields are in order of alignment. If `in_use` is after `position` no alignment padding will be needed between the fields.

smithp35 (commented):
Some small nits, but overall looks good to me. Maybe worth @vhscampos to look over the llvmlibc parts.

arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
voltur01: Yes, there seem to be too many cookies - I removed the suffix and made all names start with a verb.

arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
voltur01: Updated this to receive an unsigned value and added check/error return at the call site.

arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
voltur01: Oh, good spot: this is a left over the previous implementation of argv parsing. Removed.

arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
voltur01: Good point - updated.

arm-software/embedded/llvmlibc-support/semihost/semihost.h
voltur01: Doing some checks, if I am getting them right, it looks like on Arm 32-bit swapping the fields would introduce more padding: inside and in between array elements in the pool.

Now it is:
```
handle    offset 0, size 4
in_use    offset 4, size 1
padding             size 3
position  offset 8, size 8
total               16 bytes
```
with fields swapped:
```
handle    offset 0,  size 4
padding   offset 4,  size 4
position  offset 8,  size 8
in_use    offset 16, size 1
padding   offset 17, size 7
total                24 bytes
```
On AArch64 the size should be the same in both cases.

voltur01 (commented):
Thank you for the review! Please see updates.

arm-software/embedded/llvmlibc-support/semihost/semihost.h
smithp35: I hadn't realised off_t was an 8-byte aligned 8-byte type on AArch32. That would make the minimum size 16 (multiple of alignment) so any saving in the padding would get eaten up by the trailing padding.

Can ignore. In future, if we need more properties could make it a size_t flags field with 1 bit for in_use, but no need for that now.

smithp35 (dismissed):
Thanks LGTM on my side. I've set approved, but could be worth waiting a day or so to see if there are any other comments from other reviewers.

arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
```diff
@@ -1,29 +1,168 @@
...
 
+#include <errno.h>
 #include <stddef.h>
+#include <stdio.h>
 #include <stdlib.h>
 #include <time.h>
 
 namespace {
+// File cookies
+inline constexpr size_t DEFAULT_FILE_COOKIE_COUNT = 4;
```
vhscampos, line 20: I believe there is no point in `inline constexpr` because this is not a header file, so the issue of multiple definitions at linking is not pertinent. Just `constexpr` is enough.

arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
```diff
@@ -1,29 +1,168 @@
...
+
+off_t get_file_position(void *cookie) {
+  return static_cast<__llvm_libc_semihost_file_cookie *>(cookie)->position;
+}
+
+void set_file_position(void *cookie, off_t position) {
+  static_cast<__llvm_libc_semihost_file_cookie *>(cookie)->position = position;
+}
+
+void advance_file_position(void *cookie, size_t amount) {
```
vhscampos, line 72: Why not receive an `off_t` for consistency?

arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
```diff
@@ -1,29 +1,168 @@
...
+  set_file_position(cookie,
+                    get_file_position(cookie) + static_cast<off_t>(amount));
+}
+} // namespace
+
+namespace {
+// Helper functions implemented here to avoid dependency on libc which is not
+// available in LLVM libc hermetic testing.
```
vhscampos, line 80: Can you expand on this? What isn't available in the hermetic testing?

arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
```diff
@@ -1,29 +1,168 @@
...
+  };
+  ssize_t not_read = semihosting_call(SYS_READ, args);
+  return not_read < 0 ? semihost_errno_negative()
+                      : static_cast<ssize_t>(size) - not_read;
+}
+
+ssize_t semihost_read_console(char *buf, size_t size) {
+  for (size_t i = 0; i < size; ++i) {
+    long ch = semihosting_call(SYS_READC, nullptr);
+    buf[i] = static_cast<char>(ch & 0xff);
```
vhscampos, line 119: The masking and casting is not required. If you want to keep some of it for clarity, keep just the static_cast

```suggestion
    buf[i] = ch;
```

arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
```diff
@@ -55,41 +194,105 @@ void __llvm_libc_exit(int status) {
...
+  }
+
+  long handle = semihost_open(path, _strlen(path), open_mode);
+  if (handle < 0) {
+    release_file(file_cookie);
+    return static_cast<int>(handle);
   }
-  return size;
+  file_cookie->stdio_cookie.handle = static_cast<size_t>(handle);
+  if (mode[0] == 'a') {
```
vhscampos, line 242: Since `mode` has been parsed at this point, consult `open_mode` instead.

arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
```diff
@@ -55,41 +194,105 @@ void __llvm_libc_exit(int status) {
...
+    base = get_file_position(cookie);
+  else if (whence == SEEK_END) {
+    base = semihost_file_length(get_handle(cookie));
+    if (base < 0)
+      return base;
+  } else
+    return -EINVAL;
+
+  off_t position = base + offset;
+  if (position < 0)
```
vhscampos: This can be problematic because `off_t` is a signed type, so the overflow is undefined behaviour.

Check for it before the operation.


```suggestion
if (std::numeric_limits<off_t>::max() - offset < base)
  return -EINVAL;
off_t position = base + offset;
```

vhscampos (commented):
I am still going through it. I will notify once I have seen all files

arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
vhscampos: Inside libc you should in fact use cpp::numeric_limits, not std::

arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
```diff
@@ -55,41 +194,105 @@ void __llvm_libc_exit(int status) {
...
+  if (error)
+    return error;
+  set_file_position(cookie, position);
+  return position;
+}
+
+int __llvm_libc_stdio_set_buffer(void *, char *, size_t, int) {
+  // Semihost streams are unbuffered.
+  return 0;
+}
```
vhscampos, line 284: Should it return an error instead? It'd fail silently

arm-software/embedded/llvmlibc-support/semihost/semihost.h
```diff
@@ -13,6 +13,8 @@
 #ifndef LLVMET_LLVMLIBC_SUPPORT_SEMIHOST_H
 #define LLVMET_LLVMLIBC_SUPPORT_SEMIHOST_H
 
+#include <llvm-libc-types/off_t.h>
+#include <llvm-libc-types/size_t.h>
 #include <llvm-libc-types/ssize_t.h>
 #include <stdint.h>
```
vhscampos, line 19: The correct way is to include `hdr/header.h` so that this can work in libc overlay mode

libc/src/__support/OSUtil/baremetal/io.h
```diff
@@ -9,6 +9,7 @@
 #ifndef LLVM_LIBC_SRC___SUPPORT_OSUTIL_BAREMETAL_IO_H
 #define LLVM_LIBC_SRC___SUPPORT_OSUTIL_BAREMETAL_IO_H
 
+#include "include/llvm-libc-types/off_t.h"
 #include "include/llvm-libc-types/size_t.h"
 #include "include/llvm-libc-types/ssize_t.h"
```
vhscampos, line 14: Same comment about including from `hdr/` instead.

libc/src/stdio/baremetal/fclose.cpp
```diff
@@ -0,0 +1,12 @@
+#include "src/stdio/fclose.h"
```
vhscampos, line 1: Missing file header. Make sure to use the coding standard

libc/src/stdio/baremetal/fclose.cpp
```diff
@@ -0,0 +1,12 @@
...
+
+#include "src/__support/OSUtil/io.h"
+#include "src/__support/common.h"
+
+namespace LIBC_NAMESPACE_DECL {
```
vhscampos, line 6: This macro is defined in src/__support/macros/config.h

Include it and make sure to add it as a dependency in CMakeLists

libc/src/stdio/baremetal/fflush.cpp
vhscampos, line 7: Update the file header to use the new coding standard style

libc/src/stdio/baremetal/fflush.cpp
```diff
@@ -8,15 +8,13 @@
 
 #include "src/stdio/fflush.h"
 
+#include "src/__support/OSUtil/io.h"
 #include "src/__support/common.h"
 
 namespace LIBC_NAMESPACE_DECL {
```
vhscampos, line 14: Ditto for macros/config.h

libc/src/stdio/baremetal/fopen.cpp
```diff
@@ -0,0 +1,25 @@
+#include "src/stdio/fopen.h"
```
vhscampos, line 1: Missing file header

libc/src/stdio/baremetal/fopen.cpp
```diff
@@ -0,0 +1,25 @@
...
+
+#include "hdr/errno_macros.h"
+#include "src/__support/OSUtil/io.h"
+#include "src/__support/common.h"
+#include "src/__support/libc_errno.h"
+
+namespace LIBC_NAMESPACE_DECL {
```
vhscampos, line 8: Ditto for macros/config.h

libc/src/stdio/baremetal/fseek.cpp
```diff
@@ -0,0 +1,23 @@
+#include "src/stdio/fseek.h"
```
vhscampos, line 1: Missing file header

libc/src/stdio/baremetal/fseek.cpp
```diff
@@ -0,0 +1,23 @@
...
+
+#include "hdr/errno_macros.h"
+#include "src/__support/OSUtil/io.h"
+#include "src/__support/common.h"
+#include "src/__support/libc_errno.h"
+
+namespace LIBC_NAMESPACE_DECL {
```
vhscampos, line 8: Ditto for macros/config.h

libc/src/stdio/baremetal/fseeko.cpp
```diff
@@ -0,0 +1,31 @@
+//===-- Implementation of fseeko for baremetal -----------------*- C++ -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
```
vhscampos, line 7: Update to the new coding standard style

libc/src/stdio/baremetal/fseeko.cpp
```diff
@@ -0,0 +1,31 @@
...
+
+#include "src/stdio/fseeko.h"
+
+#include "hdr/errno_macros.h"
+#include "src/__support/OSUtil/io.h"
+#include "src/__support/common.h"
+#include "src/__support/libc_errno.h"
+
+namespace LIBC_NAMESPACE_DECL {
```
vhscampos, line 16: Ditto for macros/config.h

libc/src/stdio/baremetal/fseek.cpp
```diff
@@ -0,0 +1,23 @@
...
+
+LLVM_LIBC_FUNCTION(int, fseek, (::FILE * stream, long offset, int whence)) {
+  if (stream == nullptr) {
+    libc_errno = EINVAL;
+    return -1;
+  }
+  off_t result = __llvm_libc_stdio_seek(stream, offset, whence);
```
vhscampos, line 15: There's a silent cast from long to off_t here (the offset argument). Add a cast for better clarity to the reader.

libc/src/stdio/baremetal/ftell.cpp
```diff
@@ -0,0 +1,24 @@
+#include "src/stdio/ftell.h"
```
vhscampos, line 1: Missing file header

libc/src/stdio/baremetal/ftell.cpp
```diff
@@ -0,0 +1,24 @@
...
+
+#include "hdr/errno_macros.h"
+#include "hdr/stdio_macros.h"
+#include "src/__support/OSUtil/io.h"
+#include "src/__support/common.h"
+#include "src/__support/libc_errno.h"
+
+namespace LIBC_NAMESPACE_DECL {
```
vhscampos, line 9: Ditto for macros/config.h

libc/src/stdio/baremetal/ftello.cpp
```diff
@@ -0,0 +1,32 @@
+//===-- Implementation of ftello for baremetal -----------------*- C++ -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
```
vhscampos: Update to the new style

libc/src/stdio/baremetal/ftello.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+#include "src/stdio/ftello.h"
+
+#include "hdr/errno_macros.h"
+#include "hdr/stdio_macros.h"
+#include "src/__support/OSUtil/io.h"
+#include "src/__support/common.h"
+#include "src/__support/libc_errno.h"
+
+namespace LIBC_NAMESPACE_DECL {
```
vhscampos, line 23: Ditto for macros/config.h

libc/src/stdio/baremetal/setbuf.cpp
```diff
@@ -0,0 +1,22 @@
+//===-- Bare-metal implementation of setbuf ------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
```
vhscampos: Update to the new style

libc/src/stdio/baremetal/setbuf.cpp
```diff
@@ -0,0 +1,22 @@
...
+
+#include "src/stdio/setbuf.h"
+#include "hdr/stdio_macros.h"
+#include "src/__support/OSUtil/io.h"
+#include "src/__support/common.h"
+
+namespace LIBC_NAMESPACE_DECL {
```
vhscampos, line 20: Ditto for macros/config.h

libc/src/stdio/baremetal/setvbuf.cpp
```diff
@@ -0,0 +1,21 @@
+//===-- Bare-metal implementation of setvbuf -----------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
```
vhscampos: Update to the new style

libc/src/stdio/baremetal/setvbuf.cpp
```diff
@@ -0,0 +1,21 @@
...
+
+#include "src/stdio/setvbuf.h"
+#include "src/__support/OSUtil/io.h"
+#include "src/__support/common.h"
+
+namespace LIBC_NAMESPACE_DECL {
```
vhscampos, line 19: Ditto for macros/config.h

libc/src/stdio/generic/CMakeLists.txt
```diff
@@ -10,6 +10,32 @@ function(add_generic_entrypoint_object name)
   endif()
 endfunction()
 
+add_generic_entrypoint_object(
+  setbuf
+  SRCS
+    ../setbuf.cpp
```
vhscampos, line 16: Should these files, setbuf.cpp and setbuf.h, be moved into the generic/ folder?

vhscampos:
@voltur01 I finished my review

arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
voltur01: Good point, fixed.

arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
voltur01: This was meant to convey that we are only moving forward, however you are right - the code will be simpler with `off_t` and more consistent - updated.

arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
voltur01: Hermetic tests are run with `-nostdlib` so that C library is not available. Each C library function is tested in isolation and cannot use any other C function, because of `-nostdlib`.
And the `no_builtin("strlen")` below is needed so that the compiler does not introduce `strlen` anyway as per the idiom recognition.

arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
voltur01: Good point, I kept the `static_cast` to make sure there is no warnings.

arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
voltur01: Indeed, updated.

arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
voltur01: Good point, would use of `__builtin_add_overflow(a, b, &result)` be justified here to handle both positive and negative cases?

Preliminary, updated like that.

arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
voltur01: Good question, I think the trade off is that now it returns success and then works, but slower than expected, while if it returned an error the application code can abort and not work at all.

So this is a stub to try to keep apps going as much as possible.

arm-software/embedded/llvmlibc-support/semihost/semihost.h
voltur01: `libsemihost` is built as a separate, external to llvmlibc, project so I do not believe we have these libc internal headers available. And we only build it in one mode since this is ATfE specific.

libc/src/__support/OSUtil/baremetal/io.h
voltur01: Here this makes sense - updated.

libc/src/stdio/baremetal/fclose.cpp
voltur01: Thanks - added!

libc/src/stdio/baremetal/fclose.cpp
voltur01: Added, thanks.

libc/src/stdio/baremetal/fflush.cpp
voltur01: Updated.

libc/src/stdio/baremetal/fflush.cpp
voltur01: Updated.

libc/src/stdio/baremetal/fopen.cpp
voltur01: Added.

libc/src/stdio/baremetal/fopen.cpp
voltur01: Done.

libc/src/stdio/baremetal/fseek.cpp
voltur01: Added.

libc/src/stdio/baremetal/fseek.cpp
voltur01: Fixed.

libc/src/stdio/baremetal/fseek.cpp
voltur01: Added.

libc/src/stdio/baremetal/fseeko.cpp
voltur01: Updated.

libc/src/stdio/baremetal/fseeko.cpp
voltur01: Updated.

libc/src/stdio/baremetal/ftell.cpp
voltur01: Added.

libc/src/stdio/baremetal/ftell.cpp
voltur01: Fixed.

libc/src/stdio/baremetal/ftello.cpp
voltur01: Updated.

libc/src/stdio/baremetal/ftello.cpp
voltur01: Fixed.

libc/src/stdio/baremetal/setbuf.cpp
voltur01: Updated.

libc/src/stdio/baremetal/setbuf.cpp
voltur01: Fixed.

libc/src/stdio/baremetal/setvbuf.cpp
voltur01: Updated.

libc/src/stdio/baremetal/setvbuf.cpp
voltur01: Done.

libc/src/stdio/generic/CMakeLists.txt
voltur01: It may be a good idea, however it looks like an unrelated refactoring, so maybe better keep out of scope of this change to make the point of the RFC. 

Or maybe this can be part of upstreaming, assuming the approach is accepted - I will make a note for myself.

voltur01 (commented):
Thank you for the detailed review @vhscampos ! I updated the PR.

vhscampos (dismissed):
LGTM!

arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
vhscampos: Actually when I saw the original code, I though that the `if` was to detect overflows, but in fact it's for detecting negative positions? If so, the overflow detection could be overkill here. Something we can iron out on the final PR for upstream

voltur01:
This was split into new proper PRs:

    LLVM libc embedding API: https://github.com/llvm/llvm-project/pull/221211
    ATfE semihosting implementation: https://github.com/arm/arm-toolchain/pull/1041
    ATfE fstream enablement: https://github.com/arm/arm-toolchain/pull/1040

Closing.

PR #1012, opened 2026-08-20
[ATfE] Add expected output of samples to READMEs

dcandler:
Some of the samples have quite different behaviour, so it may not be obvious when they are not working correctly, especially when some deliberately exit with a hardfault. Adding the expected output to the README files should make it easier to see when a sample is doing what it is meant to.

This also contains a small fix to the `cpp-baremetal-semihosting-newlib` makefile, since only one of the images was being converted to hex.

### DIFF
--- a/arm-software/embedded/newlib-samples/src/cpp-baremetal-semihosting-newlib/README.md
+++ b/arm-software/embedded/newlib-samples/src/cpp-baremetal-semihosting-newlib/README.md
@@ -9,7 +9,14 @@ to build C++ programs with `newlib`.
 A recent version of QEMU is required to run the sample, because of the
 dependency on details of semihosting implementation in QEMU.
 
-Running the sample using `make run` should print a message:
+Running the sample using `make run` should compile and run two files,
+`hello.hex` and `hello-exn.hex`.
+`hello.hex`, compiled without exceptions, will print a message indicating the
+exception was not caught:
+```
+Hello World!
+```
+Whereas `hello-exn.hex` compiled with exceptions, will catch it:
 ```
 Hello World!
 Exception caught.
--- a/arm-software/embedded/newlib-samples/src/cpp-baremetal-semihosting-newlib/README.md
+++ b/arm-software/embedded/newlib-samples/src/cpp-baremetal-semihosting-newlib/README.md
@@ -8,3 +8,9 @@ to build C++ programs with `newlib`.
 
 A recent version of QEMU is required to run the sample, because of the
 dependency on details of semihosting implementation in QEMU.
+
+Running the sample using `make run` should print a message:
+```
+Hello World!
+Exception caught.
+```
--- a/arm-software/embedded/samples/src/cpp-baremetal-semihosting-cfi/README.md
+++ b/arm-software/embedded/samples/src/cpp-baremetal-semihosting-cfi/README.md
@@ -5,9 +5,24 @@ This sample shows how to use
 sanitizer to detect certain kinds of undefined behavior 
 that can subvert the control flow of C++ code at run-time.
 
-The default target `make build` enables CFI, 
-to see the behavior without CFI use the following commands:
+The default target `make build` enables CFI, so that `make run` will encounter
+a hard fault during execution and display a partial register dump, e.g:
 ```
-make build-no-cfi
-make run
+ARM fault: hardfault
+        R0:   0x00000001
+        R1:   0x20000d3c
+        R2:   0x2000093c
+        R3:   0x00000000
+        R12:  0x00000000
+        LR:   0x00011407
+        PC:   0x000115ec
+        XPSR: 0x01000000
+make: [Makefile:28: run] Error 1 (ignored)
+```
+Running `make build-no-cfi` followed by `make run` will demonstrate the
+behavior without CFI, displaying a message as the undefined behaviour is not
+caught:
+```
+Bad
+C++ CFI sample
 ```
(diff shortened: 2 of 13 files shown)


arm-software/embedded/samples/src/cpp-baremetal-semihosting-cfi/README.md
```diff
@@ -5,9 +5,24 @@ This sample shows how to use
 sanitizer to detect certain kinds of undefined behavior 
 that can subvert the control flow of C++ code at run-time.
 
-The default target `make build` enables CFI, 
-to see the behavior without CFI use the following commands:
+The default target `make build` enables CFI, so that `make run` will encounter
+a hard fault during execution and display a partial register dump, e.g:
```
voltur01, line 9: Should we highlight that the dump is different for picolibc and LLVM libc? or at least say it depends on libc.

arm-software/embedded/samples/src/cpp-baremetal-semihosting-prof/README.md
```diff
@@ -4,4 +4,26 @@ This sample shows how to use instrumentation to emit profile data
 and use it to show code coverage.
 
 Use `make run` to build with instrumentation, run and collect the raw profile data,
-then output a visualization of the code coverage.
+then output a visualization of the code coverage, e.g:
+```
+    1|       |/* Copyright (c) 2023-2025, Arm Limited and affiliates.
```
HugoSilvaSantos, line 9: NIT: Should the copyright date be updated to 2026?
Or is it irrelevant in this context?

arm-software/embedded/newlib-samples/src/cpp-baremetal-semihosting-newlib/README.md
```diff
@@ -8,3 +8,9 @@ to build C++ programs with `newlib`.
 
 A recent version of QEMU is required to run the sample, because of the
 dependency on details of semihosting implementation in QEMU.
+
+Running the sample using `make run` should print a message:
+```
+Hello World!
```
statham-arm, line 21: According to my notes from testing it last month, and also looking at the Makefile now, this sample builds two separate images, `hello` and `hello-exn`. `hello` just prints "Hello World!", and `hello-exn` prints that and _also_ then throws an exception and demonstrates catching it. `make run` looks as if it runs both in succession, so shouldn't we expect the output from both programs here?

arm-software/embedded/samples/src/cpp-baremetal-semihosting-ubsan/README.md
```diff
@@ -7,10 +7,26 @@ to detect undefined behavior in C++ code at run-time.
...
+detected, e.g:
+```
+UBSAN: add-overflow (recovered)
 
-Use the following commands to invoke the trap mode:
+C++ UBSan sample
+```
+Alternatively, `make build-trap` followed by `make run` will instead invoke
+the trap mode, so that the same exits with a hardfault and partial register
+dump, e.g:
```
voltur01, line 20: Same here, different per libc.

voltur01 (commented):
One idea, otherwise LGTM.

arm-software/embedded/samples/src/cpp-baremetal-semihosting-cfi/README.md
dcandler: That will indeed likely avoid confusion in future, so done.

arm-software/embedded/samples/src/cpp-baremetal-semihosting-ubsan/README.md
dcandler: That will indeed likely avoid confusion in future, so done.

arm-software/embedded/samples/src/cpp-baremetal-semihosting-prof/README.md
dcandler: This just repeats the contents of the source file, all the sources for the samples have the same date currently.

arm-software/embedded/newlib-samples/src/cpp-baremetal-semihosting-newlib/README.md
dcandler: True, I did both outputs for the similar `cpp-baremetal-semihosting-exceptions` sample so I'm not sure why I missed this one. Although, while updating this, I did then notice the makefile was in fact only converting one into a hex file instead of both, so I've also corrected that at the same time.

voltur01: Yes, the two are run in sequence and this seems to be said above:
>`make run` should compile and run two files

statham-arm: Ah, and you've corrected the Makefile in a way that also adds the missing dependency, great!

statham-arm (approved):
I'm happy that my feedback is addressed.

arm-software/embedded/newlib-samples/src/cpp-baremetal-semihosting-newlib/README.md
voltur01: > only converting one into a hex file

Hm, good point and two ideas: either we should make it consistent and use both as `hex` or rather why would not we just use `elf` in all cases and save one step in Makefiles? It maybe that the original sample we tried with actual micro:bit board thus had to use a `hex` file.

voltur01 (approved):
LGTM thank you for the updates!

Issue #965, opened 2026-07-30
[Downstream Change] Fix epilogue code generation issue related to commit ccef34d on ATfE 23 branch


john-brawn-arm:
**What is the change**
Commit https://github.com/arm/arm-toolchain/commit/ccef34d45ae1ec28c129612a53f24b0e160912aa has caused a severe regression on an important embedded benchmark. It needs to be reverted on the ATfE 23 branch to fix the regression. There are later changes that build on top of this commit which will also need to be reverted. The list of commits to revert is:

https://github.com/arm/arm-toolchain/commit/ca970f3bb3ce41365dfc85049db44489c88cb0a5
https://github.com/arm/arm-toolchain/commit/a47be1995a65a411527062e7ca7e92fff8d6ad1b
https://github.com/arm/arm-toolchain/commit/024a691e7784e7ebe3783ddaf557ee07a127d934
https://github.com/arm/arm-toolchain/commit/ccef34d45ae1ec28c129612a53f24b0e160912aa

**Why this change cannot be done upstream**
The regression is only indirectly caused by the above commit, so reverting it isn't an acceptable long-term solution. There's work being done upstream to fix this in the correct way:
https://github.com/llvm/llvm-project/pull/210061
https://github.com/llvm/llvm-project/pull/211274
https://github.com/llvm/llvm-project/pull/212230
but this likely won't all be finished in time for the ATfE 23 release.

amendment: reverting the offending commits will impact ATfL too much. But a recent upstream main commit fixes the same issue with on first sight none of the detrimental regressions. Pull request https://github.com/arm/arm-toolchain/pull/1015 supersedes the above suggested revert.

PR #1015, opened 2026-08-21
[Downstream Change] [LV] Prefer fixed VFs for epilogue vectorization (#211814)

stuij:
Unless forced (via `-scalable-vectorization=always`), or if the fixed vector cost is much higher than scalable vectors (which occurs for some loops, such as histograms), prefer fixed vectors for the epilogue.

There are currently some hidden costs for scalable epilogues that are not modeled (such as reduced post-vectorization unrolling).

This restriction likely can be lifted/reworked once we support predicated vector epilogue.

Downstream issue: #965


stuij:
This patch, which was merged in upstream LLVM last week, is a much less invasive answer to the embedded regression that #966 addresses.

pawosm-arm (approved):
Following Kiran's recommendation.

dcandler (approved):

PR #1016, opened 2026-08-21
Move 23.x automerge into its own workflow

dcandler:
Currently, one automerge workflow handles automerging for both trunk and release branches. However there are some divergences in objective between the two branches. The trunk branch always should be kept to up date with upstream, whereas the release branch should only merge up to a release tag. This requires stopping, and manual dispatches.

Rather than fill the workflow with conditional logic to handle the two branches differently, this patch splits 23.x into its own workflow. This makes it easier to disable merging on the release brnach where necessary, and allows for an input on the workflow_dispatch to manually trigger a merge up to a specific point, such as a commit corresponding to an upstream release tag.

Both workflows have been added to the same concurrency group, so that they cannot run at the same time. The current automerge script should not be run concurrently, and could potentially encounter a race condition if both branches generate merge conflicts at the same time.

### DIFF
--- a/.github/workflows/automerge_23x.yml
+++ b/.github/workflows/automerge_23x.yml
@@ -1,13 +1,18 @@
 # This workflow executes the automerge python script to automatically merge
-# changes from the `main` branch of upstream LLVM into the `arm-software`
+# changes from the `release/23.x` branch of upstream LLVM into the `release/arm-software/23.x`
 # branch of the arm/arm-toolchain repository.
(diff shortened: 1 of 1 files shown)


HugoSilvaSantos (approved):

pawosm-arm (approved):
I assume it doesn't affect our release cycle.

dcandler:
> I assume it doesn't affect our release cycle.

It automates a previously manual step (updating the branch to the final upstream release tag)

PR #1018, opened 2026-08-24
[ATfL][docs] Extract ATfL-specific notes to a separate file

pawosm-arm:


pawosm-arm:
I think the stale part of release notes should be exposed in a separate file (so they stop polluting the `CHANGELOG.md` file) like the one introduced by this patch. After merged, this should be cherry-picked to the release branch.

@garyjcarroll if this gets accepted, would it be easily doable to introduce this as a part of the public release notes?

amilendra:
> I think the stale part of release notes should be exposed in a separate file (so they stop polluting the `CHANGELOG.md` file) like the one introduced by this patch. After merged, this should be cherry-picked to the release branch.
> 
> @garyjcarroll if this gets accepted, would it be easily doable to introduce this as a part of the public release notes?

Shouldn't this patch also remove the same content which is in `CHANGELOG.md`?

pawosm-arm:
> Shouldn't this patch also remove the same content which is in `CHANGELOG.md`?

I'm not sure. In theory, the past entries in the Changelog should not change.

pawosm-arm:
ping...

pawosm-arm:
The collective decision was to postpone this docs 'refactor' to the next release, so it seems it won't be cherry-picked to the ongoing release branch.

pawosm-arm:
Apparently we came up with a better idea (a common section at the beginning of the changelog file). Closing it.

PR #1029, opened 2026-08-27
[ATfE] Update _LIBCPP_HAS_C8RTOMB_MBRTOC8 patch for LLVM libc

voltur01:
LLVM libc does not declare c8rtomb() and mbrtoc8() functions similar to old versions of picolibc, thus replace the unconditional definition of _LIBCPP_HAS_C8RTOMB_MBRTOC8 (that was there as a workaround for the current latest version of picolibc) with a check to define it only for picolibc versions that provide the functions.

This is still a workaround to enabled libc++ testing with LLVM libc before the issue is properly resolved in upstream: https://github.com/llvm/llvm-project/pull/165439


statham-arm (approved):
LGTM, but please improve the commit message, because it doesn't match the code change. You say you're just making the #define picolibc-specific, but the patch is also checking the picolibc version number, which it previously didn't. Please at least mention that, and say why we need to add it (or why we should have had it all along).

voltur01:
Updated - thank you!

statham-arm (approved):
Much better, thanks – the upstream PR link is also helpful.

PR #1032, opened 2026-09-01
[ATfE] Add rule for second file in newlib sample

dcandler:
In the cpp-baremetal-semihosting-newlib sample, the hello-exn.hex rule will not be valid unless hello.elf is built first, since that rule builds both hello.elf and hello-exn.elf. To solve this, hello-exn.elf has been given a separate rule.


voltur01 (approved):
Thanks!

PR #1034, opened 2026-09-01
[llvmlibc-support] Fallback handling for stderr open failure

statham-arm:
Not all semihosting implementations support the convention of opening the special file ":tt" in append mode to indicate stderr. Our Corstone-310 model returns failure from SYS_OPEN if you try. This causes the program to run without any stderr at all, which is worse than squashing stderr into stdout: you can't even find out what went wrong. Better to fall back to opening another copy of stdout, so that we have somewhere to report errors to.


voltur01 (approved):
Thank you for addressing this!

PR #1037, opened 2026-09-03
[libc] Set Root PAS in AArch64 baremetal page tables

MarkMurrayARM:
Set descriptor bit 11 for AArch64 baremetal normal-memory block mappings so FEAT_RME interprets them as Root PAS rather than Secure PAS. Keep NS clear and leave R-profile mappings unchanged.


arm-software/embedded/llvmlibc-support/crt0/memory_8a.h
```diff
@@ -22,6 +22,10 @@ namespace memory {
 
 using namespace sysreg;
 
+static constexpr unsigned long MMU_BLOCK_NSE = 1UL << 11;
+static constexpr unsigned long MMU_NORMAL_FLAGS = 0x405 | MMU_BLOCK_NSE;
```
voltur01: What are these numbers? 0x405 and 0x40d below - would it be possible to either provide a named constant or at least a comment?

arm-software/embedded/llvmlibc-support/crt0/memory_8a.h
```diff
@@ -91,13 +95,15 @@ void setup_mmu(volatile unsigned long *pagetable, unsigned long stackheap_start,
     pagetable[i] = 0;
   }
 
-  // Page occupied by the image
-  unsigned long image_attrs = 0x405; // Index = 1, AF=1
+  // Page occupied by the image. The A-profile FVP crt0 runs at EL3. With
```
voltur01: Is this FVP specific? Do we need to refer to FVP or only EL3, be it QEMU or real hardware as well.

voltur01:
Thank you for addressing my comments!

arm-software/embedded/llvmlibc-support/crt0/memory_8a.h
```diff
@@ -91,13 +104,16 @@ void setup_mmu(volatile unsigned long *pagetable, unsigned long stackheap_start,
     pagetable[i] = 0;
   }
 
-  // Page occupied by the image
-  unsigned long image_attrs = 0x405; // Index = 1, AF=1
+  // Page occupied by the image. AArch64 A-profile startup can run at EL3 on
+  // FVPs or real hardware. With FEAT_RME, descriptor bit 11 is NSE rather than
+  // nG; NS=0,NSE=1 selects Root PAS for EL3 execution. Without RME, setting
```
smithp35: From what I can tell at EL2 and EL3 all translations are global. Without RME the bit is Res(0) but should be ignored by the hardware. With RME at EL3 bit 11 is the NSE bit. At EL0 and EL1 the bit is used for NG.

Perhaps alter the comment to:
"At EL3 with FEAT_RME, descriptor bit 11 is NSE, otherwise it is RES(0) and ignored by the hardware. At EL0 and EL1 bit 11 is non-global, but as we don't change ASID this has no effect."

smithp35 (commented):
Only one comment nit. Otherwise LGTM

smithp35 (approved):
LGTM thanks for the update.

PR #1040, opened 2026-09-04
[ATfE] Enable ATfE downstream libc++ fstream support for LLVM libc

voltur01:
This depends on new LLVM libc embedding API and ATfE semihosting implementation for it.

Downstream issue: #375


voltur01:
Depends on 
* LLVM libc embedding API: https://github.com/llvm/llvm-project/pull/221211
* ATfE semihosting implementation for it: https://github.com/arm/arm-toolchain/pull/1041

PR #1041, opened 2026-09-04
[ATfE] LLVM libc semihosting support for file IO

voltur01:
This provides the semihosting implementation for the new LLVM libc file IO embedding API as discussed in https://discourse.llvm.org/t/rfc-llvm-libc-bare-metal-file-io-support-prototype/91662

The following embedding API libc/src/__support/OSUtil/baremetal/io.h functions are implemented:

- __llvm_libc_stdio_open
- __llvm_libc_stdio_close
- __llvm_libc_stdio_seek
- __llvm_libc_stdio_set_buffer
- __llvm_libc_stdio_flush
- __llvm_libc_stdio_ungetc
- __llvm_libc_stdio_remove
- __llvm_libc_stdio_rename

__llvm_libc_stdio_ungetc is not provided since the semihosting implementation is unbuffered, so the original LLVM libc single char ungetc behavior is kept.

The semihosting implementation keeps a small struct as the cookie with either just the semihosting handle for standard streams or also current position to support seeking for file streams.

### DIFF
--- a/arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
+++ b/arm-software/embedded/llvmlibc-support/semihost/semihost.cpp
@@ -1,30 +1,192 @@
 //
-// Copyright (c) 2022-2025, Arm Limited and affiliates.
+// Copyright (c) 2022-2026, Arm Limited and affiliates.
 //
-// Part of the Arm Toolchain project, under the Apache License v2.0 with LLVM Exceptions. 
-// See https://llvm.org/LICENSE.txt for license information.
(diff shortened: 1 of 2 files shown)


voltur01:
Depends on the LLVM libc embedding API: https://github.com/llvm/llvm-project/pull/221211

voltur01:
Marked as draft since the embedding API is still under discussion: https://discourse.llvm.org/t/rfc-llvm-libc-bare-metal-file-io-support-prototype/91662/5

vhscampos (approved):
LGTM

PR #1042, opened 2026-09-07
[ATfL][build.sh] Try if the use of the lld linker still needs to be forced

pawosm-arm:


pawosm-arm:
If this doesn't fail any of the overnight tests, we can discuss any potential simplifications further.

arm-software/linux/build.sh
```diff
@@ -356,8 +356,8 @@ print_forced_cmake_flags_cache() {
 }
 
 bootstrap_compiler_default_config() {
```
amilendra: This would completely remove the configuration files. I'd rather remove `bootstrap_compiler_default_config` and the calls to `bootstrap_compiler_default_config` and keep the original configuration files untouched.

amilendra: Or if you want to keep the changes as minimal as possible, just make `bootstrap_compiler_default_config` do nothing, and later remove `bootstrap_compiler_default_config` and calls to `bootstrap_compiler_default_config` if overnight tests pass.

pawosm-arm: > This would completely remove the configuration files. I'd rather remove `bootstrap_compiler_default_config` and the calls to `bootstrap_compiler_default_config` and keep the original configuration files untouched.

Ok, I've removed `bootstrap_compiler_default_config` entirely. And in case of the genuine bootstrap compiler, there is no configuration files.

pawosm-arm: No I don't like it. I'd rather restore the previous. What if someone presses Ctrl+C when the 'insane' state of the config files is in power? Any subsequent re-run of build.sh would keep them in power.

amilendra: If you really want a reset after an aborted build, isn't `bootstrap_compiler_build` the best place to do that?
```
bootstrap_compiler_build() {
    mkdir -p "${BUILD_DIR}/stage/bootstrap_compiler"
    rm -f "${BUILD_DIR}"/bootstrap_compiler/bin/clang.cfg
    rm -f "${BUILD_DIR}"/bootstrap_compiler/bin/clang++.cfg
...
```

Anyway, I'll approve this version because testing that `-fuse-ld=lld` is no longer needed is the main objective and this patch does it. We can consider elegance after we get the test results.

amilendra (approved):

PR #1043, opened 2026-09-07
[ATfE] Avoid overloading | operator in run_fvp script

dcandler:
Overloading `|` in this way requires Python 3.10, however LLVM's minimum Python version is still only 3.8.


vhscampos (approved):

voltur01 (approved):
LGTM, thank you!

PR #1045, opened 2026-09-08
[ATfE] Package helper files with the FVP sample

voltur01:
Rework the FVP sample to avoid the need to download helper files from GitHub directly. This also avoids possible mismatch between a specific release version of the sample vs latest GitHub helpers.


HugoSilvaSantos (approved):
LGTM!

PR #1046, opened 2026-09-09
[ATfE] Update xfails and separate libcxx test suites

dcandler:
Now that libcxx tests are run using multiple C libraries, the name of the suite should include the C library so that the results are not abiguous.

To decouple the xfails script from the libcxx test configuration, the xfail list for libcxx/libcxxabi/libunwind are generated individually before the suite is run. This avoids needing any special handling in the test names.

Additionally, the C library has been changed to a property of the xfail class since it is fixed, and does not necessitate a callable function to check this.

Finally, some additionally xfails have been added for arg_list_test and dwarf_expression_stack.

### DIFF
--- a/arm-software/embedded/arm-runtimes/test-support/xfails.py
+++ b/arm-software/embedded/arm-runtimes/test-support/xfails.py
@@ -32,6 +32,7 @@ class XFail(NamedTuple):
     testnames: List[str]  # The tests to include.
     result: NewResult  # The expected result.
     project: str  # Affected project.
+    libc: str = None  # Affected C library.
     variants: List[str] = None  # Affected library variants, if applicable.
     conditional: Callable = None  # A function that will test whether an xfail applies.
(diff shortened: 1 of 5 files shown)


arm-software/embedded/arm-runtimes/test-support/xfails.py
```diff
@@ -32,6 +32,7 @@ class XFail(NamedTuple):
     testnames: List[str]  # The tests to include.
     result: NewResult  # The expected result.
     project: str  # Affected project.
+    libc: str = None  # Affected C library.
```
vhscampos, line 35: Would it make sense to have `libc` as a list of strings so that we can have xfails that cover the two C libraries?

dcandler: The property is there for xfails that only affect a specific C library. If both C libraries are affected, then it can be left unset and it will automatically apply to both.

vhscampos (approved):

PR #1050, opened 2026-09-15
[ATfL][build.sh] write non-empty dummy bootstrap_check_all.xml

amilendra:
Jenkins complains when an empty file is archived.
So write a dummy results file that explicitly states that tests were skipped because a bootstrap compiler was used.


pawosm-arm (approved):
Nice one

PR #1051, opened 2026-09-15
[ATfL] Remove GVN flag from config file

tblah:
No benefit was measured due to the use of this flag. Remove it so that results are more closely comparable with upstream.


momchil-velikov (approved):
LGTM

pawosm-arm (approved):

pawosm-arm (changes requested):
it leaves -mllvm -mllvm (twice)

pawosm-arm:
I've noticed it in a last split second :(

PR #1053, opened 2026-09-15
[ATfL] Remove GVN flag from config file

tblah:
No benefit was measured due to the use of this flag. Remove it so that results are more closely comparable with upstream.


tblah:
@pawosm-arm this should also remove -mllvm

pawosm-arm (approved):

PR #1054, opened 2026-09-17
[ATfE] Propagate test failures in pre/post-merge actions

voltur01:
Propagate test failures in GitHub pre/post-merge actions to mark them red if a test failure occurs by checking the lit test results XML file for errors and failures.

Store build, test logs and the test results XML file to allow further investigation of a failure.


arm-software/embedded/scripts/fail_on_test_failures.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+def main() -> int:
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "directory", type=Path, help="Directory containing test reports"
+    )
+    args = parser.parse_args()
+
+    result_files = sorted(args.directory.rglob("lit_results.junit.xml"))
```
dcandler, line 23: Globing for the test results like this doesn't strike me as the most reliable method, particularly since `lit_results.junit.xml` is not guaranteed to be the correct file name. The shell script running this python script controls the name and location of the lit xml files via `LIT_OPTS`, making it liable to change and the connection will be non-obvious.

One alternative I can think of is to instead of setting one relative results path for all lit tests, run each suite with its own `LIT_OPTS` parameter to collect all the results in a specific location. Then they don't have to be searched for, and the python script can run on that location for all xml files.

This would also help with uploading the files as artifacts: the upload preserves the directory structure, which currently leaves the results files buried under multiple levels of build folders unnecessarily.

arm-software/embedded/scripts/fail_on_test_failures.py
```diff
@@ -0,0 +1,69 @@
...
+    failures = 0
+    errors = 0
+    for result_file in result_files:
+        file_failures = 0
+        file_errors = 0
+        try:
+            # JUnit reports can be large. Stream the XML so it is still
+            # validated without retaining the complete document in memory.
+            for _, element in ElementTree.iterparse(result_file, events=("end",)):
+                tag = element.tag.rsplit("}", 1)[-1]
```
dcandler, line 40: What is being split here? I'm not sure I've ever seen braces in a tag name

voltur01: Tags may have XML namespace in front of them: `{namespace-uri}tag-name` in the internal parser representation.

arm-software/embedded/scripts/fail_on_test_failures.py
voltur01: Right, I can definitely make sure that the shell script passes the file name to the test result checker to make it explicit.

I tried looking into collecting all logs in one place with names based on test suite, variant, C library, etc and it seems to spread all over the place very quickly - not sure how this will play with the CMake scripts refactoring and in general whether it belongs to this change or is a separate refactoring. What do you think?

dcandler: It shouldn't spread anywhere else, it can all be done in the shell script. I meant something like this:

```
declare -a check_targets=(
    "check-all"
    "check-compiler-rt-armv7a_hard_vfpv3_d16_exn_rtti_unaligned"
    "check-compiler-rt-armv7m_hard_fpv5_d16_exn_rtti_unaligned_size"
    "check-cxx-armv7a_hard_vfpv3_d16_exn_rtti_unaligned"
    "check-cxx-armv7m_hard_fpv5_d16_exn_rtti_unaligned_size"
    "check-cxxabi-armv7a_hard_vfpv3_d16_exn_rtti_unaligned"
    "check-cxxabi-armv7m_hard_fpv5_d16_exn_rtti_unaligned_size"
    "check-picolibc-armv7a_hard_vfpv3_d16_exn_rtti_unaligned"
    "check-picolibc-armv7m_hard_fpv5_d16_exn_rtti_unaligned_size"
    "check-unwind-armv7a_hard_vfpv3_d16_exn_rtti_unaligned"
    "check-unwind-armv7m_hard_fpv5_d16_exn_rtti_unaligned_size"
)

for target in "${check_targets[@]}"
do
    export LIT_OPTS="--ignore-fail --xunit-xml-output=${RESULTS_DIR}/${target}_lit_results.junit.xml"
    ninja "${target}"
done

for target in "${check_targets[@]}"
do
    python3 "${SCRIPT_DIR}"/fail_on_test_failures.py "${RESULTS_DIR}/${target}_lit_results.junit.xml"
done

```
(Assuming the `fail_on_test_failures.py` script is modified to run on one file instead of a directory)

That sets a per-target xml output, so they all can all get collected in one place each with a unique name. Then the python script can run on them individually, and the entire folder can be uploaded as an artifact afterwards.

It also has the advantage that if an xml file is missing for some reason, you get an actual error. Whereas scanning for all xml files will still show a pass if finds at least one file (since it doesn't know how many to expect).

arm-software/embedded/scripts/fail_on_test_failures.py
dcandler: Ah, what I meant was that I haven't seen it in the JUnit XML format, since that uses specific tag names. But I suppose being able to read any XML document doesn't do any harm here.

arm-software/embedded/scripts/fail_on_test_failures.py
voltur01: Oh, I see, thank you for the clarification!

I was thinking in opposite directions, in fact:
1. I was going to limit the variants that we build so that in the test script we would just run all available checks (instead of providing a subset of checks here).
2. I tried to implement the naming in CMake so that the logic is reused and not duplicated in these shell scripts.

There may be a middle ground, I hope: let me try creating a test runner script that probes ninja for available checks and then executes them one by one - this way this logic can be reused and variants stay in the build step.

PR #1057, opened 2026-09-17
[ATfE] Remove dwarf_expression_stack test from xfails

vhscampos:
The test has been marked as unsupported in llvm-project because it requires Unix headers, which we don't provide in ATfE.

So now the test can be removed from the xfail, since it doesn't even run anymore.


simpal01:
LGTM

simpal01 (approved):

PR #1058, opened 2026-09-19
[ATfL][build.sh] Add the ability to fail early when any tests are not passing

pawosm-arm:
For local builds and experiments, waiting to the end of the building and packaging process only to find out that not all of the tests were passing is really annoying. When actively working locally on this codebase it is better if a build fails early.


arm-software/linux/build.sh
```diff
@@ -824,7 +832,11 @@ make_and_clean_directory "${LOGS_DIR:?}"
 # The lit test suites do not generate xml results by default.
 # This can be enabled with the --xunit-xml-output option.
 # Each check target appends its own --xunit-xml-output path under LOGS_DIR.
-export LIT_OPTS="${LIT_OPTS:+${LIT_OPTS} }--ignore-fail"
+if [ "$TESTS_MAY_FAIL" == "true" ]; then
```
amilendra, line 835: Shouldn't this be exposed in the `help` output as well?

pawosm-arm: I'm not considering it an 'official' feature.

amilendra (approved):

vrukesh:
This PR was created and merged over the weekend itself ! This does not give a fair chance to everyone for commenting. ;-)

As per commit message it is for "local builds and experiments". Then, can default flag be "false" ? This will help to keep default flows as it is in CI. Otherwise, we will not be able to catch all the test failures in CI runs.  Hope my understanding is correct here.

There is no description added for the flag to indicate its purpose !

pawosm-arm:
> As per commit message it is for "local builds and experiments". Then, can default flag be "false" ?

The flag is 'true' exactly for that reason. Intentionally the change should not be observable in our CI runners which should still be able to complete the build and only warn that there were failing tests. Locally, I'm changing the default setting as it is easier for me to work on a new code if things are failing early when tests are not passing (because of my changes for example).

